PR #12, opened 2019-10-14
schema: add git branch, describe

khilman:
Signed-off-by: Kevin Hilman <khilman@baylibre.com>

### DIFF
--- a/kcidb/db_schema.py
+++ b/kcidb/db_schema.py
@@ -44,6 +44,17 @@ TABLE_MAP = dict(
                         "in the Git repository",
         ),
         Field(
+            "git_repository_commit_name", "STRING",
+            description="A human-readable name of the commit containing the "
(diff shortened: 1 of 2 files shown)


spbnick:
We should definitely add these, but let's work a little on the description and precision of the definition. I pushed a commit on top of yours, which:
* renames `git_repository_describe` to `git_repository_commit_name` (using the terminology from `git-describe(1)`), for clarity;
* switches the order of the two new fields to have `git_repository_commit_hash` and `git_repository_commit_name` together; and
* makes the descriptions more specific.

If you agree with the changes, we can squash and merge this.

Otherwise, let's not do changes to the actual shared database before the code in this repo can handle it. Including adding new data or modifying the schema which breaks it. Let's experiment on private datasets first, and make the shared dataset a safe place.

khilman:
I squashed the two patches, so ready to merge.

PR #17, opened 2019-10-18
Further document and restrict io schema

spbnick:


spbnick:
Also sneaking in a cosmetic change to kcidb.io_schema.validate(), which is not yet used.

PR #23, opened 2020-01-28
Support basic upgrades

spbnick:


README.md
```diff
@@ -77,6 +77,39 @@ can be displayed by `kcidb-schema`.
...
+
+2. Create a new dataset (`kernelci02` in project `kernelci` here) with the new
+   schema:
+
+        bq mk --project_id=kernelci kernelci02
+        # Using new-schema kcidb
+        kcidb-init -d kernelci02
+
+3. Switch all data submitters to using new-schema kcidb and the newly-created
+   dataset.
```
khilman: Is there a way via BQ API to make the old dataset read-only also?

spbnick: The `bq` tool has alpha support for access control modification, but you gotta be on the list of testers to use it. Otherwise you can go to the console and drop the access control manually.

spbnick: Here's a command that should return access control information for the dataset, if you're allowed to use that feature:

    bq get-iam-policy --project_id kernelci kernelci01

spbnick:
I redid the version handling to correct the definition of "backward-compatible" and to make the schema validate the data compatibility fully by itself.

I also added disabling write access of the old dataset to the upgrade steps.

gctucker:
I don't have the time to fully review this but LGTM in principle.

PR #24, opened 2020-02-11
Add tests catalog

spbnick:
Add tests catalog stored in "tests.yaml" file.

Add a tool for validating it, called "kcidb-tests-validate", and a tool
for printing a flat list of tests, called "kcidb-tests-list". The former
will not only validate the YAML syntax and the schema, but will also
check if test homepages are reachable (if you specify the -u/--urls
option). The latter could be useful for e.g. checking if names in your
CI system match the kcidb names correctly.

I'm hesitating to just move the database into the Python code at this
moment, but maybe we'll do that eventually.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,21 @@ install:
 script:
     - flake8 kcidb
     - pylint kcidb
+    - find -name '*.yaml' | xargs yamllint
+    - kcidb-tests-validate --urls < tests.yaml
+    - |
+        test_list=$(
+            python3 <(
+                echo "import sys, yaml"
+                echo "for k in yaml.safe_load(sys.stdin).keys():"
+                echo "    print(k)"
+            ) < tests.yaml
+        )
+        test_list_diff=$(diff -u <(echo "$test_list") \
+                                 <(echo "$test_list" | sort)) ||
+            {
+                echo "Tests out of alphabetic order." >&2
+                echo "Change the order as below:" >&2
+                echo "$test_list_diff" >&2
+                false
+            }
--- a/kcidb/__init__.py
+++ b/kcidb/__init__.py
@@ -5,11 +5,14 @@ import decimal
 import json
 import sys
 from datetime import datetime
+import yaml
+import requests
 import jsonschema
 from google.cloud import bigquery
 from google.api_core.exceptions import BadRequest
 from kcidb import db_schema
 from kcidb import io_schema
+from kcidb import tests_schema
 
 
 class Client:
@@ -233,4 +236,38 @@ def validate_main():
     except jsonschema.exceptions.ValidationError as err:
         print(err, sys.stderr)
         return 2
+    return 0
+
+
+def tests_validate_main():
+    """Execute the kcidb-tests-validate command-line tool"""
+    description = 'kcidb-tests-validate - Validate test catalog YAML'
+    parser = argparse.ArgumentParser(description=description)
+    parser.add_argument(
+        "-u", "--urls",
+        action='store_true',
+        help="Verify URLs in the catalog are accessible"
+    )
+    args = parser.parse_args()
+
+    try:
+        catalog = yaml.safe_load(sys.stdin)
+    except yaml.YAMLError as err:
+        print(err, sys.stderr)
+        return 1
+
+    try:
+        tests_schema.validate(catalog)
+    except jsonschema.exceptions.ValidationError as err:
+        print(err, sys.stderr)
+        return 2
+
+    if args.urls:
+        try:
+            for test in catalog.values():
+                requests.head(test['home']).raise_for_status()
+        except requests.RequestException as err:
+            print(err, sys.stderr)
+            return 3
+
     return 0
--- a/tests.yaml
+++ b/tests.yaml
@@ -1,0 +1,266 @@
+#
+# Catalog of canonical test identifiers to be used when submitting test
+# results to kcidb.
+#
+# Find the test you are running in the catalog, and use its key as the "path"
+# property when submitting the test result to kcidb.
+#
+# Each "test" here can be a test program or a collection thereof, but not one
+# of the tests executed by such a program. You are encouraged to report the
+# latter to kcidb, but they're not recorded or regulated by this catalog.
+#
+# When adding the test corresponding to a program, use the program's
+# widely-known name, e.g. the same one you would call it at a conference. If
+# the program is only known inside your company, prepend its name with the
+# company name.
+#
+# The "title" property should describe the test briefly, and must be
+# single-line. The optional "description" property could go into greater
+# detail and be multi-line.
+#
+# The "home" property should point to the canonical, upstream location, and is
+# intended for identification and as a reference to more information on the
+# test, rather than a pointer to exact code being executed. The latter should
+# be described within test objects submitted to kcidb instead.
+#
+# E.g. the Linux Test Project test suite *should* be here under the
+# widely-known name "ltp". However, while its own tests, such as "chdir01" or
+# "clone03" are welcome to be reported to kcidb as "ltp.chdir01" or
+# "ltp.clone03", only the "ltp" test should be in this catalog. The LTP
+# project's home page or its main upstream repository should be referred to in
+# the "home" property, and not e.g. the specific code being executed, because
+# code versions will differ between test result origins and even executions.
+#
+# Please keep the tests sorted alphabetically by keys.
+#
+# You can use the "kcidb-tests-validate" tool to check if this file is valid.
+# Like this:
+#
+#   kcidb-tests-validate < tests.yaml
+#
+# If you add the -u/--urls option, it will also check if all URLs are
+# reachable:
+#
+#   kcidb-tests-validate --urls < tests.yaml
+#
+audit:
+  title: Basic Audit Regression Test Suite for the Linux Kernel
+  home: https://github.com/linux-audit/audit-testsuite
+blktests:
+  title: Test framework for the Linux kernel block layer and storage stack
+  home: https://github.com/osandov/blktests
+bootrr:
+  title: Bootrr test suite
+  home: https://github.com/andersson/bootrr
+bpf:
+  title: In-tree BPF test suite module
+  home: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/lib/test_bpf.c
+dacapobench:
+  title: DaCapo Java benchmarking suite
+  home: http://dacapobench.sourceforge.net/
+fio:
+  title: Flexible I/O tester
+  home: https://github.com/axboe/fio
+fwts:
+  title: Firmware Test Suite
+  home: https://wiki.ubuntu.com/FirmwareTestSuite/
+jcstress:
+  title: The Java Concurrency Stress tests
+  home: https://wiki.openjdk.java.net/display/CodeTools/jcstress
+kselftest:
+  title: Kernel self-tests
+  home: https://kselftest.wiki.kernel.org/
+kvm-unit-tests:
+  title: KVM unit tests
+  home: https://www.linux-kvm.org/page/KVM-unit-tests
+libhugetlbfs:
+  title: Libhugetlbfs test suite
+  home: https://github.com/libhugetlbfs/libhugetlbfs
+ltp:
+  title: Linux Test Project test suite
+  home: https://linux-test-project.github.io
+nfs_connectathon:
+  title: NFS Connectathon test suite
+  home: http://wiki.linux-nfs.org/wiki/index.php/Connectathon_test_suite
+perftool:
+  title: Perf tool test suite
+  home: https://github.com/rfmvh/perftool-testsuite.git
+pjdfstest:
+  title: A test suite for exercising filesystem-oriented POSIX system calls
+  home: https://github.com/pjd/pjdfstest
+podman:
+  title: Podman test suite
+  home: https://github.com/containers/libpod/tree/master/test
+redhat_acpi_table:
+  title: Red Hat's test for retrieving ACPI tables
+  home: https://github.com/CKI-project/tests-beaker/tree/master/acpi/acpitable
+redhat_amtu:
+  title: Red Hat's Abstract Machine Test Utility
+  description: |
+    Abstract Machine Test Utility (AMTU) is an administrative utility
+    that checks whether the underlying protection mechanisms of the
+    hardware are being enforced. These checks are a requirement of the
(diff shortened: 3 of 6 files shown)


gctucker:
@spbnick I think this is essentially trying to recreate [`test-configs.yaml`](https://github.com/kernelci/kernelci-core/blob/master/test-configs.yaml).

Maybe we should start integrating `kcidb` into the KernelCI core functions and configuration, to generalise it so it can include other test suites such as the CKI ones.

spbnick:
I think that should definitely be our target.

However, our current objective for kcidb is to bring CI systems together. Being at the last stage of the pipeline, reporting, it's easiest and fastest to agree on, and get CI systems to participate. If that's at all possible, we should start here. Once we get the systems invested in cooperation, we would be able to move further back the pipeline, testing, building, etc.

For those reasons and with those targets in mind, we should keep kcidb as simple as possible, and as easy to join as possible, and adding more features and involving it with describing tests (as opposed to purely cataloguing them) would be a hindrance. I'm viewing kcidb more as a vehicle and a field for cooperation, rather than a far-reaching software solution.

Red Hat and Kernel CI could work to make a test cataloguing system which would work for both of us, to replace test-configs.yaml and kpet and kpet-db, but that would only be two of us, and at the same time we'll make it harder for others to join. We need to bring others in, to come up with a standard in Kernel CI, in general. We need to go wide, not deep. Once we have the first "base" - reporting, and everyone participating, then we can go deeper, work on common test descriptions, pipelines, and so on.

With that in mind, let's focus on starting with a simple catalog of test names we can use for correlating results. Just something simple and easy to agree on, or disagree and modify.

gctucker:
> I think that should definitely be our target.
> 
> However, our current objective for kcidb is to bring CI systems together. Being at the last stage of the pipeline, reporting, it's easiest and fastest to agree on, and get CI systems to participate. If that's at all possible, we should start here. Once we get the systems invested in cooperation, we would be able to move further back the pipeline, testing, building, etc.

I agree, things need to happen one step at a time.  Starting with the common database and generating a report based on it is a valid way to go.

My point was that defining a test catalogue is already going beyond reporting.  It is about starting to solve the same thing as kernelci-core is already doing, that is to say coordinating various components into a heterogeneous system.

> For those reasons and with those targets in mind, we should keep kcidb as simple as possible, and as easy to join as possible, and adding more features and involving it with describing tests (as opposed to purely cataloguing them) would be a hindrance. I'm viewing kcidb more as a vehicle and a field for cooperation, rather than a far-reaching software solution.

Keeping `kcidb` as simple as possible: to me that means providing a tool that can simply push some results to a database.  Anything related to what the content of the data may be about is adding complexity to it and starts creating ties with other parts with the pipeline.  It's easy enough to reshape the kernelci-core YAML configuration and have abstract test suites defined there.  In fact, that's what it was intended for.

> Red Hat and Kernel CI could work to make a test cataloguing system which would work for both of us, to replace test-configs.yaml and kpet and kpet-db, but that would only be two of us, and at the same time we'll make it harder for others to join. We need to bring others in, to come up with a standard in Kernel CI, in general. We need to go wide, not deep. Once we have the first "base" - reporting, and everyone participating, then we can go deeper, work on common test descriptions, pipelines, and so on.

One thing I have to disagree with is that there isn't a KernelCI vs CKI duality.  There is KernelCI and contributors to it, including CKI among other test labs and systems  The current KernelCI architecture is designed to be generic and distributed, not specific to a use-case like most product-centric CI systems are.  Of course it isn't perfect and things need to evolve to facilitate other types of contributions, but starting with the predicate that we need a new design from scratch without first considering what we currently have doesn't seem like a wise move to me.

> With that in mind, let's focus on starting with a simple catalog of test names we can use for correlating results. Just something simple and easy to agree on, or disagree and modify.

It is just as easy to agree on a YAML file in kernelci-core as it is in kcidb.  I can see why you may find it easier to work on `kcidb` having started it, but would a complete stranger to KernelCI find it easier than `kernelci-core` tools?

I don't want to be slowing down progress, we do need to experiment and explore new areas but I'm worried that what appears like quick wins may easily end up being counter-productive.

tests.yaml
```diff
@@ -0,0 +1,239 @@
...
+#
+#   kcidb-tests-list < tests.yaml
+#
+title: The top-level "test" signifying testing in general
+description: |
+    Report results on this test when you don't have any more details on your
+    testing, and/or when you want to signal your testing for the build has
+    concluded.
+tests:
+    storage:
```
gctucker: What is the rationale for having such categories?  How about having a flat list of test suites instead, maybe with tags to group them if needed?

I fear having top-level categories might be a bit artificial and subjective, and if we enforce them in the test names we may be causing unnecessary complications.

tests.yaml
```diff
@@ -0,0 +1,239 @@
...
+        title: Storage subsystem tests
+        tests:
+            fio:
+                title: Flexible I/O tester
+                home: https://github.com/axboe/fio
+            blktests:
+                title: Test framework for the Linux kernel block layer and storage stack
+                home: https://github.com/osandov/blktests
+            redhat_scsi_vpd:
```
gctucker: It would seem better imho to call these CKI tests `cki_*` rather than `redhat_*`.

tests.yaml
```diff
@@ -0,0 +1,239 @@
...
+            redhat_pmtu:
+                title: Red Hat's tests for Path MTU discovery
+                home: https://github.com/CKI-project/tests-beaker/tree/master/networking/route/pmtu
+            redhat_route:
+                title: Red Hat's IP routing tests
+                home: https://github.com/CKI-project/tests-beaker/tree/master/networking/route/route_func
+            redhat_sctp_auth_sockopts:
+                title: Red Hat's SCTP authentication socket options tests
+                home: https://github.com/CKI-project/tests-beaker/tree/master/networking/sctp/auth/sockopts
+            redhat_sock_fuzz:
```
gctucker: `socket_fuzz` would seem more consistent with how the others are named, reusing the path `socket/fuzz`.  I don't mind really, but if it caught my eye it might catch others' too.

tests.yaml
```diff
@@ -0,0 +1,239 @@
...
+            redhat_igmp:
+                title: Red Hat's IGMP sanity tests
+                home: https://github.com/CKI-project/tests-beaker/tree/master/networking/igmp/conformance
+    security:
+        title: Security-related tests
+        tests:
+            selinux:
+                title: SELinux Regression Test Suite for the Linux Kernel
+                home: https://github.com/SELinuxProject/selinux-testsuite
+            audit:
```
gctucker: Maybe `linux-audit` would be a safer choice, as `audit` is a common word.

tests.yaml
spbnick: Yes, this is my fear as well. I agree this is subjective, and could be an unnecessary point of contention. I also think we can go with a flat list.

My only argument for this was that a flat list would be harder to page through when viewing results, and harder to navigate when cataloging new tests (although that's much less of a problem).

OTOH, we can have tags and even multiple categorization schemes outside of kcidb and use those when displaying results. However, I was hesitating to go with that, as the requisite queries/logic could become quite a bit complicated and would take a while to implement, and meanwhile we'd have to deal with a flat list.

Still, I don't feel strongly about this, and I think a flat list is a good way to go in general. We get less arguing when cataloging and agreeing on the names, and we can have as many categorization schemes as we want. So if you agree, and don't see the above arguments as a problem, let's go with a flat list :smiley:.

tests.yaml
spbnick: I see what you mean, but those tests predate CKI by a wide margin. I don't think there are any tests which are CKI-specific (as in created to run in CKI) here, or maybe one or two only. They only have "redhat_" in front of them, because they're only known inside Red Hat. Furthermore, you shouldn't have too much of a hard time running them outside CKI, and it's our target to improve that.

So, I'd rather keep "redhat_" in the names, as that would be more to the point.

tests.yaml
spbnick: Yeah, it could be clearer to name it "redhat_socket_fuzz", indeed. I'll change that.

tests.yaml
spbnick: Eehhh, this is all about "linux", and those are the **official** tests for **the** audit subsystem. Those are the tests which involve `kernel/audit.c` after all. So, I think this test suite deserves an unqualified name "audit".

tests.yaml
spbnick: One note: if we go with the flat list, it would be best to have test categorization/tagging data inside whatever database we'll use for storing test results. At least in addition to whatever place will be authoritative, if any. That would make it possible to organize the data using queries only and might simplify the code overall.

spbnick:
Thank you for the review, and for the thoughtful comments, @gctucker! I also think it should be our target to get an agreement and start pushing data first, and merging this into kcidb would probably be the fastest way. I'm not planning on getting this further anytime soon, and if anything we'll only need categorization data for the interface. Again, this is not about describing the tests themselves, but only keeping an authoritative list of names, plus organizing them a bit. I'd be happy to work on integrating these into Kernel CI proper, provided it makes sense there.

Please tell me what you think about using the flat list, I'll amend the PR accordingly, and let's get to actually submitting the results using whatever names we agree upon :smiley:.

tests.yaml
gctucker: > So if you agree, and don't see the above arguments as a problem, let's go with a flat list .

Yes I'm still in favour of a flat list, and I agree with what you wrote.

The tags should indeed be stored in the database in test records.  However, they need to come from somewhere, so having them in the test catalogue and using them when submitting the results would be one possibility.  If not in the test catalogue, it could also be in `test-configs.yaml` or an equivalent configuration file depending on how the YAML data ends up being shaped.

tests.yaml
gctucker: OK fine, I thought of this because the projects are hosted under `CKI-project` on GitHub.  Maybe if they had been in another more general Red Hat organisation it would have seemed more natural to call them `redhat_*`.  But I guess you guys are the authority on this topic, and as long as the tests aren't completely tied to CKI and Beaker then great :)

tests.yaml
gctucker: Well we may have tests for non-Linux systems in the future.  But yes `audit` should be fine within the current KernelCI context.

tests.yaml
spbnick: Well, somebody **had** to get them outside of the company, and it just so happened that we needed that the most :grin:

They're tied most to Beaker, and much less so to CKI, if at all. Beaker is much, much older. Still we're working on getting them untied as much as possible.

spbnick:
Alright, updates done:

* Flattened the list by removing the categories, and prefixing names of tests in "sound" category with "sound_" for clarity.
* I kept the schema the same and the `kcidb-tests-list` tool, to get this merged faster and leave us a way back to categories, if we find out something doesn't work. I'll post an issue to remind us to flatten the schema as well and remove the tool, if we see we truly don't need them.
* Renamed `redhat_sock_fuzz` to `redhat_socket_fuzz`.
* Sorted the resulting "tests" dictionary by key alphabetically to make it easier to find the tests, and added a rule of thumb to the description.
* Added a yamllint configuration and an execution to Travis CI to catch possible test name duplication and other YAML issues.

spbnick:
More updates:
* Add a check for alphabetic test ordering to Travis CI
* Add more tests executed by Red Hat (but not necessarily currently reported for various reasons) - in a separate commit to simplify review.

tests.yaml
```diff
@@ -77,24 +77,48 @@ tests:
...
+  bpf:
+    title: In-tree BPF test suite module
+    home: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/lib/test_bpf.c
   dacapobench:
     title: DaCapo Java benchmarking suite
     home: http://dacapobench.sourceforge.net/
   fio:
     title: Flexible I/O tester
     home: https://github.com/axboe/fio
+  fwts:
```
gctucker: Arguably not a Linux kernel test suite, but it exercises the kernel so might be fine in this respect.

The results are very likely to be tied to the firmware revision present on the DUT though, rather than the kernel, which can complicate things a bit.

.travis.yml
```diff
@@ -7,3 +7,13 @@ install:
 script:
     - flake8 kcidb
     - pylint kcidb
+    - find -name '*.yaml' | xargs yamllint
+    - kcidb-tests-validate --urls < tests.yaml
+    - |
+        sort_diff=$(diff -u <(kcidb-tests-list < tests.yaml) \
```
gctucker: That works until `kcidb-tests-list` gets "improved" to start printing the list in alphabetical order ;)

kcidb/__init__.py
```diff
@@ -234,3 +237,62 @@ def validate_main():
...
+        catalog = yaml.safe_load(sys.stdin)
+    except yaml.YAMLError as err:
+        print(err, sys.stderr)
+        return 1
+
+    try:
+        tests_schema.validate(catalog)
+    except jsonschema.exceptions.ValidationError as err:
+        print(err, sys.stderr)
+        return 2
```
gctucker: Is there a convention about returning anything else than 0 or 1 in this context?

tests.yaml
spbnick: Yeah, this is borderline, but if we see it is useless (i.e. unreliable) we can just stop reporting it and drop it from the catalog. So far we're running it for RHEL only, but we can likely run it for upstream as well.

.travis.yml
spbnick: True, but it won't need to be improved if we keep this in :grin:
And the point is to have the YAML sorted, not the tool's output.
We actually might not need the tool if we get rid of categories.
The YAML file will be clear enough by itself.

kcidb/__init__.py
spbnick: Not really, but nothing usually checks for anything else but non-zero, and in theory it could be useful :)

spbnick:
Alright, I went ahead flattened the catalog, and squashed all commits into one.

We might have some changes to a couple Red Hat tests soon, but they can go in whenever, either this or a follow-up PR.

tests.yaml
gctucker: OK - also there have been discussions in the past about using KernelCI infrastructure to test firmwares and bootloaders, in which case the kernel would be a known good one and not a moving part.  I guess it's benign to have `fwts` in the catalog as long as we all know what to expect from this kind of tests.

kcidb/__init__.py
gctucker: That's a bit far-fetched but sure why not :)

PR #26, opened 2020-02-27
tests: Add boot test

spbnick:
Catalog a test signifying the kernel booted successfully. Call it "boot".

### DIFF
--- a/tests.yaml
+++ b/tests.yaml
@@ -49,6 +49,9 @@ audit:
 blktests:
   title: Test framework for the Linux kernel block layer and storage stack
   home: https://github.com/osandov/blktests
+boot:
+  title: Machine booted with the tested kernel
+  home: https://www.kernel.org/
 bootrr:
   title: Bootrr test suite
   home: https://github.com/andersson/bootrr


spbnick:
I think we reached a tentative agreement with @gctucker on IRC over this. At least we didn't have much of disagreement. I'll merge this now, and setup CKI to start sending the data for it.

However, this will actually come to a head when Kernel CI tries to send its own data with the IDs from the catalog, and adds its own IDs.

gctucker:
@spbnick There will be no boot results coming from the kernelci-backend as we're only sending actual test results.  Boot results are being removed from kernelci.org to be replaced by the `baseline` test suite (login + dmesg checks + bootrr).

Do you actually have "boot" results to submit from the RadHat lab / CKI?

spbnick:
We have boot results submitted under a separate "test". I understand that you moved away from a separate "boot" test. However, we have to agree on how to name common parts of testing/reporting so that developers can more easily find them, regardless of the origin CI system. 

If Red Hat submits them under "boot", and Kernel CI submits them as something else, that would be confusing and burdensome for developers.

I think the name "boot" is the clearest possible name for trying to boot a machine to a login prompt. However, I'm willing to name it otherwise so long as it's a thing by itself and is not combined with something else. If only because we (CKI, and likely others) can run any combination of tests (including none), but the "boot" will always be there.

Regarding whether "boot" is a test or not, I would say there are several things to consider. First of all, "booting" is a fundamental piece of functionality for an OS kernel, and it has to be tested. Yes, we cannot run any tests without booting, so in that sense it's special. However, I would argue that there are other tests like that and "booting" is just one special case.

There's other functionality like that in the kernel and it can still have a separate test. E.g. creating a loop device. If creating a loop device fails, we cannot run other loop device tests meaningfully (same as with "boot" and all other tests). However, we would still want loop device creation as a separate test just for clarity and ease of determining what broke.

If we only ever needed one loop device created for all other loop device tests, and it took considerable testing time to do it, we might have chosen to make that test be the setup for all other loop device tests, as well.

We can say that this is what's happening with the "boot" test: we definitely need to know whether the kernel boots or not, regardless of whether other tests ran or not, but we choose to make that test be the setup for all other tests, just because it takes time to boot the machine, and because every other test needs it.

spbnick:
Oh and, of course, what constitutes boot will be different, even if we say that's e.g. "execution until a login prompt". Because the login prompt could be graphics or text, we could have different distros, we can have different distro releases, and even with the same distro and release, different hardware could result in different user-space services to be started, let alone different kernel modules to be loaded.

And that's OK. We can move the cut-off point for the "boot" back and forth. Just let's agree on something.

spbnick:
> I was just wondering if CKI was actually already producing "boot" results,
> and it sounds like it is.

Yes, we do. Here are a few examples (using the merged "boot" ID):

    https://staging.kernelci.org:3000/d/test/test?orgId=1&var-origin=redhat&var-origin_id=107180011
    https://staging.kernelci.org:3000/d/test/test?orgId=1&var-origin=redhat&var-origin_id=107179970
    https://staging.kernelci.org:3000/d/test/test?orgId=1&var-origin=redhat&var-origin_id=107179806

> KernelCI has almost become synonymous with boot testing, and we need to show
> it's not true any more. Also we found that having very fast checks made a
> lot more sense than a dumb "boot", as you can easily log in with many kernel
> errors and the system may crash and burn 5 seconds later... Hence the
> "baseline" test plan we now have on kernelci.org.

Sure, any checks on top of simple checking for a login prompt are good. I think
we have something built into the test framework that keeps watching the dmesg
log for any panics, oops'es, etc. and attributes them to the currently-running
test, including "boot", and sets the appropriate test status if it sees them.

I would say it's OK if one CI system's "boot" test does a little more, or a
little fewer checks, compared to the others, as long as it does roughly the
same. However, I would report any other considerable tests separately.

> But yes we need some kind of least common denominator for the common
> database. We can easily use the `baseline.login` test case here from current
> kernelci.org results, and `baseline.bootrr.*` separately. It sounds like you
> also have something that would map directly with a "login" boot test in CKI.

I'm not sure what `baseline.login` exactly does, but if it simply checks that
we got to the login prompt and e.g. that there were no major issues with that,
then that would map to what we're already reporting as "boot" test. So if you
could report it with "boot" path, I think it would work well.

> We probably need to also have a way to report serious kernel errors that did
> not cause a failure to login.

I think it's OK to just report it as part of the "boot" test. However, if it's
hard to put those parts together and report as one, you can report the boot
itself and the checking separately as subtests of "boot". E.g. as "boot.run"
and "boot.check", and the UI should be able to summarize them under "boot".

> So how about having a list of baseline test
> case names documented somewhere, maybe even with some handy scripts, and
> that would be the URL for this boot/baseline "test suite"? For example, it
> could be a kernelci/baseline repo on GitHub. Each contributor to the common
> database could send results for the test cases they care to run, such as
> `baseline.login` as a bare minimum. Other test cases could include kernel
> errors such as panic etc...

The point of test catalog and test "paths" is to identify the common test
suites all CI systems could share and possibly run. If a CI system runs its
own unique test suite, then it deserves a separate entry. If nobody else knows
it (yet), then we should make it clear where it comes from (which "namespace"
it belongs to), and the way we do it is by prefixing the suite name with the
origin name. That's where all those "redhat_*" tests come from in the catalog.

However, if a CI system runs a test suite which is simply a combination of
other generally-known test suites, then it shouldn't report it as e.g.
"redhat_blah", but instead should report those test suites separately, under
their own names.

An example of that for CKI was the internally-called "jvm" test suite, which
actually ran two generally-known test suites: "DaCapo Java benchmarks" and
"jcstress" suite. So, instead of reporting it as "redhat_jvm", we've split the
test suite into two, and are reporting them separately as "dacapobench" and
"jcstress". This increases the chance that somebody else would run and report
them under the same names, thus allowing us to correlate the data and make it
easier for developers to interpret results.

So, if Kernel CI's "baseline" test plan is running other test suites which are
better known, and more often executed by themselves, than it is as a whole,
then it would be best to report its results separately, under the names of
those suites, and not e.g. as "kernelci_baseline".

E.g. have a mapping of "baseline.login" => "boot" and "baseline.bootrr" =>
"bootrr", and similar for the rest. Because (presuming) it is more likely that
other CI systems will run the "bootrr" suite themselves, than that they will
run the "baseline" test suite itself, and they will have to run the "boot"
test in any case.

It could be more productive for us to look at the list of all tests inside the
"baseline" test plan, go through them and decide what's best to do.

Finally, if it turns out too hard to take it apart and report its pieces (such
as "bootrr") separately, we can have it named "baseline", provided it's
generally known and executed by itself, or "kernelci_baseline" otherwise, and
only extract the "boot" part from it.

Hope some of this makes sense :)

gctucker:
We all know roughly what boot means, and bootrr is a test suite on its own.  So that's fine.

My main question was, how do we deal with any other checks such as kernel errors?  Reporting a failed boot because there was a kernel error message doesn't sound right.

That's why I'm proposing to add a dimension with a list of test cases, rather than just "boot".  I'm using the "baseline" example from kernelci but it could be called something else.  The idea would be to have "baseline.login", "baseline.dmesg.crit" and things like that, to be able to report more than just boot.  Or, we could have a "dmesg" test suite if that makes more sense.

spbnick:
> My main question was, how do we deal with any other checks such as kernel errors? Reporting a failed boot because there was a kernel error message doesn't sound right.

AFAIK, CKI checks for errors in dmesg for every test we execute, not just boot, and fails the test if it finds any. @veruu, @rasibley, please correct me if I'm wrong.

I think it would be hard for us to separate the results of that checking from the pure results of the test suite. However, if we did, how would we report that, and how would we associate them to the corresponding test suite? Or should we discard them instead?

Could we instead, for the start, define a test status as the original (pure) test suite status, plus whatever extra checks the CI system or the test suite wrapper does on the produced results? I.e. check dmesg for anomalies, check the test suite output for any suspicious stuff, etc.? All CI systems have some sort of wrappers for test suites anyway and those always introduce some deviation to test results. Can we try going with that definition, and see what the developers say? And, in general, can we focus on getting something to developers faster, so we can get real feedback?

With that definition in mind, we can simply fold both the boot and the dmesg checks under the "boot" path. To free Kernel CI from having to summarize those separate results, you can report them as subtests of "boot". E.g. map "baseline.login" => "boot.login" and "baseline.dmesg.crit" => "boot.dmesg", or "baseline.dmesg.crit" => "boot.dmesg.crit", if that's easier, and so on. And you won't have to report a separate test run with just "boot" path. What do you think?

gctucker:
Right, well that's exactly what I'm suggesting except you want to call it "boot" rather than "baseline".

So the next step is to have a list of test case names to use when reporting things under the "boot" umbrella, and I think a repo with at least a README would probably be enough.  We can discuss the details of that later though.

spbnick:
Great :)

I would say, let's not go too detailed in the test catalog yet. Let's just have "boot" there. We can, of course, have the Kernel CI parts of it documented, but we shouldn't ask anyone else to adhere to that yet.

What we need now WRT test reporting is having both CI systems report test "path", and have them use the same parts of it, where they (roughly) match. That's all. CKI already reports paths we agreed upon so far, we need Kernel CI start reporting them, and have their tests added to the catalog.

gctucker:
I mean, having a repo with the list of test case names that people want to report would help to avoid the same things being reported with different names.  I'm not saying the test case names should be in the catalogue, but the URL should point to that repo with the details.

Things like `boot.login` etc...

I thought kernelci-backend was already reporting the test case paths, but it probably needs to be revisited now that we have this catalogue.

spbnick:
I agree, it would be good to have people report same things with the same names. I'm just saying we shouldn't ask people to do that, it was hard enough for us to agree on the "boot" name, and asking people to contribute even more detailed and specific results before we even have anything to show for that would be too much.

Let's document what kernel CI reports (and merge that with CKI and others later), but let's not ask people to agree to that yet. Let's build an incentive for them to do that first.

PR #28, opened 2020-03-02
Separate user and admin interfaces

spbnick:
Beside a few small fixes, this mainly deals with separating the user and administrator interfaces. The former is abstracted from the implementation and only has two operations ("submit" and "query"), while the latter goes (will go) into details and deals with infrastructure directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 KCIDB
 =====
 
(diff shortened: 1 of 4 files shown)


spbnick:
Merging due to review timeout as agreed with @khilman :)

PR #35, opened 2020-03-11
io_schema: Permit fallback to git:// repo URLs

spbnick:


spbnick:
Thanks for the review, Kevin!

PR #39, opened 2020-03-12
Merge "origin" and "origin_id" fields into "id"

spbnick:


kcidb/io/schema/__init__.py
```diff
@@ -1,8 +1,11 @@
 """Kernel CI reporting I/O schema"""
 
 from kcidb.io.schema import v1
+from kcidb.io.schema import v2
```
gctucker: Do we need to keep a copy of each schema version? Are we actually planning to have backwards-compatibility support for this PoC?

spbnick: Yes. Apart from letting all submitters upgrade gradually, we will use this when upgrading the datasets to new schema (download, convert, upload).

gctucker: I see.  Is there an automated way of migrating a dataset from v1 to v2 from a db dump?

spbnick: Not completely automated, but there are instructions, in README.md.

spbnick:
Thank you for the review, @gctucker! Merging!

PR #42, opened 2020-03-26
io.schema: Prohibit directories in resource names

spbnick:


khilman:
I don't have a strong opinion on this, but it seems a bit restrictive to me at this stage.  Why refuse directories?

If this is just going to be used for arbitrary resrouces (e.g. input/output files) it seems a bit premature to prohibit completely.

spbnick:
@khilman, if we want to let people download individual files from our eventual result file server, then we won't be able to preserve the directory names from the "name" in the download response headers (`Content-Disposition: filename=<NAME>`), it will be lost.

spbnick:
I just saw our database pushing names with directories, submitted an issue internally, and would like to prevent that from happening again.

khilman:
OK, makes sense, thanks for explanation.  I don't think staging.kernelci.org is currently pushing any directories currently.  If so, we'll have to fixup.

spbnick:
@khilman, I didn't see that from kernelci, just CKI test output files.

PR #72, opened 2020-04-22
Only send mainline notifications

spbnick:


spbnick:
CI passed, but stuck reporting here. Merging!

PR #84, opened 2020-05-12
Add v3 schema

spbnick:
Add a v3 I/O schema with the following changes.

* Require separate `origin` property for each object (again), but this time keep it in IDs too.
* Require revision's `git_repository_url` property to be either an https:// or a git:// URL, as per documentation.
* Require revision's `git_repository_commit_hash` property to be either a sha1 or a sha256 lowercase hexadecimal hash, as Git does.
* Change requirements for the revision's `id` property to be a certain format (repo@hash+patchhash), without origin, so we can correlate revisions coming from different CI systems.

Support upgrade from v2 schema, but only if the revision IDs match the new requirements (without origin).

The upgrade plan therefore is the following:
1. All submitters switch to generating new-schema IDs (but with origin still added)
2. All submitters switch to pushing to a topic accepting the new schema, which is pushed to a new-version dataset.
3. Old data is processed to update revision links, where possible, with the rest of the data dropped (shouldn't be any), and pushed to the new dataset.
4. All submitters update the schema of their data.


gctucker:
Is it really better to use the `id` field with data to look for identical revisions, rather than using actual data fields?  Is it because it's hard to create an index using multiple fields in BQ?

For example, when a new revision record comes in, to either find if it has already been created and avoid duplicates or to later find all the identical records if we tolerate duplicates, a query could use the remote git url, the base git sha1 and a hash for patches as a key.  The `id` could then be any arbitrary unique identifier, like we had before with `namespace:unique-thing`.

That's how kernelci-backend handles adding information to an existing record in Mongo DB, for example to add the result of a bisection when it completes, or to look up a build entry matching the meta-data provided with test results and link them together.  It works really well with a few custom indexes.

spbnick:
@gctucker, that's a good question, and the approach is worth considering.

First, BigQuery doesn't do indexes. There are some tips on performance instead: https://cloud.google.com/bigquery/docs/best-practices-performance-overview

Second, with BigQuery we can't really avoid duplicates and we have to tolerate them. Furthermore, I'd like to avoid complicating submission process as much as possible, do it without any queries, and I'd like to consider records immutable once they're stored, only adding data in new rows. More of a log than a database.

The problem with identifying objects in KCIDB is that the submitter should be able to link them themselves before submitting, since they should be able to submit them in any order. By limiting the identification to a single ID string we make that simpler. They only have to generate them once and just pass around afterwards. It also allows submitters to easily query the submitted and linked objects back from the database, open them in report WebUI, and so on. Finally, it makes the KCIDB code simpler.

I don't think I've explored all angles of your approach, however, and will keep thinking about it. Perhaps I'm missing something.

spbnick:
This PR in its current shape is no good, though, because the upgrade function would only work when both revisions and all their builds are present in a single submission. Otherwise it will fail to update some links. I'll have to rethink this.

I also will likely add "origin" field to *all* objects and stop extracting any information from the IDs, keeping them opaque to KCIDB and Grafana. There will be data duplication within rows (and this is what your approach avoids), but it will let us keep the database links working without updating the code, as we continue tweaking how we identify things.

gctucker:
OK, let's catch up again next week to revisit this one more time.

I think the main goal of kcidb is to come up with a "reference" schema that works for collating results from different kernel CI sources.  It's important to be able to define this independently of the particularities of any given database engine initially.  We'll probably still have to consider all alternatives and choose an optimal engine once we have a more mature schema.

spbnick:
I absolutely agree with the goals. And the current approach will work with any SQL database. BigQuery will also work with multi-field links, but currently I feel it will add too much complexity.

spbnick:
Alright, ready for review again. Please also see the updated PR description!

spbnick:
@Inakimalerba, heads-up!

gctucker:
fwiw here's the diff between v2 and v3:
```
--- kcidb/io/schema/v2.py	2020-05-19 13:46:05.965556043 +0100
+++ kcidb/io/schema/v3.py	2020-05-19 13:46:05.965556043 +0100
@@ -1,16 +1,40 @@
-"""Kernel CI reporting I/O schema v2"""
+"""Kernel CI reporting I/O schema v3"""
 
+import re
 from kcidb.io.schema.misc import Version
-from kcidb.io.schema import v1
+from kcidb.io.schema import v2
 
 # Major version number of JSON schema.
-JSON_VERSION_MAJOR = 2
+JSON_VERSION_MAJOR = 3
 
 # Minor version number of JSON schema.
 JSON_VERSION_MINOR = 0
 
-# A regular expression pattern matching permitted ID strings
-ID_PATTERN = "^[a-z0-9_]+:"
+# A regular expression pattern matching strings containing accepted Git
+# repository URLs
+GIT_REPOSITORY_URL_PATTERN = "(https|git)://.*"
+
+# A regular expression pattern matching strings containing sha1 hashes
+SHA1_PATTERN = "[0-9a-f]{40}"
+
+# A regular expression pattern matching strings containing sha256 hashes
+SHA256_PATTERN = "[0-9a-f]{64}"
+
+# A regular expression pattern matching strings containing Git repository
+# commit hash (sha1 or sha256)
+GIT_REPOSITORY_COMMIT_HASH_PATTERN = f"({SHA1_PATTERN}|{SHA256_PATTERN})"
+
+# A regular expression pattern matching strings containing revision IDs
+REVISION_ID_PATTERN = \
+    f"{GIT_REPOSITORY_URL_PATTERN}" \
+    f"@{GIT_REPOSITORY_COMMIT_HASH_PATTERN}" \
+    f"(\\+{SHA256_PATTERN})?"
+
+# A regular expression pattern matching strings containing origin name
+ORIGIN_PATTERN = "[a-z0-9_]+"
+
+# A regular expression pattern matching strings containing an object ID
+ORIGIN_ID_PATTERN = f"{ORIGIN_PATTERN}:.*"
 
 # JSON schema for a named remote resource
 JSON_RESOURCE = {
@@ -72,12 +96,38 @@
             "description":
                 "Revision ID.\n"
                 "\n"
-                "Must start with a non-empty string identifying the CI "
-                "system which submitted the revision, followed by a colon "
-                "':' character. The rest of the string is generated by the "
-                "origin CI system, and must identify the revision uniquely "
-                "among all revisions, coming from that CI system.\n",
-            "pattern": ID_PATTERN,
+                "Must be a string containing the following, in order.\n"
+                "\n"
+                "The URL of the Git repository which contains the base code "
+                "of the revision. The shortest possible https:// URL, or, if "
+                "that's not available, the shortest possible git:// URL."
+                "\n"
+                "The '@' character.\n"
+                "\n"
+                "The full commit hash of the revision's base code in the Git "
+                "repository.\n"
+                "\n"
+                "If the revision had patches applied to the base code, "
+                "the '+' character, followed by a sha256 hash over "
+                "newline-terminated sha256 hashes of each applied patch, "
+                "in order. E.g. generated with this shell command: \""
+                "sha256sum *.patch | cut -c-64 | sha256sum | cut -c-64"
+                "\".\n",
+            "pattern": f"^{REVISION_ID_PATTERN}$",
+            "examples": [
+                "https://git.kernel.org/pub/scm/linux/kernel/git/stable/" +
+                "linux-stable.git@aa73bcc376865c23e61dcebd467697b527901be8",
+                "https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/" +
+                "linux.git@c0d73a868d9b411bd2d0c8e5ff9d98bfa8563cb1" +
+                "+903638c087335b10293663c682b9aa0076f9f7be478a8e782" +
+                "8bc22e12d301b42"
+            ],
+        },
+        "origin": {
+            "type": "string",
+            "description":
+                "The name of the CI system which submitted the revision",
+            "pattern": f"^{ORIGIN_PATTERN}$",
         },
         "git_repository_url": {
             "type": "string",
@@ -86,12 +136,14 @@
                 "The URL of the Git repository which contains the base code "
                 "of the revision. The shortest possible https:// URL, or, if "
                 "that's not available, the shortest possible git:// URL.",
+            "pattern": f"^{GIT_REPOSITORY_URL_PATTERN}$",
         },
         "git_repository_commit_hash": {
             "type": "string",
             "description":
                 "The full commit hash of the revision's base code "
                 "in the Git repository",
+            "pattern": f"^{GIT_REPOSITORY_COMMIT_HASH_PATTERN}$",
         },
         "git_repository_commit_name": {
             "type": "string",
@@ -178,6 +230,7 @@
     "additionalProperties": False,
     "required": [
         "id",
+        "origin",
     ],
 }
 
@@ -192,7 +245,7 @@
             "description":
                 "ID of the built revision. The revision must "
                 "be valid for the build to be considered valid.",
-            "pattern": ID_PATTERN,
+            "pattern": f"^{REVISION_ID_PATTERN}$",
         },
         "id": {
             "type": "string",
@@ -204,7 +257,13 @@
                 "':' character. The rest of the string is generated by the "
                 "origin CI system, and must identify the build uniquely "
                 "among all builds, coming from that CI system.\n",
-            "pattern": ID_PATTERN,
+            "pattern": f"^{ORIGIN_ID_PATTERN}$",
+        },
+        "origin": {
+            "type": "string",
+            "description":
+                "The name of the CI system which submitted the build",
+            "pattern": f"^{ORIGIN_PATTERN}$",
         },
         "description": {
             "type": "string",
@@ -284,6 +343,7 @@
     "required": [
         "revision_id",
         "id",
+        "origin",
     ],
 }
 
@@ -309,7 +369,7 @@
             "description":
                 "ID of the tested build. The build must be "
                 "valid for the test run to be considered valid.",
-            "pattern": ID_PATTERN,
+            "pattern": f"^{ORIGIN_ID_PATTERN}$",
         },
         "id": {
             "type": "string",
@@ -321,7 +381,13 @@
                 "':' character. The rest of the string is generated by the "
                 "origin CI system, and must identify the test run uniquely "
                 "among all test runs, coming from that CI system.\n",
-            "pattern": ID_PATTERN,
+            "pattern": f"^{ORIGIN_ID_PATTERN}$",
+        },
+        "origin": {
+            "type": "string",
+            "description":
+                "The name of the CI system which submitted the test run",
+            "pattern": f"^{ORIGIN_PATTERN}$",
         },
         "environment": {
             "type": "object",
@@ -417,6 +483,7 @@
     "required": [
         "build_id",
         "id",
+        "origin",
     ],
 }
 
@@ -523,31 +590,21 @@
     Returns:
         The inherited data.
     """
-    # Merge *origin and *origin_id properties into *id properties
-    # pylint: disable=redefined-builtin,invalid-name
-    for collection, id_pair_map in \
-        dict(revisions=dict(id=('origin', 'origin_id')),
-             builds=dict(id=('origin', 'origin_id'),
-                         revision_id=('revision_origin',
-                                      'revision_origin_id')),
-             tests=dict(id=('origin', 'origin_id'),
-                        build_id=('build_origin',
-                                  'build_origin_id'))).items():
-        for id, pair in id_pair_map.items():
-            for obj in data.get(collection, []):
-                obj[id] = obj[pair[0]] + ':' + obj[pair[1]]
-                del obj[pair[0]]
-                del obj[pair[1]]
-
-    # Replace slashes with underscores in resource names
-    for collection, prop_list in \
-        dict(revisions=["patch_mboxes"],
-             builds=["input_files", "output_files"],
-             tests=["output_files"]).items():
-        for obj in data.get(collection, []):
-            for prop in prop_list:
-                for resource in obj.get(prop, []):
-                    resource["name"] = resource["name"].replace("/", "_")
+    # Extract origins into separate fields
+    origin_regex = re.compile(f"^({ORIGIN_PATTERN}):.*")
+    for obj_list_name in VERSION.previous.tree:
+        if obj_list_name:
+            for obj in data.get(obj_list_name, []):
+                obj["origin"] = origin_regex.search(obj["id"]).group(1)
+
+    # Remove origins from revision IDs
+    # Calm down pylint: disable=redefined-builtin,invalid-name
+    def remove_origin(id):
+        return id[id.index(':') + 1:]
+    for revision in data.get("revisions", []):
+        revision["id"] = remove_origin(revision["id"])
+    for build in data.get("builds", []):
+        build["revision_id"] = remove_origin(build["revision_id"])
 
     # Update version
     data['version'] = dict(major=JSON_VERSION_MAJOR,
@@ -564,6 +621,6 @@
 }
 
 VERSION = Version(JSON_VERSION_MAJOR, JSON_VERSION_MINOR, JSON, TREE,
-                  v1.VERSION, inherit)
+                  v2.VERSION, inherit)
 
 __all__ = ["VERSION"]
```

gctucker:
On a related note, how about renaming `git_repository_commit_hash` to `git_commit_hash`?

Each hash is really bound to each commit rather than the repository, and commits only exist in repositories.  If you want to make it clear that is'a base commit before applying patches (i.e. not a locally constructed commit), then maybe `git_base_commit_hash` would be clearer.

kcidb/io/schema/v3.py
```diff
@@ -0,0 +1,626 @@
...
+        "id": {
+            "type": "string",
+            "description":
+                "Revision ID.\n"
+                "\n"
+                "Must be a string containing the following, in order.\n"
+                "\n"
+                "The URL of the Git repository which contains the base code "
+                "of the revision. The shortest possible https:// URL, or, if "
+                "that's not available, the shortest possible git:// URL."
```
gctucker, line 103: I think this can be a bit fragile, as trees move from time to time.  In the same way that we have a catalogue of test suites, having a catalogue of trees with simple aliases to be able to refer to them indirectly would avoid this issue, and also make it easier to understand for humans.

See the list of trees in [`build-configs.yaml`](https://github.com/kernelci/kernelci-core/blob/master/build-configs.yaml#L5).

kcidb/io/schema/v3.py
```diff
@@ -0,0 +1,626 @@
...
+                "\n"
+                "The '@' character.\n"
+                "\n"
+                "The full commit hash of the revision's base code in the Git "
+                "repository.\n"
```
gctucker, line 108: It would seem wise to specify that the commit has should be SHA-1 since there are talks of using SHA-256.  Otherwise, if different sources use different SHA versions then we won't be able to join the data (unless we have a way to determine SHA-1 / SHA-256 commit equivalents, which would seem over-the-top).

kcidb/io/schema/v3.py
```diff
@@ -0,0 +1,626 @@
...
+
+# A regular expression pattern matching strings containing sha1 hashes
+SHA1_PATTERN = "[0-9a-f]{40}"
+
+# A regular expression pattern matching strings containing sha256 hashes
+SHA256_PATTERN = "[0-9a-f]{64}"
+
+# A regular expression pattern matching strings containing Git repository
+# commit hash (sha1 or sha256)
+GIT_REPOSITORY_COMMIT_HASH_PATTERN = f"({SHA1_PATTERN}|{SHA256_PATTERN})"
```
gctucker, line 25: See my other comment below - I think we need to only accept SHA-1.

spbnick:
> On a related note, how about renaming `git_repository_commit_hash` to `git_commit_hash`?
> 
> Each hash is really bound to each commit rather than the repository, and commits only exist in repositories. If you want to make it clear that is'a base commit before applying patches (i.e. not a locally constructed commit), then maybe `git_base_commit_hash` would be clearer.

Makes sense, posted #85.

kcidb/io/schema/v3.py
spbnick: I agree, we won't handle renames this way, but then the impact of losing tree continuity doesn't seem too high right now, while complicating the submission process. I think the time for such a feature is probably when we got at least a few submitting CI systems and they all feel this is needed.

spbnick: At the same time, we *can* have the alias table for use in report WebUI to shorten those URLs.

kcidb/io/schema/v3.py
spbnick: I think if the kernel development switches to SHA-256, then all the forks will have to switch, wouldn't they? And then we won't have to join any data.

spbnick:
> Following my previous comment about being database-agnostic, and "multi-field keys", would there be a way to hide the fact that in this implementation the unique fields are being glued together in the IDs?
> 
> I think it would mean not creating the IDs before submitting the data, which can complicate things when creating references. But maybe we can do something with either nested data (i.e. put builds inside revisions, and tests inside builds...) or make it easier to submit data in several steps and get the ID back each time (i.e. submit a revision, get the ID and use it when submitting builds, get the IDs and use them when submitting test results). Neither solution is ideal, but it feels like some aspects in this schema are driven by the design and particularities of BQ. It's probably hard to avoid completely in this PoC project but if we can keep things more engine-agnostic and generic, we should.

I think simplicity of submission process is a great asset. Not having a back-n-forth of "submit data - get ID - add ID to other data" helps a lot. Not having automatically-generated globally-unique IDs is a BigQuery limitation (as well as a limitation of a bunch of other distributed databases). However, it would still work fine with a traditional, centralized SQL database. So, with this approach we are actually more database-agnostic. Yes, the design is driven a lot by BigQuery limitations, but it brought the submission simplicity and greater database independence.

Please note as well that the proposed revision ID also includes the hash over patches, which are not included in any submitted field.

kcidb/io/schema/v3.py
khilman: Agreed.  We've had some trees move several times, and keeping the same "short name" (aka tree-name) with a changing URL has been very useful for the users.

I find the full URL in the current grafana UI cumbersome, and hard to scan visually (it also takes up too much screen real estate.)

gctucker: @khilman Do you agree with using tree names at submission time or the web UI based approach (which doesn't prevent against moving trees)?

spbnick: We can have one alias per multiple repo URLs to handle moving trees in the UI.
I want to avoid forcing all the CI systems to consult the mapping before they can push the data.

khilman: I think tree-names at submission time makes the most sense.  

To not overburden new CI systems, that don't know about tree-names, kcidb can manage the look-up table (ideally using existing `build-configs.yaml`).  URLs that don't have an entry in the table, can be flagged/fixed.

kcidb/io/schema/v3.py
gctucker: I'm not entirely sure but I think git would allow doing a pull from a repository which doesn't have SHA-256 hashes into one that does.  I guess the real question is whether new versions of git will keep both SHA-1 and SHA-256 hashes for a same commit, in which case we could have a problem.  I believe it won't need to do that in which case we're probably fine with allowing both.  It would still be good to actually prove that before allowing SHA-256 imho.

spbnick: Alright, allowing SHA-256 might be a little premature and we can always permit it later, without even bumping the major version number of the schema. Removed.

kcidb/io/schema/v3.py
spbnick: @khilman, I respectfully disagree. Tree names are subjective and debatable. Repo URLs are objective and straight-forward. They're simply places where you got the commit from. You **have** to specify them to Git to get your code to test. They are a hard reality. You don't have to know the tree name **at all** to get the code, build it, and run tests. They're simply a sugar coating. Having them in the data itself instead of URLs **will** complicate things.

So far we managed to avoid the need for one-way data transformation when submitting. Right now what you submit is **exactly** what you get back (if not, it should be fixed). This simplifies thinking about the system greatly. Doing a fixup on repo URLs at submission time will make stored and returned data different from submitted data. If we also put tree names into IDs, we'll have to fixup them too (both in revisions and in builds), further complicating the system. And if we don't, repo moving still won't be handled.

If we allow data transformation on submission there will be no going back. We still have a lot of problems to solve and we could do without having to think about **that**.

What we can do is have a URL->name table in the same dataset, provide command-line tools for modifying it and/or syncing it with build-configs.yaml, outside the normal report submission process (e.g. for administrators), and use that table when displaying the WebUI, and generating reports. We could use that table when making queries as well. I think I could implement that relatively quickly and we would have short tree names in the WebUI and in reports. That would handle repo moving too.

However, I think there's no place for tree names in the report data itself.

spbnick: BTW, we could put the tests.yaml into the dataset at the same time, as it would be similar, and then we could raise warnings on submission if a test isn't found. Same with repo URLs: we could raise warnings if a repo URL has no tree name assigned to it on submission.

gctucker: No need to transform the data, just submit both a URL and a tree name.  You can get that back exactly as well.

spbnick: @gctucker, I was thinking about that earlier, and gave it more thought now. This would work, provided we don't use the tree name as an identifier, but just as another property of the revision to display, or e.g. to filter by. Like this:

image

We would have to accept that not everyone will be able to submit those (at least at first), and that incorrect tree names will be submitted. We can provide support for making sure the tree names are alright, though (warning/erroring on submission, if requested).

khilman: Yes, having both is what I meant, and the name is "just info".  The definitive ID should still be the URL, and provide some warnings/errors if a known name is not matching a known URL.

gctucker: Seems like a good compromise between using solely the URLs and enforcing short tree names.

Thinking about it, somewhere in the system, there will need to be some translation mechanism between what "native" data is available on the client side and what kcidb expects.  Test names, tree names, device names, kernel revision names...  We'll need to decide if we leave that up to each client to comply and do their own translation, or if the submitting tool can help by doing some of that automatically for deterministic data.  It seems to me that upstream tree URLs and the short names fall into the deterministic category.

spbnick: @gctucker, yes, this would work for now.

I'm still of the opinion that it's better to submitt pure hard data that needs no transformation, and has no extra data, and then have "dictionaries" translating that data in UI (such as looking up tree names based on repo URLs). However, let's see how this works out and then move towards what makes more sense.

I assume we don't have any more argument about changes in *this* PR and will merge it this evening unless somebody stops me. I will submit a separate PR adding the optional "tree_name" field to the schema next (I have the changes ready locally already).

gctucker: Yes it's up to you to decide when to merge this PR, I guess it's fine with a follow-up to add the tree name fields.

Because data from random sources will be in some unknown format, there is always going to be the need for some translation to happen before data gets saved in the database. Say, CI systems have different names for a same test suite, it will need to get translated so we don't have `LTP`, `ltp`, `Linux Testing Project`, `Ltp`... but only one name to rule them all.  Pushing the "hard data that needs no transformation" doesn't work there.  In the same way, if some CI systems use a `git://` URLs and others `https://` for a same tree, some translation will need to be done somewhere.  It pretty much applies to all the data being sent to a common database, by nature if there are no ways to enforce what these things should be in the original CI systems then they'll all tend to be different and will have to be transformed.

spbnick: @gctucker, yes there's that problem.

For now I'm hoping we'll be able to prevent that by giving the submitters incentives to submit complying data through better correlation and happier developers, and also by denying to submit whatever non-compliant data we can detect, ramping up the latter as we figure out more what we need. I.e. check data we receive is correct, do not admit incorrect data and let the submitters do fixes/mapping on their side.

I want to prevent, or to at least delay submission transformation as much as possible, as it will make the whole system much more obscure and harder to think about. I would rather free up my mind for thinking about more fundamental problems right now, like correlating revisions, describing, identifying and correlating test environments, presenting reports, notifications, and so on. I think we can refine our data later, one way or another.

gctucker: > For now I'm hoping we'll be able to prevent that by giving the submitters incentives to submit complying data

Exactly, and that means having the required translations done on the client side.  We can't remove this problem, unless we adopt a more integrated pipeline mechanism where test labs use the KernelCI builds and definitions with tree names and URLs, test suite names, device names etc... which is out of scope for the moment.

Some translations will _have_ to be done either:
- by the submitter before calling the provided tool to submit results (e.g. before kcidb-submit),
- by the submitting tool itself (e.g. by kcidb-submit),
- upon reception of the results on the server side (e.g. in Google Could Functions),
- or after getting it back from the database (e.g. dashboard and emails).

Where the translation happens and how is a problem that needs to be addressed, and in fact I believe it's the main thing we need to be able to provide a solution for in the context of a common database.  Starting with putting all this on the submitting end is probably easier initially for us but it won't make it more attractive for new CI systems to join.

spbnick: Yes, this has to be done somewhere.

I'm primarily against doing it ourselves on submission, client- or server-side. We'll have to augment/prettify some of the data on reporting side (dashboard/emails), yes, but I think it's reasonable to have the submitters do the mapping specific to their CI system, and it's too much to expect that from us, it simply won't scale. However, we can help them by picking which data is necessary to submit and in which shape.

Finally, I think our discussion is getting meta, too abstract, and we might be talking about different pieces of data and having different things in mind. Let's discuss the specific pieces as we get to them instead.

gctucker: In practice, for this particular case, I think it would make a lot of sense for clients to submit data with plain git URLs and the submitting tool would look up a tree name for that URL.

spbnick: @gctucker, if we can look up a tree name automatically based on the URL, why do we need to store it with each revision at all, why not just look it up when reporting?

gctucker: Because if the tree name is in the database, then you can directly do a query using it rather than "url 1" or "url 2" or "url 3" etc...  We're back to the 4 bullet points a couple of comments ago, I'm suggesting option 2 and you're suggesting option 4.  It's going to be a recurring theme, I guess semantics and the scope where data is meaningful should help us decide where these translations need to happen ideally.  Then some practical limitations may also come into play.

spbnick: We can have the tree names in a separate table, as I proposed earlier, and do a join if we need to query by a tree name. It's not the simplest-possible query, as we would have with the tree name right there in the revision, but it's still basic and is what databases are built for.

Regardless of that, I wonder why would you prefer doing the transformation at submission time (apart from making it easier for submitters)? Perhaps I'm missing something? What's your motivation for this approach in general?

PR #86, opened 2020-05-27
move kdump url to top level

rasibley:
kdump is now a suite of tests as we are onboarding new tests which
are part of kdump, change the url to point to the top level directory
to make it apparent that it's now a suite.


spbnick:
Looks good to me, merging!
Thank you, @rasibley!

Issue #85, opened 2020-05-19
Rename git_repository_commit_hash to git_commit_hash


spbnick:
Rename the revision's `git_repository_commit_hash` property to just have `git_commit_hash` to have easier time typing and reading it.

Quoting @gctucker:
> Each hash is really bound to each commit rather than the repository, and commits only exist in repositories.

PR #90, opened 2020-06-23
io.schema: Remove repo URL from revision IDs

spbnick:


spbnick:
@Inakimalerba, heads up!

spbnick:
@gctucker, do you think you could take a look at this?

kcidb/io/schema/v3.py
```diff
@@ -26,8 +26,7 @@
 
 # A regular expression pattern matching strings containing revision IDs
 REVISION_ID_PATTERN = \
-    f"{GIT_REPOSITORY_URL_PATTERN}" \
-    f"@{GIT_REPOSITORY_COMMIT_HASH_PATTERN}" \
+    f"{GIT_REPOSITORY_COMMIT_HASH_PATTERN}" \
```
gctucker, line 29: It's not new in this PR, but maybe `GIT_COMMIT_HASH_PATTERN` would be better, i.e. without `REPOSITORY` for the same reason as with the `git_commit_hash` field.

spbnick: Yes, makes sense, I could do that as part of the fix for #85.

spbnick:
> So, what happens if a same git sha1 gets tested in 2 different trees? Do they both have a revision entry in the database with the same id but different repo URLs?

Yes. Similar to how we would have duplicate revision entries if two CI systems discovered the same revision.

spbnick:
@gctucker, do you think this can go in, or would you like me to change something?

spbnick:
> I'm not entirely sure of the consequences of having different database entries with the same ID, but that's not specific to this PR.
Yeah, we already have to deal with that, and it certainly will be "interesting" :grin:.
Thank you, merging!

PR #214, opened 2021-10-13
Add version flag to command line tools

### LINKED ISSUE #176: Add support for -v/--version option to command-line tools
As the number of releases and users grows, so grows the version confusion. Add support for `-v`/`--version` options to all kcidb command-line tools to make it easier to identify which version is being used.

amfelso:
Fixes #176, adds support for version flags

### DIFF
--- a/kcidb/misc.py
+++ b/kcidb/misc.py
@@ -8,6 +8,10 @@ import argparse
 import logging
 import json
 from textwrap import indent
+try:  # Python 3.9
+    from importlib import metadata
+except ImportError:  # Python 3.6
+    import importlib_metadata as metadata
 from google.cloud import secretmanager
 import jq
 
@@ -106,6 +110,11 @@ class ArgumentParser(argparse.ArgumentParser):
             kwargs: Keyword arguments to initialize ArgumentParser with.
         """
         super().__init__(*args, **kwargs)
+        self.add_argument(
+            '--version',
+            action='version',
+            version=f"Version {metadata.version('kcidb')}"
+        )
         self.add_argument(
             '-l', '--log-level',
             metavar="LEVEL",
--- a/setup.py
+++ b/setup.py
@@ -44,6 +44,8 @@ setuptools.setup(
         "google-cloud-pubsub<2.0.0dev",
         "google-cloud-firestore",
         "google-cloud-secret-manager<2.0.0dev",
+        # TODO: remove with minimal requirement of Python 3.8
+        "importlib_metadata",
         "jsonschema[format]",
         "requests",
         "pyyaml",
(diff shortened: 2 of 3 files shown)


kcidb/misc.py
```diff
@@ -106,6 +107,10 @@ def __init__(self, *args, **kwargs):
             kwargs: Keyword arguments to initialize ArgumentParser with.
         """
         super().__init__(*args, **kwargs)
+        self.add_argument(
+        '-v', '--version',
+         action='version',                    
+        version="Version % s" % importlib.metadata.version('kcidb'))
```
spbnick, line 113: Could you please use f-strings for formatting the output?
I'm trying to keep that as a convention in KCIDB.

spbnick:
Oh, and don't mind the "check_deployment" failures!

amfelso:
Thank you Nikolai! I'm sitting down now to see if I can make these last changes and will reach out if I need help.

spbnick:
> Thank you Nikolai! I'm sitting down now to see if I can make these last changes and will reach out if I need help.

Awesome! Leave a comment, when you're done, and I'll check it out.

Looks great, so far, BTW. However, I'm curious why the CI didn't fail even though we didn't have the importlib-metadata package in the dependencies :thinking: Perhaps we need to add installation and testing in a virtualenv to the CI. Meanwhile, do you have a way to test in a Python 3.6 virtualenv manually? I don't have Python 3.6 handy.

amfelso:
That's a great question. I'm taking some more time to test it on a virtual environment with Python 3.6 and will let you know the results!

amfelso:
As an update, I did successfully use the version flag on a virtual machine with 3.6.2 and didn't have any issues with the dependency. Do you want me to continue to look into that?

spbnick:
> As an update, I did successfully use the version flag on a virtual machine with 3.6.2 and didn't have any issues with the dependency. Do you want me to continue to look into that?

Thanks a lot! I think one of the packages kcidb depends on depends on `importlib-metadata`, that's why it worked (there's some evidence in CI logs). However, to avoid accidental breakage in case that dependency stops requiring `importlib-metadata`, please add it to dependencies in `setup.py` with a `TODO` comment saying this is to be removed once our minimal requirement is at least Python 3.8.

Apart from this, could you please remove the `-v` option, and just leave `--version`, since this would be rarely used, and we might need `-v` for something else later (like `--verbose` alias).

Otherwise, this looks great. Thanks!

setup.py
```diff
@@ -44,6 +44,7 @@
         "google-cloud-pubsub<2.0.0dev",
         "google-cloud-firestore",
         "google-cloud-secret-manager<2.0.0dev",
+        "importlib_metadata",  # TODO: remove with minimal requirement of Python 3.8
```
spbnick, line 47: Just put this above :)

spbnick: Hey, no problem, that's what internships are for :+1:

PR #218, opened 2021-10-19
Add online help for specifying database drivers

### LINKED ISSUE #186: Implement online help for database drivers
Now that we have support for multiple database drivers, and each of them can have their own format for parameters, we need to add a command-line option to each command accepting database parameters, which would output online help message describing all available drivers and format of their parameters.

We have the beginning of that in the `PARAMS_DOC` class variable for each driver, which contains the description of its parameters. We need to add support for the option to each applicable command, and put those together into a coherent message.

amfelso:
Fixes #186 by providing users with information on driver options and associated params.

### DIFF
--- a/kcidb/db/__init__.py
+++ b/kcidb/db/__init__.py
@@ -2,6 +2,7 @@
 
 import sys
 import logging
+import argparse
 import kcidb_io as io
 import kcidb.orm
 import kcidb.misc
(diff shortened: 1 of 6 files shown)


kcidb/db/__init__.py
```diff
@@ -270,8 +270,11 @@ def argparse_add_args(parser, database=None):
     assert database is None or isinstance(database, str)
     parser.add_argument(
         '-d', '--database',
-        help=('Database specification' if database is None
-              else f"Database specification. Default is {database!r}."),
+        help=("""DATABASE specification, formatted as <DRIVER>:<PARAMS>.
+        Driver options are bigquery:<PROJECT_ID>.<DATASET>,
+        sqlite:<DATABASE>, json:[FILE] and null.""" +
+              "" if database is None
+              else f"Default is {database!r}."),
```
spbnick, line 277: Do you think you could generate the message from the actual list of database drivers in `DRIVER_TYPES` instead?

amfelso: Absolutely!

kcidb/db/__init__.py
```diff
@@ -259,6 +260,30 @@ def complement(self, data):
...
+            option_strings=option_strings,
+            dest=dest,
+            default=default,
+            nargs=0,
+            help=help)
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        print("Specify database as <DRIVER>:<PARAMS>. " +
+              "Driver options and associated params (if any) " +
+              "are listed below.\n")
```
spbnick, line 280: This help message isn't really about the `-d`/`--database` option, but rather about its value, so we shouldn't start with "Specify database as <DRIVER>:<PARAMS>.", but rather with a general introduction, the fact that there are multiple databases supported, and the format of the specification string. So, something like this:

```suggestion
        print("KCIDB has several database drivers for both actual and "
              "virtual databases.\n" 
              "You can specify a particular driver to use, and its "
              "parameters, using the\n"
              "-d/--database option.\n"
              "\n"
              "The format of the option value is <DRIVER>[:<PARAMS>], "
              "where <DRIVER> is the\n"
              "name of the driver, and <PARAMS> is a (sometimes optional) "
              "driver-specific\n"
              "parameter string.\n"
              "\n"
              "For example, \"-d bigquery:kernelci-production.kcidb_01\" "
              "requests the use of\n"
              "the \"bigquery\" database driver with the parameter string\n"
              "\"kernelci-production.kcidb_01\", from which the driver "
              "extracts the Google\n"
              "Cloud project \"kernelci-production\" and the dataset "
              "\"kcidb_01\" to connect to.\n"
              "\n"
              "Available drivers and format of their parameter strings "
              "follow.\n")
```

kcidb/db/__init__.py
```diff
@@ -259,6 +260,30 @@ def complement(self, data):
...
+        for title, driver in DRIVER_TYPES.items():
+            print(title.upper() + " DRIVER \nPARAMS:\n" +
+                  driver.PARAMS_DOC)
```
spbnick, line 283: Calling the iterator variable `title` is a little imprecise. This is a driver's name, and we're *using* it to make a section title, so it's better to call it just `name`.

Then, at the moment, driver names accepted by the `-d`/`--database` option are case-sensitive (see `kcidb.db.Client.__init__()`). So we should be careful with making the only mention of a driver's name uppercase in our help message, because people could assume it should be supplied as such. We should either keep it the original case, or explicitly mention how it should be spelled. I'd say for now we can just print the driver headers something like this:
```
"json" driver
-------------
```
Further on, I think it would be good to give a brief description of each driver and its purpose, not just its parameters, so I think we need to switch from having parameter-specific docs in the drivers in `PARAMS_DOC`, to just general-purpose description including the parameter description in e.g. `DOC`.

In the end this loop could be changes something like this:
```suggestion
        for name, driver in DRIVER_TYPES.items():
            print(f"\n{name!r} driver\n" + 
                  "-" * (len(name) + 7) + "\n" +
                  driver.DOC)
```

PR #223, opened 2021-10-28
Add single function for querying BigQuery

### LINKED ISSUE #157: bigquery: Add and use a single function for queries
There are multiple invocations of `self.client.query()` throughout the BigQuery DB driver,  and each of them has similar setup and could use logging.

To reduce code duplication and simplify maintenance implement a single querying function to be called by dump()/query()/oo_query()/etc.. Add DEBUG-level logging of queries and their parameters.

amfelso:
Fixes #157


spbnick:
Looks great! Thank you, August, merging :smiley:

Issue #68, opened 2020-04-17
Generate notification messages with HTML version


spbnick:
Some e-mail viewing software may choose to display our notification messages in variable-width fonts, messing up the formatting. The most prominent example is groups.io which we use to host the maillist with test reports. Consider adding an HTML version of the report, being just the text version wrapped in `<pre></pre>` element.

spbnick:
Fixed in #224.

PR #224, opened 2021-10-28
Generate notification messages with HTML version

effulgentstar:
#68 Generate notification messages with HTML version


kcidb/monitor/test_output.py
```diff
@@ -74,4 +74,15 @@ def test_min(self):
         self.assertEqual(message['X-KCIDB-Notification-ID'], notification.id)
         self.assertEqual(message['X-KCIDB-Notification-Message-ID'], "id")
         self.assertIn("Revision detected: ", message['Subject'])
-        self.assertIn("We detected a new revision!", message.get_payload())
+        text, html = message.get_payload()
+        self.assertEqual('text/plain', text.get_content_type())
+        self.assertEqual('utf-8', text.get_content_charset())
+        content = text.get_payload()
+        if text['content-transfer-encoding'] != '7bit':
```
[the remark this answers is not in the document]
effulgentstar: Because of this: https://stackoverflow.com/questions/25710599/content-transfer-encoding-7bit-or-8-bit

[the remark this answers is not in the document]
effulgentstar: Should I take it out?

effulgentstar:
Hopefully, this is it.... 🤞

kcidb/monitor/test_output.py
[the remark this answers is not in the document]
spbnick: Regardless of whichever transfer encoding the Python's email library chooses for us (since we don't specify it explicitly), we should have the "We detected a new revision!" text in there. So we shouldn't skip that check if it's 7-bit transfer encoding, but should check it in both cases. Latin letters and basic punctuation will be represented as is in either 7-bit and 8-bit transfer encoding (because their character codes fit 7-bit, because 7-bit encoding was created to accommodate them), so the check should work in both cases.

So, yes, please remove that condition here and in the other spot, because we want to check the message contents in either case. Thank you!

kcidb/monitor/test_output.py
```diff
@@ -74,4 +74,15 @@ def test_min(self):
         self.assertEqual(message['X-KCIDB-Notification-ID'], notification.id)
         self.assertEqual(message['X-KCIDB-Notification-Message-ID'], "id")
         self.assertIn("Revision detected: ", message['Subject'])
-        self.assertIn("We detected a new revision!", message.get_payload())
+        text, html = message.get_payload()
+        self.assertEqual('text/plain', text.get_content_type())
+        self.assertEqual('utf-8', text.get_content_charset())
+        content = text.get_payload()
+        if text['content-transfer-encoding']:
```
spbnick, line 81: We don't really need to check if that header is there, it's pretty much guaranteed. Could you remove these checks?

kcidb/monitor/output.py
```diff
@@ -136,9 +138,24 @@ def render(self):
...
+        subject = self.message.summary + self.obj.summarize()
+        email["Subject"] = subject
         email["To"] = ", ".join(self.message.recipients)
         email["X-KCIDB-Notification-ID"] = self.id
         email["X-KCIDB-Notification-Message-ID"] = self.message.id
-        email.set_content(self.message.description + self.obj.describe())
+        message = self.message.description + self.obj.describe()
+        email.set_content(message)
+        formatted_subject = html.escape(subject)
+        formatted_message = html.escape(message)
```
spbnick, line 149: We shouldn't call these `formatted`, as that would be confusing (formatted usually means we used some sort of template to produce it), but instead call it e.g. `escaped`. Could you change that?

kcidb/monitor/test_output.py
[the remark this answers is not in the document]
effulgentstar: I did in the last commit

spbnick (changes requested):

kcidb/monitor/output.py
effulgentstar: Okay

kcidb/monitor/output.py
```diff
@@ -136,9 +138,24 @@ def render(self):
...
+        html_msg = (textwrap.dedent(f"""
+                        <html>
+                            <head>
+                                <title>{formatted_subject}</title>
+                            </head>
+                            <body>
+                                <pre>{formatted_message}</pre>
+                            </body>
+                        </html>
+                    """))
```
spbnick, line 159: Right, `textwrap.dedent()` wouldn't work like this, because it tries to "dedent" a text which has lines starting at the first column, because we've substituted the message, which has newlines in there with text starting at the first column.
So we gotta first "dedent" the format string, and then substitute our subject and message.

Could you switch this to first call `textwrap.dedent()` and then use the `format()` method on the resulting string?
Oh, and also, could you please remove the extra parentheses around the expression?

effulgentstar:
I made the changes you requested

spbnick:
Just two minor comments, and please fix the tests failing in CI (you can test locally if you just run "pylint" in the top directory). Otherwise I think we're ready for the next step: marking up links with `<a></a>`!

effulgentstar:
@spbnick I used tornado.escape to add anchor tags to the urls in the message, but it's not a standard module, so pylint is raising an unable to import error

kcidb/monitor/output.py
```diff
@@ -136,9 +139,26 @@ def render(self):
...
+        email["Subject"] = subject
         email["To"] = ", ".join(self.message.recipients)
         email["X-KCIDB-Notification-ID"] = self.id
         email["X-KCIDB-Notification-Message-ID"] = self.message.id
-        email.set_content(self.message.description + self.obj.describe())
+        message = self.message.description + self.obj.describe()
+        email.set_content(message)
+        escaped_subject = html.escape(subject)
+        escaped_message = html.escape(message)
+        escaped_message = tor.linkify(escaped_message)
```
spbnick, line 151: Adding Tornado as a (huge) dependency just for adding links to our notifications is not worth the trouble.
How well do you know your regular expressions? Can you use `re.sub()` to do this? We don't need anything fancy at this moment, just `http://`, `https://`, `git://` and maybe `ftp://` URLs.

spbnick (changes requested):
Somehow I keep forgetting to submit review comments, so here's a new one, and two I comments I lost before. Thank you for your persistence :smiley:

kcidb/monitor/output.py
effulgentstar: Okay

kcidb/monitor/output.py
```diff
@@ -136,9 +138,26 @@ def render(self):
...
+                        <html>
+                            <head>
+                                <title>{}</title>
+                            </head>
+                            <body>
+                                <pre>{}</pre>
+                            </body>
+                        </html>
+                    """)
+        html_msg = html_msg.format(escaped_subject, escaped_message)
```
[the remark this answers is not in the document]
effulgentstar: Okay, I will

kcidb/monitor/output.py
effulgentstar: Fixed using re.

kcidb/monitor/output.py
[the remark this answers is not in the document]
effulgentstar: Just saw the comment removing the newline. I have made the change

kcidb/monitor/output.py
```diff
@@ -136,9 +138,26 @@ def render(self):
...
         email["To"] = ", ".join(self.message.recipients)
         email["X-KCIDB-Notification-ID"] = self.id
         email["X-KCIDB-Notification-Message-ID"] = self.message.id
-        email.set_content(self.message.description + self.obj.describe())
+        message = self.message.description + self.obj.describe()
+        email.set_content(message)
+        escaped_subject = html.escape(subject)
+        escaped_message = html.escape(message)
+        urls = re.compile(r'((http|https|git|ftp)?://[^|\s]+)')
+        escaped_message = urls.sub('<a href= "\\1">\\1</a>', escaped_message)
```
spbnick, line 151: This looks about right, however:
* There's not much sense to compile a regular expression separately, if we're not going to reuse the compiled regex, so could you just pass it to re.sub directly?
* Why did you put `?` after the `(http|https|git|ftp)`? URL wouldn't be valid without a schema, wouldn't it?
* Why is there `|` inside `[^|\s]` specifically? What case are you trying to handle?
* The [`html.escape()`](https://docs.python.org/3/library/html.html#html.escape) function does not escape quotes by default. Therefore it is not safe to put our substitution into the `href` attribute value as is. One easy fix, I think, could be to pass `quote = True` to both `html.escape()` invocations and simply let quotes be escaped everywhere.
* I think you can avoid the outside parentheses in the pattern, and substitute the whole match with `\g<0>` instead, but the latter is quite obscure, so I'm not sure if that would be better :smile:

effulgentstar: The | in [^|\s] is a typo.. Its [^\s]

kcidb/monitor/output.py
```diff
@@ -136,9 +138,26 @@ def render(self):
...
+                        <html>
+                            <head>
+                                <title>{}</title>
+                            </head>
+                            <body>
+                                <pre>{}</pre>
+                            </body>
+                        </html>
+                    """).format(escaped_subject, escaped_message)
+        email.add_alternative(html_msg, subtype='html')
```
spbnick, line 162: While you still haven't submitted an amended version, could you please reduce the indent of the string so the closing quote is on the same column as the start of the statement? E.g.:
```suggestion
        html_msg = textwrap.dedent("""\
            <html>
                <head>
                    <title>{}</title>
                </head>
                <body>
                    <pre>{}</pre>
                </body>
            </html>
        """).format(escaped_subject, escaped_message)
        email.add_alternative(html_msg, subtype='html')
```

effulgentstar: @spbnick, I have made the changes you requested.

kcidb/monitor/output.py
```diff
@@ -136,9 +138,25 @@ def render(self):
...
+        email["Subject"] = subject
         email["To"] = ", ".join(self.message.recipients)
         email["X-KCIDB-Notification-ID"] = self.id
         email["X-KCIDB-Notification-Message-ID"] = self.message.id
-        email.set_content(self.message.description + self.obj.describe())
+        message = self.message.description + self.obj.describe()
+        email.set_content(message)
+        escaped_subject = html.escape(subject, quote=True)
+        escaped_message = html.escape(message, quote=True)
+        escaped_message = re.sub(r'((http|https|git|ftp)://[^\s]+)', escaped_message)
```
spbnick, line 150: Thanks! The regular expression looks good to me now. However, this invocation of re.sub() won't work correctly. Check the [usage docs](https://docs.python.org/3/library/re.html#re.sub).

Also you can make pylint happy if you wrap this line.

effulgentstar: Arh.... I forgot to put back the repl. Will do so now

effulgentstar: @spbnick, I made changes. Please review

kcidb/monitor/output.py
```diff
@@ -136,9 +138,26 @@ def render(self):
...
         email["To"] = ", ".join(self.message.recipients)
         email["X-KCIDB-Notification-ID"] = self.id
         email["X-KCIDB-Notification-Message-ID"] = self.message.id
-        email.set_content(self.message.description + self.obj.describe())
+        message = self.message.description + self.obj.describe()
+        email.set_content(message)
+        escaped_subject = html.escape(subject, quote=True)
+        escaped_message = html.escape(message, quote=True)
+        escaped_message = re.sub(r'((http|https|git|ftp)://[^\s]+)'
+                        , '<a href= "\\1">\\1</a>', escaped_message)
```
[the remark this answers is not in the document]
effulgentstar: I have committed the changes for PR.

Kindly review it. Thanks

spbnick:
Almost perfect :smiley: Please deal with pylint warnings that CI has found, and in general, in the future just make sure CI passes (except the `check_deployment` step for now) before another review round. Thanks!

effulgentstar:
@spbnick I made changes to the test_output.py to allow for the html alternative. The file pytest is currently using to make checks keeps raising a match case error.

spbnick:
Yes, there's a problem with another test, and you can fix it similarly to the way you did the other one. You can run the tests locally using `pylint` to help you debug that. There's a lot of debug output, and that's something we need to fix, but if you scroll through it, you can eventually find the source of the error. You can make it produce a little less output, if you pass `--disable-warnings --tb=native` to `pylint`.

effulgentstar:
@spbnick, I made changes to the test_monitor.py, kindly review

spbnick:
Looks great, and nice to see all checks green :) Let me run a few tests locally as a final test...

kcidb/test_monitor.py
```diff
@@ -198,5 +198,15 @@ def test_min(self):
...
+            content = text.get_payload()
             self.assertIn(f"Test {obj_type_name} detected!\n\n",
-                          message.get_payload())
+                          content)
+            self.assertEqual('text/html', html.get_content_type())
+            self.assertEqual('utf-8', html.get_content_charset())
+            content = html.get_payload()
+            self.assertIn(f"Test {obj_type_name} detected!\n\n",
+                          content)
+
```
spbnick, line 212: Could you remove the spaces from this line?

spbnick (changes requested):
Everything works and looks great apart from that one line I mention in the inline comment.
Otherwise, ready to go :)

kcidb/test_monitor.py
effulgentstar: @spbnick I am not sure which line you are referring to. Is it 209 or 210?

spbnick: The last line in the file.

effulgentstar: Oh, okay.. Kindly review the change.

spbnick: Thanks, it's perfect now :)

spbnick:
Everything looks great now, and it works as it should. Merging!
Congrats, @effulgentstar, and thank you for your persistence :grinning:

effulgentstar:
Thank you @spbnick

Issue #79, opened 2020-04-27
Consider supporting upgrading to a specific version with kcidb-upgrade


spbnick:
The `kcidb-upgrade` tool can be used for upgrading the I/O data schema. However, it only supports upgrading from any known version to the latest version. Sometimes, though, it could be useful to upgrade to an older version. E.g. to see what an older version of KCIDB would do, or to debug an upgrading issue.

Consider adding support for a command-line option accepting the major version number of a target schema to kcidb-upgrade. The option should default to the latest version, if unspecified.

spbnick:
This could help with debugging upgrades.

spbnick:
Or, perhaps, make the target version an optional positional argument.

pgadige:
I have a query, @spbnick Could you please help me understand this specific use case.

When I ran `kcidb-upgrade` tool with no arguments, I saw a blinking cursor on a new line. The cursor continued to blink for several minutes ( > 30 minutes). I explicitly exited using KeyBoardInterrupt Ctrl + C. I wonder if this is an expected behaviour of `kcidb-upgrade` tool. Intuitively, I expected an output indicating either successful upgrade of I/O JSON data to the latest version or inform the current I/O JSON data is already the latest version. Since running `kcidb-upgrade` with no arguments doesn't complain, I assume that the usage of command is correct. Am I missing something? 

An [image](https://pasteboard.co/rfsN27vQurqT.png) of what I see on my screen.

PR #228, opened 2021-11-02
Upgrade to previous or latest I/O data schema

pgadige:
This pull request fixes the issue #79 
Add an optional positional CLI argument to `kcidb-upgrade` tool for upgrading to a valid JSON `target-schema` I/O data schema. The option has a default value of `4`, which is the latest JSON I/O data schema according to `kcidb-io` module.

### DIFF
--- a/kcidb/__init__.py
+++ b/kcidb/__init__.py
@@ -259,13 +259,22 @@ def validate_main():
 def upgrade_main():
     """Execute the kcidb-upgrade command-line tool"""
     sys.excepthook = misc.log_and_print_excepthook
-    description = 'kcidb-upgrade - Upgrade I/O JSON data to latest schema'
+    description = 'kcidb-upgrade - Upgrade I/O JSON data to a valid previous schema or latest schema'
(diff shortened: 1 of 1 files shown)


spbnick:
Thank you, @pgadige, this is a good start. However, let's not change the `kcidb_io.schema.upgrade()` function interface and keep it simple for now (just upgrading to the latest schema), until we see more important use cases.

For now you can just lookup the target schema like this:
```python
version = kcidb_io.schema.LATEST
while version and version.major != major:
    version = version.previous
if not version:
    raise Exception("No version found for major number {major}")
```
You can also generate the list of available schema versions for the help message in a similar way.

Then you can upgrade to the specific version by calling the schemas [`upgrade()` method](https://github.com/kernelci/kcidb-io/blob/30d145ecdd8c3158c86b63a5ab5794ae0671f38c/kcidb_io/schema/misc.py#L199):
```python
data = version.upgrade(data)
```
All this shouldn't require changes to kcidb-io.

spbnick:
Thank you for your contribution @pgadige. I've implemented this myself in #236

Issue #76, opened 2020-04-27
Make kcidb-schema output older schema versions on request


spbnick:

spbnick:
The `kcidb-schema` tool available after installing KCIDB supports outputting the latest version of the schema only. Like this:

```console
$ kcidb-schema
{   
    "title": "kcidb",
    "description": "Kernel CI report data. ...",
    "type": "object",
    "properties": {
        "version": {
            "type": "object",
            "properties": {
                "major": {
                    "type": "integer",
...
```

Make it accept a positional argument specifying the major version of the schema to output JSON for. Like this:
```console
$ kcidb-schema 4
{   
    "title": "kcidb",
    "description": "Kernel CI report data. ...",
    "type": "object",
    "properties": {
        "version": {
            "type": "object",
            "properties": {
                "major": {
                    "type": "integer",
...
```

spbnick:
Fixed in #231

PR #231, opened 2021-11-03
Make Kcidb-schema output older schema versions on request

effulgentstar:
#76 Make Kcidb-schema output older schema versions on request. @spbnick , kindly review

### DIFF
--- a/kcidb/__init__.py
+++ b/kcidb/__init__.py
@@ -252,11 +252,26 @@ def query_main():
 def schema_main():
     """Execute the kcidb-schema command-line tool"""
     sys.excepthook = misc.log_and_print_excepthook
-    description = 'kcidb-schema - Output latest I/O JSON schema'
+    description = 'kcidb-schema - Output latest or older I/O JSON schema'
     parser = misc.OutputArgumentParser(description=description)
+    version = io.schema.LATEST
+    parser.add_argument(
+        'schema_version',
+        metavar='SCHEMA_VERSION',
+        type=int,
+        help='Major version number of the schema to output. '
+             f'Default is the latest version (currently {version.major})',
+        nargs='?',
+        default=version.major
+    )
     args = parser.parse_args()
-    misc.json_dump(io.schema.LATEST.json, sys.stdout,
-                   indent=args.indent, seq=args.seq)
+    major = args.schema_version
+    while version and version.major != major:
+        version = version.previous
+    if not version:
+        raise Exception(f"No version found for major number {major}")
+    misc.json_dump(version.json, sys.stdout, indent=args.indent,
+                   seq=args.seq)
 
 
 def validate_main():


kcidb/misc.py
```diff
@@ -187,6 +208,14 @@ def argparse_output_add_args(parser):
...
         action='store_true'
     )
+    parser.add_argument(
+        'schema_version',
+        metavar='Schema Version',
+        type=schema_versions,
+        help='KCIDB schema version. Default is LATEST',
+        nargs='?',
+        default="LATEST",
+    )
```
spbnick, line 218: This wouldn't be the right place to add the argument specification, as this function is used to setup command-line parsing for commands outputting **any** JSON, not just the schema.

In general, it's good practice to look for all ways the function is used before making changes like that, to not accidentally introduce unwanted behavior.

kcidb/misc.py
```diff
@@ -165,6 +165,27 @@ def non_negative_int(string):
...
+
+    Raises:
+        argparse.ArgumentTypeError: the value is not a schema version.
+    """
+    if value == 'LATEST':
+        return value
+    if not re.match("^[1-4]+$", value):
+        raise argparse.ArgumentTypeError(
+            f'{repr(value)} is not a previous schema version')
+    return "V"+value
```
spbnick, line 186: Nice job figuring out how to do this, but let's do it the simple way first.
Just add the argument definition inside the `schema_main()`.

You can build a list of available schema versions using a loop like this:
```python
schema_major_versions = []
version = kcidb_io.schema.LATEST
while version:
    schema_major_versions.append(version.major)
    version = version.previous
```
Then you can use it to define available choices for the argument values. We can start with the absent argument meaning the latest version, without an explicit "LATEST" value. Also, no need to require the "V" to be present there, just the number is enough and creates less hassle.

Once you get your `major` version value from the argument, you can lookup the schema like this:
```python
version = kcidb_io.schema.LATEST
while version and version.major != major:
    version = version.previous
if not version:
    raise Exception("No version found for major number {major}")
```

spbnick (changes requested):
Good start, @effulgentstar, but let's simplify it a little :grin:

effulgentstar:
@spbnick.. I made changes. Kindly review

effulgentstar:
@spbnick.. I made changes. Kindly review

kcidb/__init__.py
```diff
@@ -252,11 +252,31 @@ def query_main():
...
     """Execute the kcidb-schema command-line tool"""
     sys.excepthook = misc.log_and_print_excepthook
-    description = 'kcidb-schema - Output latest I/O JSON schema'
+    description = 'kcidb-schema - Output latest or older I/O JSON schema'
     parser = misc.OutputArgumentParser(description=description)
+    schema_major_versions = []
+    version = io.schema.LATEST
+    while version:
+        schema_major_versions.append(version.major)
+        version = version.previous
```
spbnick, line 261: I see you're collecting the version numbers into `schema_major_versions` and then not using it. What was your intention here?

kcidb/__init__.py
```diff
@@ -252,11 +252,31 @@ def query_main():
...
+    parser.add_argument(
+        'schema_version',
+        metavar='Schema Version',
+        type=int,
+        help='KCIDB schema version. Default is 4',
+        nargs='?',
+        default=4,
```
spbnick, line 268: Instead of hard-coding the latest version number in here, which will require updating later (and could be forgotten), could you just leave the default unspecified, and use the latest version, if the argument is not set?

Or, maybe as a better alternative, could you take the actual latest version number from `json.schema.LATEST` and put it here and in the help message?

spbnick (changes requested):
You're on the right path, @effulgentstar, but I have a couple comments/requests. Thank you.

kcidb/__init__.py
effulgentstar: @spbnick kindly review

kcidb/__init__.py
```diff
@@ -252,11 +252,26 @@ def query_main():
...
+        type=int,
+        help=f'KCIDB schema version. Default is {version.major}',
+        nargs='?',
+    )
     args = parser.parse_args()
-    misc.json_dump(io.schema.LATEST.json, sys.stdout,
-                   indent=args.indent, seq=args.seq)
+    major = args.schema_version
+    if major is None:
+        major = version.major
```
spbnick, line 268: Actually, if you just specify `version.major` as the `default` for the argument you're adding, you won't have to handle it here.

kcidb/__init__.py
```diff
@@ -252,11 +252,26 @@ def query_main():
...
     sys.excepthook = misc.log_and_print_excepthook
-    description = 'kcidb-schema - Output latest I/O JSON schema'
+    description = 'kcidb-schema - Output latest or older I/O JSON schema'
     parser = misc.OutputArgumentParser(description=description)
+    version = io.schema.LATEST
+    parser.add_argument(
+        'schema_version',
+        metavar='Schema Version',
+        type=int,
+        help=f'KCIDB schema version. Default is {version.major}',
```
spbnick, line 262: Could you be more specific about what the schema version would be used for, here? Also, using `KCIDB` here is a little redundant. Finally, it would be good to mention that the default is always the latest version, so people don't have to wonder what would happen after an upgrade.
Overall, I think this would be better:
```suggestion
        help=f'Major version number of the schema to output. '
        'Default is the latest version (currently {version.major})',
```

spbnick (changes requested):
Thank you, @effulgentstar, this works perfectly well already! I have three mostly cosmetic comments, after addressing which, I think we can merge this.

Sorry it took me a while to get around to this. I got a flu vaccine shot yesterday and am taking a break today, feeling fuzzy and unable to concentrate.

effulgentstar:
@spbnick, hope you are feeling much better. I made changes. This: Run kcidb-tests-validate --urls < tests.yaml
HTTPError: 502 Server Error: Bad Gateway for url: https://github.com/SELinuxProject/selinux-testsuite is failing

spbnick:
Looks and works great, thank you, @effulgentstar!
The Bad Gateway was a GitHub problem and it worked after retrying.
Merging!

effulgentstar:
Thank you @spbnick for your guidance.

PR #232, opened 2021-11-03
Use I/O schema with $defs

spbnick:
Switch to using the I/O schema with resource/checkout/build/test defined
inside the top-level $defs property and referenced from there, instead
of duplicated throughout the schema.

Concerns kernelci/kcidb-io#11


spbnick:
Heads-up, @mrbazzan

PR #239, opened 2021-11-20
Consider making kcidb-validate output validated reports in kcidb-validate  #145

### LINKED ISSUE #145: Consider making kcidb-validate output validated reports
It could be useful to make `kcidb-validate` output reports after they were validated, similar to other validation tools. This could help the user see which exact report was valid and which wasn't, when it aborts on validation failure.

mharyam:
Close #145 

- Add print to kcidb-validate to output reports after they were validated

### DIFF
--- a/kcidb/__init__.py
+++ b/kcidb/__init__.py
@@ -250,11 +250,16 @@ def validate_main():
     """Execute the kcidb-validate command-line tool"""
     sys.excepthook = misc.log_and_print_excepthook
     description = 'kcidb-validate - Validate I/O JSON data'
-    parser = misc.ArgumentParser(description=description)
+    parser = misc.OutputArgumentParser(description=description)
     misc.argparse_schema_add_args(parser, "validate against")
     args = parser.parse_args()
-    for data in misc.json_load_stream_fd(sys.stdin.fileno()):
-        args.schema_version.validate(data)
+    misc.json_dump_stream(
+        (
+            args.schema_version.validate(data)
+            for data in misc.json_load_stream_fd(sys.stdin.fileno())
+        ),
+        sys.stdout, indent=args.indent, seq=args.seq
+    )
 
 
 def upgrade_main():
--- a/test_kcidb.py
+++ b/test_kcidb.py
@@ -169,11 +169,15 @@ class KCIDBMainFunctionsTestCase(kcidb.unittest.TestCase):
 
     def test_validate_main(self):
         """Check kcidb-validate works"""
+        empty = json.dumps(dict(version=dict(major=4, minor=0))) + "\n"
+        second_empty = json.dumps(dict(version=dict(major=3, minor=0))) + "\n"
+
         self.assertExecutes('', "kcidb.validate_main")
-        self.assertExecutes('{"version":{"major":4,"minor":0}}',
-                            "kcidb.validate_main")
-        self.assertExecutes('{"version":{"major":3,"minor":0}}',
-                            "kcidb.validate_main", "3")
+        self.assertExecutes(empty, "kcidb.merge_main", "--indent=0",
+                            stdout_re=re.escape(empty))
+        self.assertExecutes(second_empty,
+                            "kcidb.validate_main", "--indent=0",
+                            stdout_re=re.escape(second_empty))
         self.assertExecutes('{"version":{"major":2,"minor":0}}',
                             "kcidb.validate_main", "1",
                             status=1,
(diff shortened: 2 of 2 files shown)


spbnick:
Thank you, Maryam. Please "ping" me in a comment or on Slack, when you're ready for review.

mharyam:
> Thank you, Maryam. Please "ping" me in a comment or on Slack, when you're ready for review.

Thank you 
I will

spbnick (changes requested):
Yep, that's basically what needs to be done. However, could you please re-implement this using `misc.json_dump_stream()` and `misc.OutputArgumentParser`, similarly to how `upgrade_main()` does it? This way the tool would act more like others.

kcidb/__init__.py
```diff
@@ -253,8 +253,13 @@ def validate_main():
     parser = misc.ArgumentParser(description=description)
     misc.argparse_schema_add_args(parser, "validate against")
     args = parser.parse_args()
-    for data in misc.json_load_stream_fd(sys.stdin.fileno()):
-        print(args.schema_version.validate(data))
+    misc.json_dump_stream(
+        (
+            args.schema_version.upgrade(io.schema.validate(data), copy=False)
```
spbnick, line 258: This changes the function and the `kcidb-validate` tool to also upgrade the data, which is not in the issue you're trying to fix: https://github.com/kernelci/kcidb/issues/145

Your first attempt was basically fine, and I'm just asking you to make the tool's interface match the other tools.

Ping me on Slack, if you need help figuring this out.

kcidb/unittest.py
```diff
@@ -85,7 +85,7 @@ def assertExecutes(self, stdin, name, *args,
         if result.returncode != status:
             errors.append(f"Expected exit status {status}, "
                           f"got {result.returncode}")
-        if not re.fullmatch(stdout_re, result.stdout, re.DOTALL):
+        if not re.match(stdout_re, result.stdout, re.DOTALL):
```
spbnick, line 88: This is side-stepping the problem. The use of `fullmatch()` is intentional here. Something that should be documented in the function's docstring, though. Please revert this change, and if you can figure out how, document that the regular expressions are matched fully.

test_kcidb.py
```diff
@@ -171,7 +171,8 @@ def test_validate_main(self):
         """Check kcidb-validate works"""
         self.assertExecutes('', "kcidb.validate_main")
         self.assertExecutes('{"version":{"major":4,"minor":0}}',
-                            "kcidb.validate_main")
+                            "kcidb.validate_main", "--indent=0",
+                            stdout_re='{"version": {"major": 4, "minor": 0}}')
```
spbnick, line 175: Ah, you're getting there!

spbnick (changes requested):
Ah, this is better :smiley: The code itself is correct, but tests need more work.

The `assertExecutes` function is quite involved. Basically, it imitates execution of a separate program given its "main" function. We need to imitate it, because we need to be able to run tests without installing KCIDB, and the executables are only created by setuptools upon installation. I.e. there are no executables in the source code. So, `assertExecutes` creates a new process calling its `main` function, feeds it the specified standard input, captures the produced stdout and stderr, and the exit status, and then checks that they match the specified requirements.

So, e.g. this line:
```
self.assertExecutes('', "kcidb.validate_main")
```
Creates a new process, which calls the `kcidb.validate_main` function in the same manner as `kcidb-validate` would, feeds it the (default) empty string through standard input (stdin), reads its standard output (stdout) and standard error (stderr) streams, as well as its exit status, and checks that they match the (default) requirements of "", "", and 0 respectively.

Hope this helps you understand what the test is doing :)

Please also try running `kcidb-validate` with various inputs both before and after your change to get a feel of what it does, and what your change does.

spbnick:
Heey, this is very good, looks great!

However, could you please amend the commit subject to match what's being done? Something like "Output validated reports in kcidb-validate".

Also, do you think you could add another commit on top adding two more cases to the test:
* validate three valid reports in a row (in a single execution of `kcidb-validate`);
* validate three reports in a row, where the middle one is invalid, but the two others are valid?

It would be best if all three were different, in both cases, so we could make sure the ordering is correct and the right one is considered invalid.

spbnick:
This looks very nice! Thank you, Maryam! All code and tests are fine, but could you please amend commit subjects to better describe the changes?

The first one could be something like "Output validated reports in kcidb-validate", and the second something like "tests: Add two more kcidb-validate cases".

mharyam:
> This looks very nice! Thank you, Maryam! All code and tests are fine, but could you please amend commit subjects to better describe the changes?
> 
> The first one could be something like "Output validated reports in kcidb-validate", and the second something like "tests: Add two more kcidb-validate cases".

Done

spbnick:
Looks good to me. Thanks a lot! Merging.

PR #247, opened 2021-11-28
Implement kcidb.oo test

### LINKED ISSUE #207: Implement kcidb.oo module tests
After reimplementing the OO module we don't have tests for it anymore. Implement unit tests for the kcidb.oo module.

mrbazzan:
Fixes #207

### DIFF
--- a/kcidb/test_oo.py
+++ b/kcidb/test_oo.py
@@ -1,0 +1,203 @@
+"""kcidb.oo test module"""
+
+import kcidb
+
+
+class KCIDBOOClientTestCase(kcidb.unittest.TestCase):
+    """Test case for oo.Client"""
+    # Conform to unittest conventions, pylint: disable=invalid-name
+
+    def assertResponseContains(self, pattern_str, types, num_instances):
+        """
+        Check that the query for the specified pattern string returns
+        objects of types with specified names, in specified quantity,
+        and nothing else.
+        Args:
+            pattern_str:    A string which depicts the object to be retrieved.
+            types:          A set of specified schema type name(s).
+            num_instances:  The number of instance of object returned for
+                            each schema type.
+        """
+        assert isinstance(pattern_str, str)
+        assert isinstance(types, set)
+        assert types <= set(kcidb.oo.CLASSES)
+        assert isinstance(num_instances, int)
+
+        data = self.client.query(kcidb.orm.Pattern.parse(pattern_str))
+
+        self.assertEqual(set(data), types)
+        for type_name, type_objs in data.items():
+            self.assertEqual(len(type_objs), num_instances,
+                             msg=f"Invalid number of {type_name} instances")
+            for i, obj in enumerate(type_objs):
+                self.assertIsInstance(obj, kcidb.oo.CLASSES[type_name],
+                                      msg=f"Invalid type of {type_name} #{i}")
+
+    def setUp(self):
+        """Database setup and OO client creation"""
+        source = kcidb.db.Client('sqlite::memory:')
+        source.init()
+        source.load(
+            {
+                "version": {
+                    "major": 4,
+                    "minor": 0
+                },
+                "checkouts": [
+                    {
+                        "id": "_:kernelci:5acb9c2a7bc836e"
+                              "9e5172bbcd2311499c5b4e5f1",
+                        "origin": "kernelci",
+                        "git_commit_hash": "5acb9c2a7bc836e9e5172bb"
+                                           "cd2311499c5b4e5f1",
+                        "git_commit_name": "v5.15-4077-g5acb9c2a7bc8",
+                        "patchset_hash": ""
+                    },
+                    {
+                        "id": "_:redhat:5acb9c2a7bc836e9619c65f9"
+                              "709de72e90f2efd0",
+                        "git_commit_hash": "5acb9c2a7bc836e9619c65f9"
+                              "709de72e90f2efd0",
+                        "origin": "redhat", "patchset_hash": ""
+                    },
+                    {
+                        "id": "_:google:bd355732283c23a365f7c55206c03"
+                              "85100d1c389", "origin": "google",
+                        "git_commit_hash": "bd355732283c23a365f7c55206c038"
+                                           "5100d1c389",
+                        "patchset_hash": ""
+                    },
+                ],
+                "builds": [
+                    {
+                        "id": "google:google.org:a1d993c3n4c448b2j0l1hbf1",
+                        "origin": "google",
+                        "checkout_id": "_:google:bd355732283c23a365f7c"
+                                       "55206c0385100d1c389"
+                    },
+                    {
+                        "id": "kernelci:kernelci.org:619c6a30636e8fe042f2efa0",
+                        "origin": "kernelci",
+                        "checkout_id": "_:kernelci:5acb9c2a7bc836e9e5172"
+                                       "bbcd2311499c5b4e5f1"
+                    },
+                    {
+                        "id": "kernelci:kernelci.org:619c6c8d764403e3d4f2efff",
+                        "origin": "kernelci",
+                        "checkout_id": "_:kernelci:5acb9c2a7bc836e9e5172"
+                                       "bbcd2311499c5b4e5f1"
+                    },
+                    {
+                        "id": "kernelci:kernelci.org:619c7e1a2e53f8feb4f2efae",
+                        "origin": "kernelci",
+                        "checkout_id": "_:kernelci:5acb9c2a7bc836e9e5172"
+                                       "bbcd2311499c5b4e5f1"
+                    },
+                    {
+                        "id": "redhat:redhat.org:619c65f9709de72e90f2efd0",
+                        "origin": "redhat",
+                        "checkout_id": "_:redhat:5acb9c2a7bc836e9619c6"
+                                       "5f9709de72e90f2efd0"
+                    }
+                ],
+                "tests": [
+                    {
+                        "id": "google:google.org:a19di3j5h67f8d9475f26v11",
+                        "build_id": "google:google.org:a1d993c3n4c448b2"
+                                    "j0l1hbf1",
+                        "origin": "google",
+                    },
+                    {
+                        "waived": False,
+                        "start_time": "2021-11-23T03:52:13.660000+00:00",
+                        "path": "baseline.login",
+                        "status": "PASS",
+                        "id": "kernelci:kernelci.org:619c656de1fb4af479f2efaa",
+                        "build_id": "kernelci:kernelci.org:619c64b171"
+                                    "2847eccbf2efac",
+                        "origin": "kernelci"
+                    },
+                    {
+                        "build_id": "redhat:redhat.org:619c65f970"
+                                    "9de72e90f2efd0",
+                        "id": "redhat:redhat.org:b9d8be63bc2abca63165"
+                              "de5fd74f0f6d2f0b0d1c",
+                        "origin": "redhat",
+                        "waived": True,
+                        "status": "DONE"
+                    },
+                ]
+            }
+        )
+        self.client = kcidb.oo.Client(source)
+
+    def test_client_variable(self):
+        """Check instance for client variable"""
+        self.assertIsInstance(self.client, kcidb.oo.Client)
+
+    def test_zero_object(self):
+        """Check that empty object is returned."""
+        self.assertResponseContains("", set(), 0)
+
+    def test_one_object(self):
+        """Check that one object is returned for every type"""
+
+        self.assertResponseContains(
+            ">test[kernelci:kernelci.org:619c656de1fb4af479f2efaa]#",
+            {"test"}, 1
+        )
+
+        self.assertResponseContains(
+            ">build[kernelci:kernelci.org:619c6a30636e8fe042f2efa0]#",
+            {"build"}, 1
+        )
+
+        self.assertResponseContains(
+            ">checkout[_:redhat:5acb9c2a7bc836e9619c65f9709de72e90f2efd0]#",
+            {"checkout"}, 1
+        )
+
+        self.assertResponseContains(
+            '>revision["5acb9c2a7bc836e9e5172bbcd2311499c5b4e5f1", ""]#',
+            {"revision"}, 1
+        )
+
+        self.assertResponseContains(
+            '>revision["bd355732283c23a365f7c55206c0385100d1c389", ""]#'
+            '>checkout#>build#>test#',
+            {"revision", "checkout", "build", "test"}, 1
+        )
+
+    def test_two_object(self):
+        """Check that two objects are returned for every type"""
+
+        self.assertResponseContains(
+            ">test[kernelci:kernelci.org:619c656de1fb4af479f2efaa;"
+            "redhat:redhat.org:b9d8be63bc2abca63165de5fd74f0f6d2f0b0d1c]#",
+            {"test"}, 2
+        )
+
+        self.assertResponseContains(
+            ">build[redhat:redhat.org:619c65f9709de72e90f2efd0;"
+            "kernelci:kernelci.org:619c6c8d764403e3d4f2efff]#",
+            {"build"}, 2
+        )
+
+        self.assertResponseContains(
+            ">checkout[_:kernelci:5acb9c2a7bc836e9e5172bbcd2311499c5b4e5f1;"
+            "_:redhat:5acb9c2a7bc836e9619c65f9709de72e90f2efd0]#",
+            {"checkout"}, 2
+        )
+
+        self.assertResponseContains(
+            '>revision["5acb9c2a7bc836e9e5172bbcd2311499c5b4e5f1", "";'
+            '"bd355732283c23a365f7c55206c0385100d1c389", ""]#',
+            {"revision"}, 2
+        )
+
+        self.assertResponseContains(
+            '>revision["bd355732283c23a365f7c55206c0385100d1c389", "";'
+            '"5acb9c2a7bc836e9619c65f9709de72e90f2efd0", ""]#'
+            '>checkout#>build#>test#',
+            {"test", "build", "checkout", "revision"}, 2
+        )
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea/
 __pycache__
 *.egg-info
 *.pyc


mrbazzan:
CC @spbnick :
- I tried to run the test by entering `python3 kcidb/test_oo.py` but I got an `AttributeError: partially initialized module 'unittest' has no attribute 'TestCase' ...`. I think this is because of the unittest.py module in the kcidb folder. How do you run the test files?

NB: This is just a quick overview of what I'm doing before I write other corner cases

kcidb/test_oo.py
```diff
@@ -0,0 +1,101 @@
+"""kcidb.oo test module"""
+
+import os
+import kcidb
+import kcidb.db as db
+import kcidb.oo as oo
+
+from kcidb.test_orm import parse
```
spbnick, line 8: The `parse()` in `kcidb.test_orm` is just a tiny convenience function. Let's not create unnecessary dependencies, especially between tests, which are often ephemeral. Just copy `parse()` over here, if you need it (it seems you don't right now).

Please note, though, that `parse()` in `kcidb.test_orm` is using its own schema, and not **the** schema defined in `kcidb.orm.SCHEMA` and used by `kcidb.oo`. So it won't work properly for you.

kcidb/test_oo.py
```diff
@@ -0,0 +1,101 @@
...
+                        "origin": "kernelci"
+                    }
+                ],
+            }
+        )
+
+    def test_simple_build(self):
+        self.assertEqual(
+            self.client.query(
+                set(kcidb.orm.Pattern.parse('>test%#', [{('kernelci:kernelci.org:619c656de1fb4af479f2efac',)}]))
```
spbnick, line 82: `kcidb.orm.Pattern.parse()` already returns a set, so no need to call `set()` here.

kcidb/test_oo.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def _pre_setup(self):
+        self.path = 'sqlite:test_oo.sqlite3'
+        self.source = db.Client(self.path)
+
+        self.initialize_and_load_data()
+        self.assertTrue(isinstance(self.source, kcidb.orm.Source))
+
+    def setUp(self):
+        self._pre_setup()
```
spbnick, line 21: You don't really need to create and remove files. Just use SQLite's in-memory database support. I.e. specify the `sqlite::memory:` "database" to `db.Client()`. This way you won't need teardown.

spbnick (commented):
Nice start, @mrbazzan!

I left a few inline comments, which may, or may not be useful :grin:

I usually invoke tests using the `pytest` command. E.g.:

    pytest kcidb/test_orm.py

We might need to eventually fix the error you're seeing, just to be neat, but right now I'm not sure how to do that, and it's not critical.

Thank you!

spbnick:
I would also recommend passing `--disable-warnings` to pytest until we figure out what to do with the mess the Google Cloud modules spew out.

kcidb/test_oo.py
mrbazzan: I wanted to use it, but I ended up not using it. 
I forgot to remove the import.

mrbazzan:
> Nice start, @mrbazzan!
> 
> I left a few inline comments, which may, or may not be useful grin
> 
> I usually invoke tests using the `pytest` command. E.g.:
> 
> ```
> pytest kcidb/test_orm.py
> ```
> 
> We might need to eventually fix the error you're seeing, just to be neat, but right now I'm not sure how to do that, and it's not critical.
> 
> Thank you!

So, it's possible to run tests designed with `unittest` module with `pytest`. :O 
Thanks.

For the error, I think renaming the file might do the trick.

spbnick:
> For the error, I think renaming the file might do the trick.

Yeah it will work, but I don't want to do that. I specifically want it to be called `kcidb.unittest`, meaning "kcidb extensions of unittest".

kcidb/test_oo.py
```diff
@@ -0,0 +1,960 @@
...
+        self.initialize_and_load_data()
+        self.assertTrue(isinstance(self.source, kcidb.orm.Source))
+
+        self.client = oo.Client(self.source)
+
+    def initialize_and_load_data(self):
+        """Function to load some known data into the database"""
+
+        if not self.source.is_initialized():
+            self.source.init()
```
spbnick, line 36: You can just call `self.source.init()` in `setUp()` unconditionally, right after creating the database client.

kcidb/test_oo.py
```diff
@@ -0,0 +1,960 @@
...
+
+        self.source.load(
+            {
+                "version": {
+                    "major": 3,
```
spbnick, line 41: Could you create data with v4 schema instead? You can take your existing JSON data and just pass it through `kcidb-upgrade` to upgrade it to v4.

kcidb/test_oo.py
```diff
@@ -0,0 +1,960 @@
...
+                        "id": "kernelci:kernelci.org:619c656de1fb4af479f2efae",
+                        "build_id":
+                            "kernelci:kernelci.org:619c64b1712847eccbf2efac",
+                        "origin": "kernelci"
+                    }
+                ]
+            }
+        )
+
+        self.source.load(
```
spbnick, line 104: Any particular reason why you are loading the data in two chunks and not just one?

spbnick (changes requested):
I see you're moving further into the woods, niiice :grin:

What you're doing so far, actually mostly exercises the `kcidb.orm` module, and not `kcidb.oo` module, since you're basically just checking that the OO data is retrieved from the database. A `kcidb.oo` module test would instead access object fields and not the (private) `_data` field. That would also be the reason to not add the `get_data()` method you're asking about.

Would you like to perhaps switch to making a `kcidb.orm` module test first (we need those too!) out of what you've done so far, and once that's done, make a `kcidb.oo` one? The test would need to just create a database client and use it like a `Source` by calling its `oo_query()` method and checking it gets the right OO data (what you're checking for in the `_data` field right now).

Could you also make sure that CI passes (except the deployment check) before requesting reviews in the future?

Thanks :smiley:

mrbazzan:
> 
> Would you like to perhaps switch to making a `kcidb.orm` module test first (we need those too!) out of what you've done so far, and once that's done, make a `kcidb.oo` one? The test would need to just create a database client and use it like a `Source` by calling its `oo_query()` method and checking it gets the right OO data (what you're checking for in the `_data` field right now).
> 

Okay. That seems like the way to go. I'll create more test for the `kcidb.orm` module and after that create for the `kcidb.oo` module

mrbazzan:
> 
> Would you like to perhaps switch to making a `kcidb.orm` module test first (we need those too!) out of what you've done so far, and once that's done, make a `kcidb.oo` one? The test would need to just create a database client and use it like a `Source` by calling its `oo_query()` method and checking it gets the right OO data (what you're checking for in the `_data` field right now).
> 

Okay. That seems like the way to go. I'll create more test for the `kcidb.orm` module and after that create for the `kcidb.oo` module

mrbazzan:
> 
> Could you also make sure that CI passes (except the deployment check) before requesting reviews in the future?
> 

Okay. :)

mrbazzan:
@spbnick 
I have made some changes to the test. Help review.

mrbazzan:
@spbnick 
I have made some changes to the test. Help review.

spbnick:
Thank you, Abdulwasiu!

This looks mostly alright, but it is mostly repeating the `kcidb.orm` test you're making: checking that particular patterns return particular objects (please correct me if I missed something). Most of what makes this happen resides in the `kcidb.orm` module and is checked by your test. Here we don't really need to check that patterns are handled appropriately. Just checking that an object of every type makes it successfully through the `oo.Client` would be enough for the start, for the first commit.

Afterwards it would be good to add checking how all the properties of the objects are handled, but it could be another PR.
Thanks!

mrbazzan:
> Just checking that an object of every type makes it successfully through the `oo.Client` would be enough for the start, for the first commit.
> 

I have tried to refactor it to just check that an object is returned. I still don't know if i'm testing the right things though. 
Please review :)

mrbazzan:
> Afterwards it would be good to add checking how all the properties of the objects are handled, but it could be another PR. Thanks!

I don't really understand.

spbnick:
> > Afterwards it would be good to add checking how all the properties of the objects are handled, but it could be another PR. Thanks!
> 
> I don't really understand.

Things like checking all the cases handled by `kcidb.oo.Object.__getitem__()`, checking that all implementers of `BuildContainer` and `TestContainer` implement the methods correctly. That `Node` returns correct values for its properties, and so on. We can get to it later :)

kcidb/test_oo.py
```diff
@@ -0,0 +1,454 @@
+"""kcidb.oo test module"""
+
+import kcidb
+from kcidb import db
+from kcidb import oo
+
+
+class KCIDB00ClientTestCase(kcidb.unittest.TestCase):
```
spbnick, line 8: Whis is this using zeroes instead of ohs :grinning:?

kcidb/test_oo.py
```diff
@@ -0,0 +1,454 @@
...
+    """Test case for kcidb.oo module"""
+
+    def setUp(self):
+        """Database setup and oo client creation"""
+        self.source = db.Client('sqlite::memory:')
+        self.source.init()
+
+        self.initialize_and_load_data()
```
spbnick, line 16: This doesn't initialize the source, so I suppose it should be renamed.

kcidb/test_oo.py
```diff
@@ -0,0 +1,454 @@
...
+        self.assertTrue(isinstance(self.source, kcidb.orm.Source))
```
spbnick, line 17: If we need that test, then it shouldn't be here, but in the "orm" test instead. Please remove.

kcidb/test_oo.py
```diff
@@ -0,0 +1,454 @@
...
+from kcidb import db
+from kcidb import oo
+
+
+class KCIDB00ClientTestCase(kcidb.unittest.TestCase):
+    """Test case for kcidb.oo module"""
+
+    def setUp(self):
+        """Database setup and oo client creation"""
+        self.source = db.Client('sqlite::memory:')
```
spbnick, line 13: We don't really use `self.source` anywhere, but in `initialize_and_load_data`, so let's not store it in the instance, and instead just pass it to the latter explicitly.

spbnick (changes requested):
Ah, this is better, much simpler already. Thanks, Abdulwasiu!

Now, we don't really need most of what is still left. Just check that you get an instance of the correct type, for each simplest corresponding pattern. Maybe throw in a check that two objects of the same type and 2-4 objects of different types can be returned. That should be enough for the start.

I.e. check that '>revision[...]' returns a kcidb.oo.Revision, '>checkout[...]' - kcidb.oo.Checkout, and so on, plus doubles, and mixes thereof. Oh, and also check that an empty result could be returned.

mrbazzan:
> Ah, this is better, much simpler already. Thanks, Abdulwasiu!
> 
> Now, we don't really need most of what is still left (let the orm test handle it). Just check that you get an instance of the correct type, for each simplest corresponding pattern. Maybe throw in a check that two objects of the same type and 2-4 objects of different types can be returned. That should be enough for the start.
> 
> I.e. check that '>revision[...]' returns a kcidb.oo.Revision, '>checkout[...]' - kcidb.oo.Checkout, and so on, plus doubles, and mixes thereof. Oh, and also check that an empty result could be returned.

@spbnick 
Done! 
Please review when you can.

spbnick (changes requested):
Sorry it took me a while to get around to another review. I was occupied with preparing a talk for FOSDEM.

Thank you, Abdulwasiu. However, there are still too many unnecessary tests here. The OO module really doesn't care about which patterns are used to fetch objects, and its behavior never changes because of it directly. So we don't need to test all the possible ways we could retrieve e.g. a test. That's the job for the ORM module test, where all that logic resides.

For the first round we only need to check that the OO module can handle any type of object being returned from the underlying ORM module, and any basic combination thereof.

From that point of view, splitting the tests into methods by type doesn't really make sense. A better organization could be by the composition of the data returned from the ORM module. E.g.

* Empty
* One object, for every type in, turn
* Two objects of the same type, for every type, in turn
* One object of every type, together
* Two objects of every type, together

And that's it for the start.
That would be 1 + 4 + 4 + 1 + 1 = 11 queries in total.

Hope that makes sense :smiley: Thanks!

mrbazzan:
> And that's it for the start. That would be 1 + 4 + 4 + 1 + 1 = 11 queries in total.
> 

Review implemented. 
@spbnick Please check when you can.

kcidb/test_oo.py
```diff
@@ -0,0 +1,192 @@
...
+
+        self.check_type_instance(self.query_str(
+            ">checkout[_:redhat:5acb9c2a7bc836e9619c65f9709de72e90f2efd0]#"),
+            "checkout", kcidb.oo.Checkout
+        )
+
+        self.check_type_instance(self.query_str(
+            '>revision["5acb9c2a7bc836e9e5172bbcd2311499c5b4e5f1", ""]#'),
+            'revision', kcidb.oo.Revision
+        )
```
spbnick, line 157: This is close to "One object, for every type, in turn", but doesn't check that only one instance of each object was returned.

kcidb/test_oo.py
```diff
@@ -0,0 +1,192 @@
...
+            ">checkout[_:kernelci:5acb9c2a7bc836e9e5172bbcd2311499c5b4e5f1;"
+            "_:redhat:5acb9c2a7bc836e9619c65f9709de72e90f2efd0]#"),
+            "checkout", kcidb.oo.Checkout
+        )
+
+        self.check_type_instance(self.query_str(
+            '>revision["5acb9c2a7bc836e9e5172bbcd2311499c5b4e5f1", "";'
+            '"bd355732283c23a365f7c55206c0385100d1c389", ""]#'),
+            'revision', kcidb.oo.Revision
+        )
```
spbnick, line 181: I see that these are trying to check for two objects of each type, but they don't actually check how many objects were returned.

kcidb/test_oo.py
```diff
@@ -0,0 +1,192 @@
...
+        self.check_data(self.query_str(
+            '>revision["bd355732283c23a365f7c55206c0385100d1c389", ""]#'
+            '>checkout#>build#>test#')
+        )
+
+        self.check_data(self.query_str(
+            '>revision["bd355732283c23a365f7c55206c0385100d1c389", "";'
+            '"5acb9c2a7bc836e9619c65f9709de72e90f2efd0", ""]#'
+            '>checkout#>build#>test#')
+        )
```
spbnick, line 192: I see that these are trying to check for multiple objects, but they don't actually test how many and which types are returned.

kcidb/test_oo.py
```diff
@@ -0,0 +1,192 @@
...
+        """Run `check_type_instance` on each type."""
+        for type_name in obj:
+            if type_name == 'revision':
+                self.check_type_instance(obj, type_name, kcidb.oo.Revision)
+            elif type_name == 'checkout':
+                self.check_type_instance(obj, type_name, kcidb.oo.Checkout)
+            elif type_name == 'build':
+                self.check_type_instance(obj, type_name, kcidb.oo.Build)
+            else:
+                self.check_type_instance(obj, type_name, kcidb.oo.Test)
```
spbnick, line 24: Hmm, why don't you use the kcidb.oo.CLASSES dictionary?

spbnick (changes requested):
Ah, you're getting there! Thank you, Abdulwasiu :)

There are a few issues with the tests I mentioned in the inline comments. Also, could you call the `check_*` functions `assertSomething` instead to match the `unittest` convention? And specify what assertion they're actually verifying, instead of just saying which objects they check? E.g. something like `assertObjectsPresent` or such.

Also, if you're putting multiple tests inside one function, could you please comment what each block is verifying?

kcidb/test_oo.py
mrbazzan: I've updated it. `kcidb.oo.CLASSES` is a no-brainer.

mrbazzan:
> Also, if you're putting multiple tests inside one function, could you please comment what each block is verifying?

I've split the tests into functions(by number of objects returned). It looks better this way.

mrbazzan:
@spbnick Please review when you can.

kcidb/test_oo.py
```diff
@@ -0,0 +1,224 @@
...
+
+    def assertIsRevisionInstance(self, item, no_instance):
+        """Check that the instance of a oo.Revision is returned"""
+        self._check_type_instance(item, "revision", no_instance)
+
+    def assertContainsAllSchemaType(self, obj, no_instance):
+        """Check that the returned object contains all the schema type."""
+        assert set(obj) == set(kcidb.oo.CLASSES)
+        for type_name in obj:
+            self._check_type_instance(obj, type_name, no_instance)
```
spbnick, line 47: These are pretty good, but they still don't check that the returned dataset doesn't contain instances of types it shouldn't contain. E.g. if we're requesting revisions, then we shouldn't get tests in the response dataset, and these functions don't test that.

I think you could replace all these with just one method checking for a set of specified types and specified number of instances, while also checking no other types are present. I also notice that you have a query for each invocation, so we can combine this into something like this:
```python
def assertResponseContains(self, pattern_str, types, num_instances):
    """
    Check that the query for the specified pattern string returns objects of types
    with specified names, in specified quantity, and nothing else.

    Args:
        ...
    """
    # For Abdulwasiu: These are not a part of the test, but are just validating the arguments
    assert isinstance(pattern_str, str)
    assert isinstance(types, set)
    assert types < set(kcidb.oo.CLASSES)
    assert isinstance(num_instances, int)

    data = self.client.query(kcidb.orm.Pattern.parse(pattern_string))

    # For Abdulwasiu: These are actual test checks
    self.assertEqual(set(data), types)
    for type_name, type_objs in data.items():
        self.assertEqual(len(type_objs), num_instances,
                         msg=f"Invalid number of {type_name} instances")
        for i, obj in enumerate(type_objs):
            self.assertIsInstance(obj, kcidb.oo.CLASSES[type_name],
                                  msg=f"Invalid type of {type_name} #{i}")
```

I renamed `no_instance` to `num_instances` for clarity, because the former could be read as "no instance", meaning "there is no instance".

kcidb/test_oo.py
```diff
@@ -0,0 +1,224 @@
...
+                ]
+            }
+        )
+        self.client = kcidb.oo.Client(source)
+
+    def test_client_variable(self):
+        """Check instance for client variable"""
+        self.assertIsInstance(self.client, kcidb.oo.Client)
+
+    def test_one_object(self):
```
spbnick, line 157: We still need to check the zero-object response.

kcidb/test_oo.py
```diff
@@ -0,0 +1,224 @@
...
+        )
+
+    def test_two_object(self):
+        """Check that two objects are returned for every type"""
+
+        self.assertIsTestInstance(self.query_str(
+            ">test[kernelci:kernelci.org:619c656de1fb4af479f2efaa;"
+            "redhat:redhat.org:b9d8be63bc2abca63165de5fd74f0f6d2f0b0d1c]#"),
+            self.two
+        )
```
spbnick, line 187: So, using the function above, this, for example, could be written as:
```python
self.assertResponseContains(
    ">test[kernelci:kernelci.org:619c656de1fb4af479f2efaa;"
    "redhat:redhat.org:b9d8be63bc2abca63165de5fd74f0f6d2f0b0d1c]#",
    {"test"}, 2
)
```

kcidb/test_oo.py
```diff
@@ -0,0 +1,224 @@
...
+        """Check that the correct instance is returned from the OO Client."""
+
+        # check that empty query string returns empty object.
+        self.assertEqual(self.query_str(""), {})
+
+        # Check that one object of every type is returned together.
+        self.assertContainsAllSchemaType(self.query_str(
+            '>revision["bd355732283c23a365f7c55206c0385100d1c389", ""]#'
+            '>checkout#>build#>test#'), self.one
+        )
```
spbnick, line 217: And, as another example, this could be rewritten as:
```python
self.assertResponseContains(
    '>revision["bd355732283c23a365f7c55206c0385100d1c389", ""]#'
    '>checkout#>build#>test#',
    {"revision", "checkout", "build", "test"}, 1
)
```

kcidb/test_oo.py
```diff
@@ -0,0 +1,224 @@
...
+        """Check that the returned object contains all the schema type."""
+        assert set(obj) == set(kcidb.oo.CLASSES)
+        for type_name in obj:
+            self._check_type_instance(obj, type_name, no_instance)
+
+    def query_str(self, pattern_string):
+        """Run query on parsed string"""
+        return self.client.query(kcidb.orm.Pattern.parse(
+            pattern_string
+        ))
```
spbnick, line 53: If we integrate querying into the function above, we won't need this one.

spbnick (changes requested):
This is looking pretty good, we're getting there :) Thank you, Abdulwasiu. However, we still need to check the zero-object response, and that there are no objects of other types in the responses. I left a few suggestions and requests inline.

kcidb/test_oo.py
mrbazzan: https://github.com/kernelci/kcidb/blob/7dbc3dd0a8d8e1fc9ca4c6ee37c84e7a492bb63b/kcidb/test_oo.py#L211

Is this not enough for a zero-object response?

spbnick: Ah, I missed it, didn't expect it to be there :) Maybe move it to a `test_zero_objects()`, or such, for clarity? Thanks :)

mrbazzan: Alright.

kcidb/test_oo.py
```diff
@@ -0,0 +1,224 @@
+"""kcidb.oo test module"""
+
+import kcidb
+
+
+class KCIDBOOClientTestCase(kcidb.unittest.TestCase):
+    """Test case for oo.Client"""
+    # Conform to unittest conventions, pylint: disable=invalid-name
+    one = 1
+    two = 2
```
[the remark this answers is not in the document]
mrbazzan: Of course.

kcidb/test_oo.py
mrbazzan: > I renamed `no_instance` to `num_instances` for clarity, because the former could be read as "no instance", meaning "there is no instance".

👍👍

spbnick (changes requested):
Thank you very much, Abdulwasiu. The code is good to go. However, as the last tweak, could you please amend the commit subjects to use imperative mood, as in "commanding things to change"?

So, instead of ".idea/ added to .gitignore", write something like "Add .idea/ to .gitignore", and instead of "tests: kcidb.oo.test", write "tests: Add kcidb.oo test". This way it is easier to think about the changes when reverting and merging commits, and it makes it easier to read logs, as the mood is uniform.

Thank you!

mrbazzan:
> However, as the last tweak, could you please amend the commit subjects to use imperative mood, as in "commanding things to change"?
>
@spbnick Done!

spbnick (approved):
Looks great! Thank you, Abdulwasiu :smiley: Merging!

PR #255, opened 2021-12-02
Revert "Upgrade google-cloud pubsub"

spbnick:
Revert the upgrade of google-cloud-pubsub dependency in an attempt to
curb memory usage/leaks.

This reverts commit 5e49f8e76ab0841290ece4dc8942877d50d315bc.


spbnick:
KCIDB is likely not the cause of memory leaks, closing.

PR #257, opened 2021-12-07
Fixup submitter guide

spbnick:
Fixup/update a few small things in the Submitter Guide.


spbnick:
The auxilary file check failure is unrelated and will likely fix itself later. Merging!

PR #258, opened 2021-12-10
Addition to kcidb.orm test

mrbazzan:
@spbnick 
This is the PR that contains the added test for `kcidb.orm` module.

Please review

### DIFF
--- a/kcidb/test_orm.py
+++ b/kcidb/test_orm.py
@@ -1,5 +1,7 @@
 """kcdib.orm module tests"""
 
+# Over 1000 lines, pylint: disable=too-many-lines
+
 from jinja2 import Template
 import kcidb_io
 import kcidb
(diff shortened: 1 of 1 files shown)


kcidb/test_orm.py
```diff
@@ -1,5 +1,7 @@
 """kcdib.orm module tests"""
 
+# Over 1000 lines, pylint: disable=too-many-lines
```
mrbazzan, line 3: The lines exceed 1000. I hope this is okay?

spbnick: Yeah, for tests we can kinda tolerate it, but thanks for caring about it :+1:

spbnick:
This looks pretty good, thanks a lot, Abdulwasiu!

However, there are a lot of lines just saying `"something": None`, which are making the test harder to read. Do you think you could write a function which generates a dictionary with all the fields set to None for a dictionary of a particular type, and then just use that as a template, adding fields we actually expect to be set?

Something like:
```python
    empty_build = that_function(SCHEMA.types["build"])

    builds = [
        {
            **empty_build,
            "id": "kernelci:kernelci.org:619c6a30636e8fe042f2efa0",
        },
    ]
```

Or maybe just a function taking the non-null fields and returning a dictionary with the missing fields set to None?

Then we won't have so many lines of code in the test, and we won't have to update them all after adding or removing fields from the schema.

mrbazzan:
> Then we won't have so many lines of code in the test, and we won't have to update them all after adding or removing fields from the schema.

Done!
@spbnick Please review when you can.

kcidb/test_orm.py
```diff
@@ -498,3 +500,637 @@ def test_repr(self):
...
+                          "210cf85c5ec8d896",
+                    "origin": "kernelci",
+                    "patchset_hash": ""
+                })
+            ]
+        )
+
+        self.assertEqual(
+            self.source.oo_query(
+                kcidb.orm.Pattern.parse(
```
spbnick, line 836: You seem to always parse the pattern and then immediately query it, so how about you define a method in KCIDBORMSourceTestCase doing just that. Call it something like `query_str`, meaning "run OO query with a pattern string". Then you get to save two lines of code, and quite a few characters per each query, plus one indent level, making the code considerably shorter and easier to read.

kcidb/test_orm.py
```diff
@@ -498,3 +500,637 @@ def test_repr(self):
...
+                ]
+            }
+        )
+
+    def setUp(self):
+        """Database setup and oo client creation"""
+        self.source = kcidb.db.Client('sqlite::memory:')
+        self.source.init()
+
+        self.initialize_and_load_data()
```
spbnick, line 744: This doesn't initialize the source, only loads data, and is probably due for a rename.

kcidb/test_orm.py
```diff
@@ -498,3 +500,637 @@ def test_repr(self):
...
+        self.assertTrue(isinstance(self.source, kcidb.orm.Source))
```
spbnick, line 745: We shouldn't really be running tests in setup. If you want to have a test for this, please create a separate test case.

spbnick (changes requested):
Thanks a lot for keeping at it, Abdulwasiu :smiley: I mostly have style comments this time (or is it "again"?) :grin:

spbnick:
Also, thanks a lot for keeping CI passing!

mrbazzan:
> Also, thanks a lot for keeping CI passing!

@spbnick. Of course. Thanks for taking the time to review. 
I have implemented the requested changes, please review when you can

kcidb/test_orm.py
```diff
@@ -498,3 +498,494 @@ def test_repr(self):
...
+
+    Returns:
+        The dictionary for checkout type
+         with the missing fields set to None.
+    """
+    raw_data = {}
+    for item in kcidb.orm.SCHEMA.types["checkout"] \
+            .json_schema['properties'].keys():
+        raw_data[item] = kwargs.get(item, None)
+    return raw_data
```
spbnick, line 517: Instead of copying this code between the three or four (don't you need `raw_revision` as well?), could you please extract the common part into a function called `raw_data` and use it in all of these? It would accept the type name as a positional argument and the `**kwargs` as the field dictionary.

kcidb/test_orm.py
```diff
@@ -498,3 +498,494 @@ def test_repr(self):
         self.assertEqual(repr(pattern(None, True, "checkout",
                               {("",)})),
                          ">checkout[\"\"]#")
+
+
+def raw_checkout(**kwargs):
+    """
+    Template for checkout type
```
spbnick, line 505: Please word the function description in imperative mood (here and everywhere else). I.e. like a command. E.g. this one could be "Generate raw data for a checkout from values for some fields". See also documentation for most other functions for reference.

spbnick (changes requested):
Aaahhh, this is much better, thank you :relieved: This is much shorter and now I can see the forest for the trees :grin: I left a couple minor style comments inline, and while I'm not really happy about wrapping/formatting of all this code, I'll wave it in just to get it over with.

However, there's one important part missing: a test for revision retrieval. You have tests for checkouts, builds, and tests, but no "revisions" here. Could you add that?

Thank you!

mrbazzan:
> ... and while I'm not really happy about wrapping/formatting of all this code, I'll wave it in just to get it over with.
> 

I actually made it that way. I was trying not to go past 1000 lines. It felt weird to me too. I think I will just revert it back to the more pleasing edition.

mrbazzan:
@spbnick
Please review.

spbnick (approved):
Looks good, and certainly good for the start. Thanks a lot, @mrbazzan! Merging :)

PR #259, opened 2021-12-13
cloud: Look for Bash in PATH

spbnick:


spbnick:
@mharyam, the `cloud` fix we just did on your machine is now merged!

PR #266, opened 2021-12-17
deploy: Deploy main commits to production #256

### LINKED ISSUE #265: Deploy "main" commits to production

```bash
    --submitter=kernelci-ci-cki \
```

mharyam:
close #265

1. Create a workflow for check deployment
2. Add command to deploy to staging and production

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,15 +1,21 @@
-name: check_deployment
+name: deploy
 
 "on":
   push:
     branches: [main]
   pull_request:
     branches: [main]
+  workflow_run:
+    workflows: [ "test" ]
+    types:
+      - completed
+
 
 jobs:
-  check_deployment:
+  deploy:
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v2
       - name: Check credentials
         env:
           STAGING_GCLOUD_PROJECT: ${{ secrets.STAGING_GCLOUD_PROJECT }}
@@ -46,7 +52,7 @@ jobs:
           project_id: ${{ secrets.STAGING_GCLOUD_PROJECT }}
           service_account_key: ${{ secrets.STAGING_GCLOUD_SA_KEY }}
           export_default_credentials: true
-      - name: Deploy and Withdraw to Staging
+      - name: Deploy to Staging
         run: |
           ./cloud deploy kernelci-staging "" 1 -v \
               --smtp-password-file=<(echo foobar)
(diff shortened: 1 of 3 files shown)


spbnick (changes requested):
We gotta come up with a better name for the workflow, and actually with a workflow naming scheme in general. How about renaming the "main" workflow to "test" and the new one to "deploy"?

This new workflow would not be a "deployment check", but an actual deployment. So we don't need withdrawing. Let's start with deploying to the empty namespace on staging for the start (we don't have credentials for production in our repo anyway, yet).

We would need to setup dependency on the "test" workflow, so that we test the merged code before deploying.

See these for reference:
https://docs.github.com/en/actions/deployment/about-deployments/deploying-with-github-actions
https://stackoverflow.com/questions/58457140/dependencies-between-workflows-on-github-actions

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,65 @@
+name: deploy
+
+"on":
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
```
spbnick, line 7: We don't need to trigger on pull requests, so remove this one. We're deploying only on "pushes" to "main".

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,65 @@
...
+  workflow_run:
+    workflows: [ "test" ]
```
spbnick, line 9: Now, the question here is whether this will make the workflow trigger whenever the test workflow is executed, including when it's executed on PRs? Hmm. If that's so, then we need to do something else.

spbnick: We need to only trigger deployment when a change was merged, *and* the "test" workflow executed successfully. Could you figure out how to make that happen?

spbnick: Perhaps we need to use this instead: https://docs.github.com/en/actions/learn-github-actions/reusing-workflows

In essence, we need the "test" workflow run on pull requests, and when a change is merged we need to first run the "test" workflow, and then the "deploy" workflow.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,65 @@
...
+      - name: Install
+        run: |
+          python -m pip install --upgrade pip setuptools
+          pip3 install --upgrade '.[dev]'
+      - name: Setup Google Cloud SDK
+        uses: google-github-actions/setup-gcloud@master
+        with:
+          project_id: ${{ secrets.STAGING_GCLOUD_PROJECT }}
+          service_account_key: ${{ secrets.STAGING_GCLOUD_SA_KEY }}
+          export_default_credentials: true
```
spbnick, line 54: I wish we could reuse all of the above steps as well, instead of copying them, but let's start with the simply copy.

PR #267, opened 2021-12-17
deploy: Deploy main commits to staging #256

### LINKED ISSUE #265: Deploy "main" commits to production

mharyam:
close #265

Create a workflow for check deployment
Add command to deploy to staging and production

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,66 @@
+name: deploy
+
+
+# Only trigger, when the test workflow succeeded
+"on":
+  push:
+    branches: [main]
+  workflow_run:
+    workflows: ["test"]
+    types:
+      - completed
+
+
+jobs:
+  deploy:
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,67 @@
...
+  workflow_run:
+    workflows: ["test"]
+    types:
+      - completed
+
+
+jobs:
+  deploy:
+    # this job will only run if the PR has been merged
+    if: github.event.pull_request.merged == true
```
spbnick, line 17: Hmm, is this necessary, did you check? The "main" branch is where PRs land in, not e.g. where they're created from. Plus we also want to run the CI and the deployment if commits where pushed to "main" directly, without a PR.

spbnick (commented):
Looks great! I only have one inline question. Once that is resolved, we can merge this and see what happens :grin:

PR #269, opened 2021-12-22
pick abandoned notification #61

### LINKED ISSUE #61: Pick up abandoned notifications
At the moment posted notifications are picked up with a "new Google Cloud Firestore document" trigger to a Google Cloud Function. That is only triggered once upon posting the notification. If for whatever reason the `kcidb_send_notification` function failed after picking up that notification, and before sending it, and the next retry was within the pick up timeout, the function is going to succeed and is never going to be retried.

Consider adding a "cron job" going over all unsent notifications and sending them to wipe that up.

An alternative could be perhaps making the `kcidb_send_notification` function fail if it detects a picked-up notification, which wasn't sent, so that it could be retried until either the pick up timeout expires or the notification is sent.

mharyam:
close #61


cloud
```diff
@@ -627,6 +627,16 @@ function cloud_functions_deploy() {
...
+    cloud_function_deploy "$project" "${prefix}pick_notification" \
+                          "$source" \
+                          --entry-point kcidb_pick_notification \
+                          --env-vars-file "$env_yaml" \
+                          --runtime python37 \
+                          --trigger-event "${trigger_event}" \
+                          --trigger-resource "${trigger_resource}" \
+                          --memory 256MB \
+                          --retry \
+                          --timeout 540
```
spbnick, line 639: This will make `kcidb_pick_notification` trigger off new notifications getting created, but we need to trigger it off messages from your cloud scheduler job arriving to the message queue topic you would create for it.

Start with copying the `cloud_function_deploy` invocation for `kcidb_load_queue` function.
Copy and modify creation of the `kcidb_trigger` topic to create your own.
Copy and modify creation of the `kcidb_trigger` Cloud Scheduler job to create your own.

cloud
```diff
@@ -710,6 +720,9 @@ function scheduler_deploy() {
     # Deploy the jobs
     scheduler_job_pubsub_deploy "$project" "${prefix}trigger" \
                                 "$trigger_topic" '* * * * *' '{}'
+
+#    scheduler_job_pick_notification "$project" "${prefix}pick_notifications" \
+#                                     "$trigger_topic" '* * * * *' '{}'
```
spbnick, line 725: You don't need to copy the whole `scheduler_job_pubsub_deploy` function. You can just call `scheduler_job_pick_notification` function with the name of the topic you would create for the Cloud Scheduler to put its messages into (instead of `$trigger_topic`).

main.py
```diff
@@ -256,3 +255,40 @@ def kcidb_send_notification(data, context):
...
+            if cc_addrs:
+                message.replace_header("CC", cc_addrs + ", " + EXTRA_CC)
+            else:
+                message["CC"] = EXTRA_CC
+        # Connect to the SMTP server
+        smtp = smtplib.SMTP(host=SMTP_HOST, port=SMTP_PORT)
+        smtp.ehlo()
+        smtp.starttls()
+        smtp.ehlo()
+        smtp.login(SMTP_USER, SMTP_PASSWORD)
```
spbnick, line 285: You'll have to take care of specifying the correct `SMTP_USER` and `SMTP_PASSWORD`, if you want to actually send mail and test that this works. That is provided you're also using GMail.

spbnick (changes requested):

main.py
```diff
@@ -214,7 +213,7 @@ def kcidb_spool_notifications(event, context):
     # Spool notifications from subscriptions
     for notification in kcidb.monitor.match(OO_CLIENT.query(pattern_set)):
         if not SELECTED_SUBSCRIPTIONS or \
-           notification.subscription in SELECTED_SUBSCRIPTIONS:
+                notification.subscription in SELECTED_SUBSCRIPTIONS:
```
spbnick, line 216: Is there a need to change this indent and to remove the empty line above?

main.py
```diff
@@ -231,6 +230,31 @@ def kcidb_send_notification(data, context):
...
+        message = SPOOL_CLIENT.pick(notification_id)
+        if not message:
+            continue
+        # send message via email
+        send_email_notification(message, notification_id)
+        # Acknowledge notification as sent
+        SPOOL_CLIENT.ack(notification_id)
+
+
+def send_email_notification(message, notification_id):
```
spbnick, line 254: I think it's not worth making this function know about the notification and notification ID just to log a message about it. It's really not necessary for what it has to do. If we remove the `notification_id` argument, and just log that message right before each place where we call this function, then this function would be come simpler, more reusable, and less entangled with the rest of the code. We can even remove the word `notification` from its name, and just call it `send_email`, or even better, `send_message`, matching the way we call email messages throughout the code. Do you think you can do that?

main.py
```diff
@@ -231,6 +230,31 @@ def kcidb_send_notification(data, context):
...
+    """
+    Send message via email
+    """
```
spbnick, line 257: Could you document this function fully?

cloud
```diff
@@ -1251,5 +1272,4 @@ function execute() {
                     --log-level="$log_level" \
                     --submitters="${submitters[*]@Q}"
 }
-
```
spbnick, line 1254: Are you removing this empty line on purpose?

cloud
```diff
@@ -1251,5 +1272,4 @@ function execute() {
...
-execute "$@"
+execute "$@"
```
spbnick, line 1275: Could you restore the newline after this line? If we keep adding and removing them, it will be just unnecessary noise in git history.

spbnick (changes requested):
This is looking pretty good, we're almost there!

Apart from (minor) inline comments, I think we need better names instead of `kcidb_trigger` and `kcidb_pick_notification`. They're similar in purpose, but don't have similar names and that doesn't help understanding. I like the word "trigger" in `kcidb_trigger`, but we need to make it more specific how about we rename these two to `kcidb_load_queue_trigger` and `kcidb_pick_notifications_trigger`? For both the Cloud Scheduler jobs and the topics, I mean?

Finally, could you tell how you're testing those?

main.py
mharyam: PEP 8 in pycharm was complaining of the indent 
But I have reverted it back

mharyam:
> This is looking pretty good, we're almost there!
> 
> Apart from (minor) inline comments, I think we need better names instead of `kcidb_trigger` and `kcidb_pick_notification`. They're similar in purpose, but don't have similar names and that doesn't help understanding. I like the word "trigger" in `kcidb_trigger`, but we need to make it more specific how about we rename these two to `kcidb_load_queue_trigger` and `kcidb_pick_notifications_trigger`? For both the Cloud Scheduler jobs and the topics, I mean?
> 
> Finally, could you tell how you're testing those?

I am testing it by running the cloud script to deploy and withdraw the functions I worked on

main.py
spbnick: Ah, good. Does CI catch this? If not, then we need to fix CI.

However, in any case, we need to change this in a separate commit, since it has nothing to do with the PR's change. At this point you would need to organize your commits not by just how you made them locally, but by how they will go into the repo. So far it seems there should be one commit for whitespace/formatting fixes with explanations what required them, and one commit for the main change.

spbnick:
> I am testing it by running the cloud script to deploy and withdraw the functions I worked on

This is a good start, but we need to test the actual functionality. E.g. that an abandoned notification is picked and sent as an email. Manually will do for the start and for this PR, but we'll need automated tests for this soon.

cloud
```diff
@@ -435,6 +435,7 @@ function pubsub_deploy() {
                           new_debug_subscription \
                           updated_topic \
                           updated_debug_subscription \
+                          notification_topic \
```
spbnick, line 438: I know the function docs, are not much use so far, but we gotta maintain consistency. Could you add the new option to the function docs?

cloud
```diff
@@ -907,7 +925,8 @@ function execute_command() {
     # Convert submitters from a string of words to an array
     eval "declare -a submitters=($submitters)"
 
-    declare -r trigger_topic="${prefix}trigger"
+    declare -r trigger_topic="${prefix}load_queue_trigger"
+    declare -r notification_topic="${prefix}pick_notifications_trigger"
```
spbnick, line 929: Could you rename the variables and the option arguments to the functions, as well, so they make better sense and it's easier to track them? I.e. change `trigger_topic` to `load_queue_trigger_topic`, and `notification_topic` to `pick_notifications_trigger_topic`?

What you named `notification_topic` is still a "trigger topic" in the sense that it's only used for triggering a function execution, and its messages don't carry any information. And now that we have two "trigger topics", we need to distinguish them somehow too.

That's a lot of repetition, of course, but that's because we're at the start of implementation of our deployment and have little reuse and decoupling, still.

main.py
```diff
@@ -231,6 +230,38 @@ def kcidb_send_notification(data, context):
...
+    LOGGER.info("SENDING %s", notification_id)
+    # send message via email
+    send_message(message)
+    # Acknowledge notification as sent
+    SPOOL_CLIENT.ack(notification_id)
+
+
+def kcidb_pick_notifications():
+    """
+    Pick abandoned notifications and resend them
```
spbnick, line 243: We're not "resending" them, because we haven't sent them yet. We're just sending them. Please fix. Also please put a full stop at the end of the sentence (we probably need to add pydocstyle to CI).

main.py
```diff
@@ -231,6 +230,38 @@ def kcidb_send_notification(data, context):
...
+        LOGGER.info("SENDING %s", notification_id)
+        # send message via email
+        send_message(message)
+        # Acknowledge notification as sent
+        SPOOL_CLIENT.ack(notification_id)
+
+
+def send_message(message):
+    """
+    Send message via email
```
spbnick, line 260: Please put a full stop at the end of the sentence here too.

spbnick (changes requested):
Ah, we're getting there, I only have style comments :grin:

We still need to test the actual functionality, though.

cloud
```diff
@@ -624,7 +630,15 @@ function cloud_functions_deploy() {
...
-                          --trigger-topic "${trigger_topic}" \
+                          --trigger-topic "${load_queue_trigger_topic}" \
+                          --memory 256MB \
+                          --timeout 540
+    cloud_function_deploy "$project" "${prefix}pick_notification" \
+                          "$source" \
+                          --entry-point kcidb_pick_notifications\
+                          --env-vars-file "$env_yaml" \
+                          --runtime python37 \
+                          --trigger-topic "${load_queue_trigger_topic}" \
```
spbnick, line 641: Shouldn't this function be triggered by `"${pick_notifications_trigger_topic}"` instead, which should be passed to the function as an option?

cloud
```diff
@@ -637,6 +651,7 @@ function cloud_functions_withdraw() {
     cloud_function_withdraw "$project" "${prefix}send_notification"
     cloud_function_withdraw "$project" "${prefix}spool_notifications"
     cloud_function_withdraw "$project" "${prefix}load_queue"
+    cloud_function_withdraw "$project" "${prefix}pick_notification"
```
spbnick, line 654: This should be plural too, shouldn't it?

cloud
```diff
@@ -702,23 +717,28 @@ function scheduler_job_withdraw() {
...
 # Args: project prefix
 function scheduler_withdraw() {
     declare -r project="$1"; shift
     declare -r prefix="$1"; shift
-    scheduler_job_withdraw "$project" "${prefix}trigger" \
-                           "$trigger_topic" '* * * * *' '{}'
+    scheduler_job_withdraw "$project" "${prefix}load_queue_trigger" \
+                           "$load_queue_trigger_topic" '* * * * *' '{}'
+    scheduler_job_withdraw "$project" "${prefix}pick_notifications_trigger" \
+                       "$pick_notifications_trigger_topic" '30 * * * *' '{}'
```
spbnick, line 741: Hey, I just noticed, we don't need the last two arguments passed to `scheduler_job_withdraw`. Also, this function was trying to pull the value of `$trigger_topic` out of nowhere - it wasn't passed one! This was only working due to the really harmful way variable scope works in shell: variables defined in the caller are seen in the called function, so code like that could work called from one place (which has the variable), but not the other (which hasn't), one of the things I dislike most about shell. Both my faults.

The proper way to deal with this would be to have a commit adding the explicit `$trigger_topic` argument first, then a commit removing the extra argument to `scheduler_job_withdraw`, and then the commit adding the new feature (with a new argument for the notification-picking topic, and without the extra arguments to `scheduler_job_withdraw`). However, before you could do that (without going crazy trying to juggle all your commits) you need to transition to always having the final commits in your branch, instead of your work commits. I'll try to explain on Slack.

cloud
```diff
@@ -965,21 +987,22 @@ function execute_command() {
...
         sections_run "$sections" scheduler_deploy \
-            "$project" "$prefix" "$trigger_topic"
+            "$project" "$prefix" "$load_queue_trigger_topic" "$pick_notifications_trigger_topic"
         sections_run "$sections" submitters_deploy \
             "$project" "$new_topic" "${submitters[@]}"
     elif [ "$command" == "withdraw" ]; then
         sections_run "$sections" submitters_withdraw \
             "$project" "$new_topic" "${submitters[@]}"
-        sections_run "$sections" scheduler_withdraw "$project" "$prefix"
+        sections_run "$sections" scheduler_withdraw "$project" "$prefix" "$pick_notifications_trigger_topic"
```
spbnick, line 996: Good job trying to pass `$pick_notifications_trigger_topic` to `scheduler_withdraw`, only it's not actually trying to use it, which we need to fix.

spbnick (changes requested):
This is moving closer to the finish line, but beside the other small issues, I now found two problems with my code we need to fix :joy:

Issue #252, opened 2021-12-01
Do not render already spooled messages


spbnick:
At the moment our message-sending function is triggered off new Firestore documents. This means that, when creating a message in the spool, it should already have its contents rendered and supplied. That's fine for messages triggered off subscriptions for the first time, but it also means that we have to render the messages every time they're triggered again. E.g. after a subscription is triggered by the first failed build or test, it will keep triggering off all the subsequent ones, and keep rendering those messages. It's not such a big deal with the current (very simple) ORM prefetching/caching scheme, but we're still wasting CPU and memory, and eventually might run into trouble with SQL query quota too.

An earlier attempt to fix this in b1d916642876d5d16869f6b41219f0158a237652 was just reverted in 6c7aa0aa726a4a943ff2a47da47424235f76034b. We need to think of something better. E.g. doing it in two steps as in b1d916642876d5d16869f6b41219f0158a237652 and adding another message queue for triggering sending, but perhaps something better could be found.

PR #270, opened 2021-12-23
Implement delayed notifications #264

### LINKED ISSUE #264: Implement delayed notifications
@broonie asked for notifications on his revisions to be sent an hour after the last build result has arrived (regardless of status). For that we need to implement delayed notifications.

We need to store the notification ID , as well as the message (or maybe better, the message templates), along with a timestamp somewhere (perhaps Firestore again, it has queries by time we already use), and setup a periodic check for pending notifications, and sending of due ones (perhaps combine with #61). The subscriptions should be able to specify the time the notification should go off, and if it's triggered again, it should overwrite that timestamp, pushing the notification further. This way we can trigger on builds, generating revision notifications (needs #230) an hour after triggering, and once they stop for an hour, the notification would go off.

mharyam:
close #264
Add implementation for delayed notification

### DIFF
--- a/kcidb/monitor/output.py
+++ b/kcidb/monitor/output.py
@@ -1,5 +1,6 @@
 """Kernel CI reporting - monitor - output"""
 
+import datetime
 import re
 import base64
 import textwrap
@@ -28,7 +29,7 @@ class NotificationMessage:
     Message for a notification about a report object state.
     """
     # It's OK pylint: disable=too-many-arguments
-    def __init__(self, to, subject, body, cc=None, bcc=None, id=""):
+    def __init__(self, to, subject, body, cc=None, bcc=None, id="", due=None):
         """
         Initialize a notification message.
 
@@ -63,6 +64,10 @@ class NotificationMessage:
                             The system will only send one notification with
                             the same ID for the same subscription for each
                             database object.
+            due:            An "aware" datetime.datetime object specifying the
+                            time notification should be sent out,
+                            or None to use
+                            datetime.datetime.now(datetime.timezone.utc).
         """
         assert isinstance(to, list)
         assert all(isinstance(address, str) for address in to)
@@ -78,6 +83,8 @@ class NotificationMessage:
         assert all(isinstance(address, str) for address in bcc)
         assert isinstance(id, str)
         assert len(id.encode()) <= 256
+        assert due is None or \
+               isinstance(due, datetime.datetime) and due.tzinfo
 
         self.to = to
         self.subject = subject
@@ -85,6 +92,7 @@ class NotificationMessage:
         self.cc = cc
         self.bcc = bcc
         self.id = id
+        self.due = due
 
 
 class Notification:
--- a/kcidb/monitor/spool/__init__.py
+++ b/kcidb/monitor/spool/__init__.py
@@ -12,7 +12,6 @@ import datetime
 import email
 import email.policy
 import dateutil.parser
-from google.cloud.exceptions import Conflict
 from google.cloud import firestore
 from kcidb.misc import ArgumentParser, log_and_print_excepthook
 from kcidb.monitor.misc import is_valid_firestore_id
@@ -99,8 +98,9 @@ class Client:
                             datetime.datetime.now(datetime.timezone.utc).
 
         Returns:
-            True, if the notification was posted onto the spool,
-            False, if not (it was already there).
+            True, if the new notification was posted onto the spool,
+            or the existing, "unpicked" notification was updated,
+            False, if not (it existed and was "picked" at the moment).
         """
         assert isinstance(notification, Notification)
         assert timestamp is None or \
@@ -108,18 +108,32 @@ class Client:
         if timestamp is None:
             timestamp = datetime.datetime.now(datetime.timezone.utc)
 
-        try:
-            self._get_doc(notification.id).create(dict(
-                created_at=timestamp,
-                # Set to a definitely timed-out time, free for picking
-                picked_until=datetime.datetime.min,
-                message=notification.render().as_string(
-                    policy=email.policy.SMTPUTF8
-                ),
-            ))
-        except Conflict:
-            return False
-        return True
+        @firestore.transactional
+        def create_or_update(transaction):
+            due = notification.message.due \
+                if notification.message.due else timestamp
+            doc = self._get_doc(notification.id)
+            snapshot = doc.get(field_paths=["picked_until"],
+                               transaction=transaction)
+            if snapshot.exists:
+                # if the notification has been picked
+                if snapshot.get("picked_until") > timestamp:
+                    return False
+                transaction.update(doc, dict(
+                    message=notification.render().
+                    as_string(policy=email.policy.SMTPUTF8), due=due))
+            else:
+                transaction.create(doc, dict(
+                    created_at=timestamp,
+                    # Set to a definitely timed-out time, free for picking
+                    picked_until=datetime.datetime.min,
+                    message=notification.render().
+                    as_string(policy=email.policy.SMTPUTF8),
+                    due=due
+                ))
+            return True
+
+        return create_or_update(self.db.transaction())
 
     def pick(self, id, timestamp=None, timeout=None):
         """
@@ -157,14 +171,16 @@ class Client:
         def pick_if_not_picked(transaction, doc, timestamp):
             """Pick notification, if not picked yet"""
             # Get the document snapshot
-            snapshot = doc.get(field_paths=["picked_until", "message"],
+            snapshot = doc.get(field_paths=["picked_until", "message", "due"],
                                transaction=transaction)
             picked_until = snapshot.get("picked_until")
             message_text = snapshot.get("message")
+            due = snapshot.get('due')
             # If the document doesn't exist, has no "picked_until" field,
             # no "message" field, or the picking has not timed out yet
+            # or the due date has not passed
             if not picked_until or not message_text or \
-               picked_until > timestamp:
+               picked_until > timestamp or due > timestamp:
                 return None
             # Parse the message
             message = self.parser.parsestr(message_text)
--- a/kcidb/test_monitor.py
+++ b/kcidb/test_monitor.py
@@ -303,5 +303,5 @@ class MatchTestCase(unittest.TestCase):
                     "Mark Brown <broonie@kernel.org>" in n.message.to
         ])
         self.assertEqual(len(subjects), 2)
-        self.assertRegex(subjects[0], r"^Builds failed for ")
-        self.assertRegex(subjects[1], r"^Tests failed for ")
+        self.assertRegex(subjects[0], r"^Testing done for ")
+        self.assertRegex(subjects[1], r"^Testing done for ")
--- a/cloud
+++ b/cloud
@@ -730,7 +730,7 @@ function scheduler_deploy() {
                                 "$load_queue_trigger_topic" '* * * * *' '{}'
     scheduler_job_pubsub_deploy "$project" "${prefix}pick_notifications_trigger" \
                                 "$pick_notifications_trigger_topic" \
-                                 '*/30 * * * *' '{}'
+                                 '*/10 * * * *' '{}'
 }
 
 # Withdraw from the scheduler
--- a/kcidb/monitor/subscriptions/mark_brown.py
+++ b/kcidb/monitor/subscriptions/mark_brown.py
@@ -1,25 +1,30 @@
 """Mark Brown's subscription"""
+from datetime import (timezone, datetime, timedelta)
 
 from kcidb.monitor.output import NotificationMessage as Message
 
 
 def match_revision(revision):
     """Match revisions of interest to Mark Brown"""
+    # Repos we're interested in
     repo_urls = {
         "https://git.kernel.org/pub/scm/linux/kernel/git/arm64/linux.git",
         "https://git.kernel.org/pub/scm/linux/kernel/git/soc/soc.git",
     }
-    subject_sfx = ' failed for {% include "revision_summary.txt.j2" %}'
-    msg_args = dict(
+    # If the revision is not from one of our repos,
+    # or there are no finished builds
+    if not repo_urls & set(revision.repo_branch_checkouts) or \
+            revision.builds_valid is None:
+        return ()
+
+    # We need to send the notification one hour after the last build came in.
+    # However, we currently have no way of distinguishing build updates
+    # from updates to build's tests, and we can't send a notification
+    # about a revision from a build update yet. So we're sending the
+    # notification one hour after any revision update.
+    return (Message(
+        subject='Testing done for {% include "revision_summary.txt.j2" %}',
         to=["Mark Brown <broonie@kernel.org>"],
         body='{% include "revision_description.txt.j2" %}',
-    )
-    if revision.checkouts_valid and revision.builds_valid is not None and \
-       repo_urls & set(revision.repo_branch_checkouts):
-        if not revision.builds_valid:
-            return (Message(subject='Builds' + subject_sfx, **msg_args),)
-        if revision.tests_root.waived is False and \
-           revision.tests_root.status not in \
-           (None, "PASS", "DONE", "SKIP"):
-            return (Message(subject='Tests' + subject_sfx, **msg_args),)
-    return ()
+        due=datetime.now(timezone.utc) + timedelta(hours=1)
+    ),)


kcidb/monitor/spool/__init__.py
```diff
@@ -157,15 +170,21 @@ def pick(self, id, timestamp=None, timeout=None):
...
             picked_until = snapshot.get("picked_until")
             message_text = snapshot.get("message")
+            time_to_send = snapshot.get('time_to_send')
             # If the document doesn't exist, has no "picked_until" field,
             # no "message" field, or the picking has not timed out yet
             if not picked_until or not message_text or \
                picked_until > timestamp:
                 return None
+            # Check that notification is ready to be sent out
+            if time_to_send > datetime.datetime.now(datetime.timezone.utc):
```
spbnick, line 185: We need to be comparing to the `timestamp` argument here instead of the current time (see the argument docs). This would let us imitate picking at any time in testing, and also allow using consistent "current time" between multiple invocations of the function and other parts of the system.

kcidb/monitor/spool/__init__.py
```diff
@@ -97,6 +97,12 @@ def post(self, notification, timestamp=None):
             timestamp:      An "aware" datetime.datetime object specifying the
                             notification creation time, or None to use
                             datetime.datetime.now(datetime.timezone.utc).
+            time_to_send:   An "aware" datetime.datetime object specifying the
```
spbnick, line 100: Instead of `time_to_send` can we call this `due`, here and everywhere else? That would be shorter.

kcidb/monitor/spool/__init__.py
```diff
@@ -97,6 +97,12 @@ def post(self, notification, timestamp=None):
...
+                            time notification should be sent out,
+                            or None to use
+                            datetime.datetime.now(datetime.timezone.utc)
+                            + datetime.timedelta(hours=1) .
```
spbnick, line 104: We shouldn't be putting the default due delay here, in this function. For most of notifications it won't be suitable. Instead, the default here should be "due immediately",  which should be the creation time, i.e. the `timestamp` argument. Then subscriptions should be able to request their own due delay.

kcidb/monitor/spool/__init__.py
```diff
@@ -116,6 +127,8 @@ def post(self, notification, timestamp=None):
                 message=notification.render().as_string(
                     policy=email.policy.SMTPUTF8
                 ),
+                notification_id=notification.id,
```
spbnick, line 130: Why are you adding the notification ID as a field here?

kcidb/monitor/spool/__init__.py
spbnick: Instead of adding a separate if statement, do you think you could put the condition into the previous one? It does the same thing: refusing to pick the notification. It would get a lot of conditions, but they will all be doing the same thing.

spbnick (changes requested):
This is pretty good, I think you're on the right track!

However, apart from the inline comments, this is missing a major part of the puzzle. For the use case of "sending notification one hour after the last build arrives" we need to recreate notifications that we already submitted, with the new "due time", and the new message. I.e. a build comes in, we create the notification one hour ahead, then 15 minutes later another build comes in, and we recreate (overwrite) the notification with the new due time, 15 minutes later, and the new message (which now includes more builds), and so on, until builds stop, and finally the notification created last is sent one hour later. We need to figure out how to change our code to still not send more than one copy of the same notification, but allow overwriting them, while avoiding race conditions.

kcidb/monitor/spool/__init__.py
mharyam: It was mentioned on the issue description

> We need to store the notification ID , as well as the message (or maybe better, the message templates), along with a timestamp somewhere (perhaps Firestore again

spbnick: Yeah, I was just throwing a rough idea at the wall, trying to think outside the box, not necessarily using Firestore. Since we're using Firestore so far, we don't need to add a property for notification ID, we're already storing it in the document ID.

kcidb/monitor/spool/__init__.py
```diff
@@ -157,14 +168,16 @@ def pick(self, id, timestamp=None, timeout=None):
...
             # Get the document snapshot
-            snapshot = doc.get(field_paths=["picked_until", "message"],
+            snapshot = doc.get(field_paths=["picked_until", "message",
+                                            "due"],
                                transaction=transaction)
             picked_until = snapshot.get("picked_until")
             message_text = snapshot.get("message")
+            due = snapshot.get('due')
             # If the document doesn't exist, has no "picked_until" field,
             # no "message" field, or the picking has not timed out yet
```
spbnick, line 178: Could you expand this comment to also describe the added condition?

kcidb/monitor/spool/__init__.py
```diff
@@ -97,6 +97,11 @@ def post(self, notification, timestamp=None):
             timestamp:      An "aware" datetime.datetime object specifying the
                             notification creation time, or None to use
                             datetime.datetime.now(datetime.timezone.utc).
+            due:            An "aware" datetime.datetime object specifying the
+                            time notification should be sent out,
+                            or None to use
+                            datetime.datetime.now(datetime.timezone.utc).
```
spbnick, line 103: Let's make the default be the value of the `timestamp` argument, thus making `due` match the creation time, making the notification due for sending immediately.

kcidb/monitor/spool/__init__.py
```diff
@@ -99,9 +99,7 @@ def post(self, notification, timestamp=None, due=None):
                             datetime.datetime.now(datetime.timezone.utc).
             due:            An "aware" datetime.datetime object specifying the
                             time notification should be sent out,
-                            or None to use
-                            datetime.datetime.now(datetime.timezone.utc).
-
+                            or None to use the timestamp.
```
spbnick, line 102: Could you change `the timestamp` to `"timestamp"` so it's clearer we're referring to the specific argument?

spbnick:
This looks much nicer now, thank you! Just one inline comment for now, but, as I said last time, we still need to implement "due" pushing:

> For the use case of "sending notification one hour after the last build arrives" we need to recreate notifications that we already submitted, with the new "due time", and the new message. I.e. a build comes in, we create the notification one hour ahead, then 15 minutes later another build comes in, and we recreate (overwrite) the notification with the new due time, 15 minutes later, and the new message (which now includes more builds), and so on, until builds stop, and finally the notification created last is sent one hour later. We need to figure out how to change our code to still not send more than one copy of the same notification, but allow overwriting them, while avoiding race conditions.

kcidb/monitor/spool/__init__.py
```diff
@@ -97,29 +98,49 @@ def post(self, notification, timestamp=None):
...
-        return True
+        @firestore.transactional
+        def create_or_update(transaction, notification, message):
+            snapshot = self._get_doc(notification.id). \
+                get(field_paths=["picked_until"], transaction=transaction)
+            if timestamp > snapshot.get("picked_until"):
+                transaction. \
+                    update(self._get_doc(notification.id), dict(
+                           created_at=timestamp,
+                           message=message))
```
spbnick, line 130: In general, could you please wrap after the opening parenthesis, and keep each argument on its own line(s), if wrapping? E.g. like this:
```suggestion
                transaction.update(
                    self._get_doc(notification.id),
                    dict(created_at=timestamp, message=message)
                )
```
But read below, as I think we don't need this call.

kcidb/monitor/spool/__init__.py
```diff
@@ -251,6 +275,26 @@ def unpicked(self, timestamp=None):
...
+        # Get the document snapshot
+        snapshot = self._get_doc(notification.id). \
+            get(field_paths=["timestamp"], transaction=transaction)
+        timestamp = snapshot.get("timestamp")
+        transaction. \
+            update(self._get_doc(notification.id), dict(
+                created_at=timestamp,
+                message=notification.render().as_string(
+                 policy=email.policy.SMTPUTF8)))
+
```
spbnick, line 297: We don't need this function anymore, could you please remove it?

kcidb/monitor/spool/__init__.py
spbnick: Also, we shouldn't be updating the `created_at` timestamp, since we're just updating the notification, not creating a new one.

kcidb/monitor/spool/__init__.py
```diff
@@ -97,29 +98,49 @@ def post(self, notification, timestamp=None):
...
+            try:
+                transaction.set(self._get_doc(notification.id), dict(
+                    created_at=timestamp,
+                    # Set to a definitely timed-out time, free for picking
+                    picked_until=datetime.datetime.min,
+                    message=message,
+                    due=due
+                ))
+            except Conflict:
+                return False
```
spbnick, line 141: I think we should be able to handle both creating and updating existing documents with a single call of `set()` here. See the third use case of [set()](https://googleapis.dev/python/firestore/latest/document.html#google.cloud.firestore_v1.document.DocumentReference.set).

Granted we check if the document exists or not, above, and are not proceeding if it exists and is picked, we can do something like this:
```python
transaction.set(
    doc, merge=["message", "due"],
    created_at=timestamp,
    # Set to a definitely timed-out time, free for picking
    picked_until=datetime.datetime.min,
    message=message,
    due=due
)
```
The above should create a document with all the fields, if it doesn't exist, and if it does, only update the `message` and `due` fields.

BTW, could you store the document reference in a variable called `doc` and reuse it, instead of calling `self._get_doc()` every time?

kcidb/monitor/spool/__init__.py
```diff
@@ -97,29 +98,49 @@ def post(self, notification, timestamp=None):
...
-                ),
-            ))
-        except Conflict:
-            return False
-        return True
+        @firestore.transactional
+        def create_or_update(transaction, notification, message):
+            snapshot = self._get_doc(notification.id). \
+                get(field_paths=["picked_until"], transaction=transaction)
+            if timestamp > snapshot.get("picked_until"):
```
spbnick, line 126: What happens here if the document doesn't exist?

spbnick (changes requested):
Could you please address all comments before requesting review next time? I don't want to write about those extra whitespace changes again :joy:
Thank you :smiley:

kcidb/monitor/spool/__init__.py
```diff
@@ -97,29 +96,48 @@ def post(self, notification, timestamp=None):
...
                             datetime.datetime.now(datetime.timezone.utc).
+            due:            An "aware" datetime.datetime object specifying the
+                            time notification should be sent out,
+                            or None to use "timestamp".
 
         Returns:
             True, if the notification was posted onto the spool,
+            or an existing notification with timestamp past the
+            picked_until time was updated,
             False, if not (it was already there).
```
spbnick, line 107: Could you change this to something higher-level and clearer, like:
```suggestion
            True, if the new notification was posted onto the spool,
            or the existing, "unpicked" notification was updated,
            False, if not (it existed and was "picked" at the moment).
```

kcidb/monitor/spool/__init__.py
```diff
@@ -97,29 +96,48 @@ def post(self, notification, timestamp=None):
...
-            self._get_doc(notification.id).create(dict(
+        @firestore.transactional
+        def create_or_update(transaction, message, doc):
+            snapshot = doc.get(field_paths=["picked_until"],
+                               transaction=transaction)
+            picked_until = snapshot.get("picked_until")
+            # If the document doesn't exist, or
+            # the timestamp is passed the picked_until field
+            if not picked_until or \
+                    timestamp > picked_until:
```
spbnick, line 129: So, how does this work if the document doesn't exist? Is `snapshot` returned in any case, and then it returns `None` for `"picked_until"`, or does `doc.get()` return `None` when the document doesn't exist? I just want to make sure we're handling both cases correctly :grin:

spbnick: BTW, could you not wrap this line? I think this will fit in a single line just fine.

kcidb/monitor/spool/__init__.py
```diff
@@ -157,14 +175,17 @@ def pick(self, id, timestamp=None, timeout=None):
         def pick_if_not_picked(transaction, doc, timestamp):
             """Pick notification, if not picked yet"""
             # Get the document snapshot
-            snapshot = doc.get(field_paths=["picked_until", "message"],
+            snapshot = doc.get(field_paths=["picked_until", "message",
+                                            "due"],
```
spbnick, line 179: Could you not wrap this line, as well? It fits into 78 characters, which is fine.

kcidb/monitor/spool/__init__.py
```diff
@@ -157,14 +175,17 @@ def pick(self, id, timestamp=None, timeout=None):
...
                                transaction=transaction)
             picked_until = snapshot.get("picked_until")
             message_text = snapshot.get("message")
+            due = snapshot.get('due')
             # If the document doesn't exist, has no "picked_until" field,
             # no "message" field, or the picking has not timed out yet
+            # or the due date has not passed
             if not picked_until or not message_text or \
-               picked_until > timestamp:
+                    picked_until > timestamp or due > timestamp:
```
spbnick, line 188: Is this whitespace change necessary?

spbnick (changes requested):
Looks very good, we're almost there! I left some style comments, and a couple questions. Thank you :grinning:

kcidb/monitor/spool/__init__.py
mharyam: Hmm, yes doc.get() will return None when the document doesn't doesn't.
I am not sure I am handling that properly  
There is a method called exists to check if a doc exist, I think that will be better to handle an empty doc
from here: https://firebase.google.com/docs/firestore/query-data/get-data

kcidb/monitor/spool/__init__.py
mharyam: No
I will fix it, sorry for the unnecessary change

kcidb/monitor/spool/__init__.py
spbnick: I think if it's documented that it will return None, it's enough to check for that. We need to get the `picked_until` field anyway, so better do it with just one call to the API (and to the "cloud").

kcidb/monitor/spool/__init__.py
spbnick: No problem, thanks :)

kcidb/monitor/spool/__init__.py
```diff
@@ -97,29 +96,46 @@ def post(self, notification, timestamp=None):
...
 
-        try:
-            self._get_doc(notification.id).create(dict(
+        @firestore.transactional
+        def create_or_update(transaction, message, doc):
+            snapshot = doc.get(field_paths=["picked_until"],
+                               transaction=transaction)
+            picked_until = snapshot.get("picked_until")
+            # If the document doesn't exist, or
+            # the timestamp is passed the picked_until field
```
spbnick, line 126: The second part of the comment is a bit redundant and says basically what the condition says. I think it will be more helpful if we say what it actually means:
```suggestion
            # the notification is "picked"
```

spbnick (changes requested):
Just one tiny style comment this time, nicely done, Maryam :) Thank you :)
Now we have to work on subscription side of this.

spbnick (changes requested):
I see that you've added the argument and the attribute to store the due time in the NotificationMessage, but where is the code actually using that?

mharyam:
> I see that you've added the argument and the attribute to store the due time in the NotificationMessage, but where is the code actually using that?

@spbnick  It is used when a notification is posted. 
The post method in kcidb.monitor.spool

spbnick:
> @spbnick It is used when a notification is posted.
> The post method in kcidb.monitor.spool
Hmm, I see nothing accessing the `due` field of NotificationMessage in your change.

spbnick (changes requested):
Left you an inline comment.

Also, here's the change to the subscription we're doing this for (or something like this):
```patch
diff --git a/kcidb/monitor/subscriptions/mark_brown.py b/kcidb/monitor/subscriptions/mark_brown.py
index 40a4190..b298a6a 100644
--- a/kcidb/monitor/subscriptions/mark_brown.py
+++ b/kcidb/monitor/subscriptions/mark_brown.py
@@ -1,25 +1,31 @@
 """Mark Brown's subscription"""
 
+from datetime import (timezone, datetime, timedelta)
 from kcidb.monitor.output import NotificationMessage as Message
 
 
 def match_revision(revision):
     """Match revisions of interest to Mark Brown"""
+    # Repos we're interested in
     repo_urls = {
         "https://git.kernel.org/pub/scm/linux/kernel/git/arm64/linux.git",
         "https://git.kernel.org/pub/scm/linux/kernel/git/soc/soc.git",
     }
-    subject_sfx = ' failed for {% include "revision_summary.txt.j2" %}'
-    msg_args = dict(
+
+    # If the revision is not from one of our repos,
+    # or there are no finished builds
+    if not repo_urls & set(revision.repo_branch_checkouts) or \
+       revision.builds_valid is None:
+        return ()
+
+    # We need to send the notification one hour after the last build came in.
+    # However, we currently have no way of distinguishing build updates
+    # from updates to build's tests, and we can't send a notification
+    # about a revision from a build update yet. So we're sending the
+    # notification one hour after any revision update.
+    return (Message(
+        subject = 'Testing done for {% include "revision_summary.txt.j2" %}',
         to=["Mark Brown <broonie@kernel.org>"],
         body='{% include "revision_description.txt.j2" %}',
-    )
-    if revision.checkouts_valid and revision.builds_valid is not None and \
-       repo_urls & set(revision.repo_branch_checkouts):
-        if not revision.builds_valid:
-            return (Message(subject='Builds' + subject_sfx, **msg_args),)
-        if revision.tests_root.waived is False and \
-           revision.tests_root.status not in \
-           (None, "PASS", "DONE", "SKIP"):
-            return (Message(subject='Tests' + subject_sfx, **msg_args),)
-    return ()
+        due=datetime.now(timezone.utc) + timedelta(hours=1)
+    ),)
```
Please put it into a separate commit with appropriate message, on top, please.

kcidb/monitor/spool/__init__.py
```diff
@@ -97,29 +96,46 @@ def post(self, notification, timestamp=None):
             timestamp:      An "aware" datetime.datetime object specifying the
                             notification creation time, or None to use
                             datetime.datetime.now(datetime.timezone.utc).
+            due:            An "aware" datetime.datetime object specifying the
+                            time notification should be sent out,
+                            or None to use "timestamp".
```
spbnick, line 101: We don't have this argument anymore. Please remove the documentation for it.

kcidb/monitor/spool/__init__.py
```diff
@@ -97,29 +96,46 @@ def post(self, notification, timestamp=None):
...
+            False, if not (it existed and was "picked" at the moment).
         """
         assert isinstance(notification, Notification)
         assert timestamp is None or \
             isinstance(timestamp, datetime.datetime) and timestamp.tzinfo
         if timestamp is None:
             timestamp = datetime.datetime.now(datetime.timezone.utc)
+        message = notification.render().as_string(policy=email.policy.SMTPUTF8)
+        due = notification.message.due
+        doc = self._get_doc(notification.id)
```
spbnick, line 115: Could you move this line into the `create_or_update` closure? Nothing outside it needs it, and this way we'll protect ourselves from accidentally adding some code accessing the doc outside that function, and thus outside the transaction, which is supposed to protect data consistency.

kcidb/monitor/spool/__init__.py
```diff
@@ -97,29 +96,46 @@ def post(self, notification, timestamp=None):
...
 
-        try:
-            self._get_doc(notification.id).create(dict(
+        @firestore.transactional
+        def create_or_update(transaction, message, doc, due):
+            snapshot = doc.get(field_paths=["picked_until"],
+                               transaction=transaction)
+            if snapshot.exists:
```
spbnick, line 121: Ah, this is a nice one!

kcidb/monitor/spool/__init__.py
```diff
@@ -97,29 +96,46 @@ def post(self, notification, timestamp=None):
...
+                # if the notification has been picked
+                if snapshot.get("picked_until") > timestamp:
+                    return False
+                transaction.update(doc, dict(message=message, due=due))
+                return True
```
spbnick, line 126: Instead of having `return True` here, could you instead use two branches of the preceding `if` statement, so that we have something like this:
```
if exists:
    update
else:
    create
```
This way we will only have one successful exit point in the function, and if we need to add something to the successful exit path we have only one place to update, and less of a chance of forgetting to add it to this one as well?

kcidb/monitor/spool/__init__.py
spbnick: Could you revert this whitespace change, if it's unnecessary?

spbnick (changes requested):
Looks good! I have only style comments at this point :)

Apart from the inline comments, could you rewrite the commit subjects a bit?

"Add ... feature" is redundant in a commit subject. We can shorten that to "Support ...". Then, this is **the** kcidb repository, so "kcidb: " is both redundant and not specific enough. So, overall, write something like "monitor: Support delayed notifications".

"Update" is also redundant in a commit message. A commit is an "update" by definition. Just "kcidb:" is both redundant and not specific again here. We could use "subscriptions:" instead. Rather long, perhaps, but that's what we call the module so far, so be it. In summary, write something like "subscriptions: Send Mark's notifications one hour after updates".

Finally, please write which tests you have run, and let's see which ones we still need.

Thanks a lot!

spbnick:
From your description in our shared doc you said that you did one test: submit data with two revisions, both matching Mark's subscription, verify that you get two notifications one hour later.

This would be a positive test for delayed notifications. We would also need a negative test. I.e. submit some data which matches someone else's notification, and check that it's sent without a delay.

Also, this is what I said on Slack earlier, about testing:

> To start with, there are the two most basic cases. First, data which doesn't match Mark's subscription, but matches some other subscriptions should generate notifications immediately.
>
> Second, data which matches Mark's subscription should generate a notification within one hour earliest, and one hour plus notification-picking period, latest.
>
> We can lower both of these for the test to speed it up.
>
> You also need to test that the notification is actually pushed forward if new matching data comes in.
>
> Try to figure out this third case yourself.
>
> There's the fourth case as well. Try to figure out what it should be.
>
> And we should also check that "picked" notifications, and sent notifications are not overwritten by the post() of the same notification, but notifications which were picked, timed out, and weren't sent are overwritten. This might be harder to test on a complete setup, so we can test that with just API use, in a python3 REPL.

mharyam:
> From your description in our shared doc you said that you did one test: submit data with two revisions, both matching Mark's subscription, verify that you get two notifications one hour later.
> 
> This would be a positive test for delayed notifications. We would also need a negative test. I.e. submit some data which matches someone else's notification, and check that it's sent without a delay.
> 
> Also, this is what I said on Slack earlier, about testing:
> 
> > To start with, there are the two most basic cases. First, data which doesn't match Mark's subscription, but matches some other subscriptions should generate notifications immediately.
> > Second, data which matches Mark's subscription should generate a notification within one hour earliest, and one hour plus notification-picking period, latest.
> > We can lower both of these for the test to speed it up.
> > You also need to test that the notification is actually pushed forward if new matching data comes in.
> > Try to figure out this third case yourself.
> > There's the fourth case as well. Try to figure out what it should be.
> > And we should also check that "picked" notifications, and sent notifications are not overwritten by the post() of the same notification, but notifications which were picked, timed out, and weren't sent are overwritten. This might be harder to test on a complete setup, so we can test that with just API use, in a python3 REPL.

@spbnick 
Should updated notification be pushed forward one hour after the current due date or one hour after the current date-time?

kcidb/test_monitor.py
```diff
@@ -303,5 +303,5 @@ def test_mark_brown(self):
                     "Mark Brown <broonie@kernel.org>" in n.message.to
         ])
         self.assertEqual(len(subjects), 2)
-        self.assertRegex(subjects[0], r"^Builds failed for ")
-        self.assertRegex(subjects[1], r"^Tests failed for ")
+        self.assertRegex(subjects[0], r"^Testing done for ")
+        self.assertRegex(subjects[1], r"^Testing done for ")
```
spbnick, line 307: Would be good to check that the due time is set correctly here. However, for that to work reliably we would need to support passing the "current time" through the monitor.match() down to subscriptions. If you can figure out how to do that quickly, that would be great. If not, we can add it later.

kcidb/monitor/spool/__init__.py
```diff
@@ -99,27 +99,43 @@ def post(self, notification, timestamp=None):
...
+                    return False
+                transaction.update(doc, dict(message=message, due=due))
+            else:
+                transaction.create(doc, dict(
+                    created_at=timestamp,
+                    # Set to a definitely timed-out time, free for picking
+                    picked_until=datetime.datetime.min,
+                    message=notification.render().as_string(
+                        policy=email.policy.SMTPUTF8
+                    ),
```
spbnick, line 133: You're rendering the notification a second time here, and that's costly. Please change to render it once only.

kcidb/monitor/output.py
```diff
@@ -1,5 +1,5 @@
 """Kernel CI reporting - monitor - output"""
-
```
spbnick, line 2: Could you please not remove this blank line? They're here in every module, as a convention, and they help readability a little.

spbnick (changes requested):
Looks pretty good, thank you, @mharyam!

Have you figured out your problem when due was set to None?

I have a few inline comments, but otherwise, please don't submit commits with spurious whitespace changes. If you think they're beneficial, let's discuss them, and perhaps do them in a separate commit. Otherwise they create unnecessary churn in git history, and are distracting during reviews. Thank you.

kcidb/test_monitor.py
mharyam: > 

@spbnick 
Yes I have figured out the problem with due 
I am checking if due is None and assigning timestamp to it 
This is the update: https://github.com/kernelci/kcidb/pull/270/commits/5100a415bf646e4d2ec88ca000ce96479df76b4c#diff-cb11dfc89a15f32781b56c749d3b12fc5292e76f816d6575ef91a4c1ca6d52d7R113

kcidb/monitor/spool/__init__.py
mharyam: No I am not 
I have removed rendering notification from the post function, it is rendered only in the create_or_update clousre 

You requested that I moved it there yesterday 
From this comment: 

> Could you move this line into the create_or_update closure? Nothing outside it needs it, and this way we'll protect ourselves from accidentally adding some code accessing the doc outside that function, and thus outside the transaction, which is supposed to protect data consistency.

kcidb/monitor/spool/__init__.py
```diff
@@ -99,27 +99,43 @@ def post(self, notification, timestamp=None):
...
-                    policy=email.policy.SMTPUTF8
-                ),
-            ))
-        except Conflict:
-            return False
-        return True
+        @firestore.transactional
+        def create_or_update(transaction):
+            message = notification.render(). \
+                as_string(policy=email.policy.SMTPUTF8)
```
spbnick, line 115: I'm talking about this call of `notification.render()`. You do it here first, and then again when creating the document.

kcidb/monitor/spool/__init__.py
spbnick: Ah, that was about assigning the `doc` variable. I left a comment at the line where you're rendering the notification first. Here you're rendering it again, and it's better to avoid that, as that's a costly operation.

kcidb/test_monitor.py
spbnick: Ah, nice! Thank you :)

spbnick: Here, however, I'm talking about adding two extra checks to the test to verify that the subscription is specifying the `due` argument to the constructor of the notification. It's not a big deal if we don't have it here now.

kcidb/monitor/spool/__init__.py
mharyam: Ahhh, I have seen it, Sorry 😢 about that.
I did not see it earlier

kcidb/monitor/spool/__init__.py
mharyam: > 



> 

Yes I have figured out the problem with due time 
This is the new update I have: https://github.com/kernelci/kcidb/pull/270/files#diff-cb11dfc89a15f32781b56c749d3b12fc5292e76f816d6575ef91a4c1ca6d52d7R116

spbnick:
OK, let me summarize the tests I think we need to do. Some of them you already did, one way or another, but now we can check and see if we need to do anything else. Below, `D` is the notification delay of `mark_brown` subscription, and `T` is the period with which the `kcidb_pick_notifications` Cloud Function is invoked. We can lower both of these for the tests, to do them faster.

- [ ] Non-delayed notifications get sent immediately.
      Scenario:
      1. Submit data matching `nikolai_kondrashov` subscription
      2. Check that an email is generated immediately.
- [ ] Delayed notifications get sent after the specified delay.
      Scenario:
      1. Submit data matching `mark_brown` subscription.
      2. Check that an email is generated between `D` and `D+T` delay since the data was submitted.
- [ ] Delayed notifications get pushed forward when an update arrives before the due time. The new due time is calculated from the time the update has arrived.
      Scenario:
      1. Submit data matching `mark_brown` subscription.
      2. Wait for `D/2` time.
      3. Submit more data matching `mark_brown` subscription.
      2. Check that an email is generated between `D` and `D+T` delay since the **second** submission was made.
- [ ] Delayed notifications are not sent again, if they were already sent, and a new update to the object arrives afterwards.
      Scenario:
      1. Submit data matching `mark_brown` subscription.
      2. Wait until the notification email is generated.
      3. Submit more data matching `mark_brown` subscription.
      4. Check that no email is generated within the `D+T` time. It's enough to just check that the notification document in Firestore has not received the new message and/or due timestamp, when doing manual tests.
- [ ] A currently-picked notification (one with `picked_until` past the current time) cannot be picked, for both delayed, and non-delayed notifications
      Scenario:
      1. Using the Python API, create and post a notification.
      2. Using the Python API, pick the posted notification (with its ID), specifying a suitable picking timeout explicitly.
      3. Before the picking timeout expires, try to pick the notification again.
      4. Check that the notification is not picked the second time (`None` is returned from `post()`)
- [ ] A notification which was picked, but not sent (for whatever reason, e.g. a crash), can be picked again after the picking timeout expires (when current time is past the `picked_until` field).
      Scenario:
      1. Using the Python API, create and post a notification.
      2. Using the Python API, pick the posted notification (with its ID), specifying a suitable picking timeout explicitly.
      3. After the picking timeout expires, try to pick the notification again.
      4. Check that the notification is picked the second time (A message object is returned from `post()`).

kcidb/monitor/spool/__init__.py
spbnick: No problem, that's what reviews are for :smile:

spbnick (commented):
This looks almost perfect, thank you very much :smile: 

I have one more request only: please either move the rendering out of `create_or_update`, or move it next to each of the `update()`/`create()` calls. The former would be how we did it before, and the latter would be trying to avoid rendering the notification again, if it was already posted, and could improve performance (fixing #252), but would be new.

Otherwise, with your code as it is right now, we're getting the worst of both worlds: if the transaction detects a collision with another process changing the Firestore document, the whole `create_or_update()` function is executed again, and the notification is rendered again. It is also rendered regardless of whether we actually create/update the notification, or not, further into the function.

If you move it out of `create_or_update()`, then we will also render the notification message regardless of whether we create/update the notification or not, but at least we render it only once, regardless of whether the transaction is retried (`create_or_update() is re-executed), or not.

If you duplicate the rendering and put it both before `update()` and before `create()`, then rendering is redone in case transaction is retried, but at least we don't do it if notification already exists and we don't need to update it (e.g. it was already sent).

I'm inclined to think, that we'll have far less transaction retrials, than notification updates after they were sent, so the latter approach (putting rendering next to both `create()` and `update()`), and thus fixing #252, would be preferable.

spbnick:
Ah, and one more thing before we merge this. Please add a commit reducing the period the `kcidb_pick_notifications()` is invoked to, say, 10 minutes. Otherwise with current 30 minutes, the notifications might get sent as much as 1.5 hours past last update, instead of 1 hour. With 10-minute period, the worst delay would be 1 hour and ten minutes.

spbnick (approved):

spbnick:
Looks great, @mharyam! Thanks a lot :) Merging!
Have a great weekend!

PR #271, opened 2021-12-28
deploy: Deploy main commits to staging

### LINKED ISSUE #265: Deploy "main" commits to production

mharyam:
close #265

Update the deploy workflow to trigger only when test workflow is done, the last fix for this failed.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,62 @@
+name: deploy
+
+"on":
+  push:
+    branches: [main]
+
+jobs:
+  call-worflow-test:
+    uses: kernelci/kernelci-api/.github/workflows/test.yml@main
+
+  deploy:
+    needs: call-worflow-test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Check credentials
+        env:
+          STAGING_GCLOUD_PROJECT: ${{ secrets.STAGING_GCLOUD_PROJECT }}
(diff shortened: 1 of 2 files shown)


spbnick:
Could you please rebase this onto latest "main".
Also, the only difference I see from main is this:
```diff
diff --git a/.github/workflows/deploy.yml b/.github/workflows/deploy.yml
index 6088611..c5f9b1d 100644
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -6,7 +6,8 @@ name: deploy
   push:
     branches: [main]
   workflow_run:
-    workflows: ["test"]
+    workflows:
+      - test
     types:
       - completed
```
which is just a YAML syntax change, which should have no effect on the workflow.

What are you trying to do here?

spbnick:
You can find a solution for dependency faster if you create a personal empty repo with dummy workflows and try to implement it there first. Just have dummy workflows outputting messages with "echo" and running "sleep 60" (which does nothing for one minute), and try to implement dependency logic there. This way you won't have to wait for me to merge your PR to see it in action.

mharyam:
> You can find a solution for dependency faster if you create a personal empty repo with dummy workflows and try to implement it there first. Just have dummy workflows outputting messages with "echo" and running "sleep 60" (which does nothing for one minute), and try to implement dependency logic there. This way you won't have to wait for me to merge your PR to see it in action.
> 
> Once you figure it out in your own personal repo, you can then implement the same here.

Thank you for the suggestion 😄 
I actually thought of that too, that is what I am working on currently

PR #272, opened 2021-12-28
pick abandoned notification #61

### LINKED ISSUE #61: Pick up abandoned notifications

mharyam:
close #61

1. Remove extra whitespace and args
2. Explicitly pass trigger_topic to scheduler_withdraw
3. Remove extra args to scheduler_job_withdraw
4. Pick notification

### DIFF
--- a/cloud
+++ b/cloud
@@ -717,8 +717,7 @@ function scheduler_deploy() {
 function scheduler_withdraw() {
     declare -r project="$1"; shift
     declare -r prefix="$1"; shift
-    scheduler_job_withdraw "$project" "${prefix}trigger" \
-                           "$trigger_topic" '* * * * *' '{}'
+    scheduler_job_withdraw "$project" "${prefix}trigger"
 }
 
 # Deploy permissions for a submitter
--- a/cloud
+++ b/cloud
@@ -421,23 +421,25 @@ function pubsub_subscription_withdraw() {
 
 # Deploy to Pub/Sub
 # Args: --project=NAME
-#       --trigger-topic=NAME
+#       --load-queue-trigger-topic=NAME
 #       --new-topic=NAME
 #       --new-load-subscription=NAME
 #       --new-debug-subscription=NAME
 #       --updated-topic=NAME
 #       --updated-debug-subscription=NAME
+#       --pick-notifications-trigger-topic=NAME
 function pubsub_deploy() {
     params="$(getopt_vars project \
-                          trigger_topic \
+                          load_queue_trigger_topic \
                           new_topic \
                           new_load_subscription \
                           new_debug_subscription \
                           updated_topic \
                           updated_debug_subscription \
+                          pick_notifications_trigger_topic \
                           -- "$@")"
     eval "$params"
-    pubsub_topic_deploy "$project" "${trigger_topic}"
+    pubsub_topic_deploy "$project" "${load_queue_trigger_topic}"
 
     pubsub_topic_deploy "$project" "${new_topic}"
     pubsub_subscription_deploy "$project" "${new_topic}" \
@@ -452,11 +454,13 @@ function pubsub_deploy() {
     pubsub_topic_deploy "$project" "${updated_topic}"
     pubsub_subscription_deploy "$project" "${updated_topic}" \
                                "${updated_debug_subscription}"
+    pubsub_topic_deploy "$project" "${pick_notifications_trigger_topic}"
 }
 
 # Withdraw from Pub/Sub
 # Args: --project=NAME
-#       --trigger-topic=NAME
+#       --load-queue-trigger-topic=NAME
+#       --pick-notifications-trigger-topic=NAME
 #       --new-topic=NAME
 #       --new-load-subscription=NAME
 #       --new-debug-subscription=NAME
@@ -464,7 +468,8 @@ function pubsub_deploy() {
 #       --updated-debug-subscription=NAME
 function pubsub_withdraw() {
     params="$(getopt_vars project \
-                          trigger_topic \
+                          load_queue_trigger_topic \
+                          pick_notifications_trigger_topic \
                           new_topic \
                           new_load_subscription \
                           new_debug_subscription \
@@ -477,7 +482,8 @@ function pubsub_withdraw() {
     pubsub_subscription_withdraw "$project" "$new_debug_subscription"
     pubsub_subscription_withdraw "$project" "$new_load_subscription"
     pubsub_topic_withdraw "$project" "$new_topic"
-    pubsub_topic_withdraw "$project" "$trigger_topic"
+    pubsub_topic_withdraw "$project" "$load_queue_trigger_topic"
+    pubsub_topic_withdraw "$project" "$pick_notifications_trigger_topic"
 }
 
 # Deploy a Cloud Function
@@ -557,7 +563,8 @@ function cloud_functions_env() {
 
 # Deploy Cloud Functions
 # Args: --project=NAME --prefix=PREFIX --source=PATH
-#       --trigger-topic=NAME
+#       --load-queue-trigger-topic=NAME
+#       --pick-notifications-trigger-topic=NAME
 #       --log-level=NAME
 #       --new-topic=NAME --new-load-subscription=NAME
 #       --updated-topic=NAME --spool-collection-path=PATH
@@ -566,7 +573,8 @@ function cloud_functions_env() {
 #       --database=SPEC
 function cloud_functions_deploy() {
     params="$(getopt_vars project prefix source \
-                          trigger_topic \
+                          load_queue_trigger_topic \
+                          pick_notifications_trigger_topic \
                           log_level \
                           new_topic new_load_subscription \
                           updated_topic spool_collection_path \
@@ -599,6 +607,14 @@ function cloud_functions_deploy() {
     trigger_event+="document.create"
     declare trigger_resource="projects/$project/databases/(default)/documents/"
     trigger_resource+="${spool_collection_path}/{notification_id}"
+    cloud_function_deploy "$project" "${prefix}pick_notifications" \
+                          "$source" \
+                          --entry-point kcidb_pick_notifications\
+                          --env-vars-file "$env_yaml" \
+                          --runtime python37 \
+                          --trigger-topic "${pick_notifications_trigger_topic}" \
+                          --memory 256MB \
+                          --timeout 540
     cloud_function_deploy "$project" "${prefix}send_notification" \
                           "$source" \
                           --entry-point kcidb_send_notification \
@@ -624,7 +640,7 @@ function cloud_functions_deploy() {
                           --entry-point kcidb_load_queue \
                           --env-vars-file "$env_yaml" \
                           --runtime python37 \
-                          --trigger-topic "${trigger_topic}" \
+                          --trigger-topic "${load_queue_trigger_topic}" \
                           --memory 256MB \
                           --timeout 540
 }
@@ -634,6 +650,7 @@ function cloud_functions_deploy() {
 function cloud_functions_withdraw() {
     declare -r project="$1"; shift
     declare -r prefix="$1"; shift
+    cloud_function_withdraw "$project" "${prefix}pick_notifications"
     cloud_function_withdraw "$project" "${prefix}send_notification"
     cloud_function_withdraw "$project" "${prefix}spool_notifications"
     cloud_function_withdraw "$project" "${prefix}load_queue"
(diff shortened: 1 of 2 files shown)


cloud
```diff
@@ -965,21 +985,23 @@ function execute_command() {
             --smtp-password-secret="$smtp_password_secret" \
             --database="$database"
         sections_run "$sections" scheduler_deploy \
-            "$project" "$prefix" "$trigger_topic"
+            "$project" "$prefix" "$load_queue_trigger_topic" "$pick_notifications_trigger_topic"
```
spbnick, line 988: Could you wrap this line to be 79 characters max?

cloud
```diff
@@ -1252,4 +1274,4 @@ function execute() {
                     --submitters="${submitters[*]@Q}"
 }
 
-execute "$@"
+execute "$@"
```
spbnick, line 1277: Could you please make sure this commit doesn't remove the newline from the end of the file?

Maybe setup your editor to not remove the newline from the last line automatically, if it's doing that? Otherwise we will continue seeing this.

main.py
```diff
@@ -9,7 +9,6 @@
 import kcidb_io
 import kcidb
 
-
```
spbnick, line 12: If this is needed for something like making pylint or some other linter happy, please move this change to the first commit, otherwise, please revert it.

main.py
```diff
@@ -231,6 +230,38 @@ def kcidb_send_notification(data, context):
     message = SPOOL_CLIENT.pick(notification_id)
     if not message:
         return
+    # log message before sending email
```
spbnick, line 233: We're not logging **the** message here, not the notification's message, that is. We're only logging the notification's ID.

In some sense we **do** log a "message", that is "SENDING foobar", but that hardly needs a comment. If we have to have a comment, let it at least be something like "Log the notification ID", even though that is quite obvious from the code, but if you feel that's needed, that's fine too.

So, please either remove this comment, or change it to something like "Log the notification ID", whichever you prefer. Please also do the same to the similar comment below.

cloud
```diff
@@ -702,23 +717,26 @@ function scheduler_job_withdraw() {
...
 
 # Withdraw from the scheduler
 # Args: project prefix
 function scheduler_withdraw() {
     declare -r project="$1"; shift
     declare -r prefix="$1"; shift
-    scheduler_job_withdraw "$project" "${prefix}trigger" \
-                           "$trigger_topic" '* * * * *' '{}'
+    declare -r load_queue_trigger_topic="$1"; shift
+    scheduler_job_withdraw "$project" "${prefix}trigger" "$load_queue_trigger_topic"
```
spbnick, line 739: Please wrap this line, so it's not longer than 79 characters (that's our convention in KCIDB).

cloud
```diff
@@ -965,21 +985,23 @@ function execute_command() {
...
         sections_run "$sections" cloud_functions_withdraw "$project" "$prefix"
         sections_run "$sections" firestore_withdraw \
             "$project" "$spool_collection_path"
         sections_run "$sections" secrets_withdraw \
             "$project" "$smtp_password_secret"
         sections_run "$sections" pubsub_withdraw \
             --project="$project" \
-            --trigger-topic="$trigger_topic" \
+            --load-queue-trigger-topic="$load_queue_trigger_topic" \
+            --pick-notifications-trigger-topic="$pick_notifications_trigger_topic" \
```
spbnick, line 1004: Could you wrap this line to 79 characters max too? You can just remove the equals sign, and have the option value in a separate argument, that will be handled fine. I.e.:
```suggestion
            --pick-notifications-trigger-topic \
                "$pick_notifications_trigger_topic" \
```

cloud
```diff
@@ -624,7 +630,15 @@ function cloud_functions_deploy() {
...
-                          --trigger-topic "${trigger_topic}" \
+                          --trigger-topic "${load_queue_trigger_topic}" \
+                          --memory 256MB \
+                          --timeout 540
+    cloud_function_deploy "$project" "${prefix}pick_notification" \
+                          "$source" \
+                          --entry-point kcidb_pick_notifications\
+                          --env-vars-file "$env_yaml" \
+                          --runtime python37 \
+                          --trigger-topic "${load_queue_trigger_topic}" \
```
spbnick, line 641: This still has to be triggered off your new trigger topic (`pick_notifications_trigger_topic`), not `load_queue_trigger_topic`, it's what we created it for, right?

spbnick (changes requested):
Heeey, nice job splitting and ordering the commits!

I left a few comments regarding the changes inline, but otherwise, I'd call that a successful git rebase exercise :grin:

One thing about the first commit, though. I didn't mean it to be about "removing unrelated whitespace changes", I'm asking you to simply remove all unrelated whitespace changes from all your commits, and if you have to make any whitespace changes (like removing that one empty line I'm commenting on), because pylint or something else requires it, then make a separate commit first, and put it first. If you don't have to do any such changes, then don't do that commit. Thanks :smiley:

cloud
```diff
@@ -624,7 +630,15 @@ function cloud_functions_deploy() {
...
-                          --trigger-topic "${trigger_topic}" \
+                          --trigger-topic "${load_queue_trigger_topic}" \
+                          --memory 256MB \
+                          --timeout 540
+    cloud_function_deploy "$project" "${prefix}pick_notification" \
+                          "$source" \
+                          --entry-point kcidb_pick_notifications\
+                          --env-vars-file "$env_yaml" \
+                          --runtime python37 \
+                          --trigger-topic "${pick_notifications_trigger_topic}" \
```
spbnick, line 641: You gotta pass `$pick_notifications_trigger_topic` to the function explicitly as well. Otherwise we have the same situation as we had with `scheduler_deploy`.

cloud
```diff
@@ -624,7 +630,15 @@ function cloud_functions_deploy() {
                           --entry-point kcidb_load_queue \
                           --env-vars-file "$env_yaml" \
                           --runtime python37 \
-                          --trigger-topic "${trigger_topic}" \
+                          --trigger-topic "${load_queue_trigger_topic}" \
+                          --memory 256MB \
+                          --timeout 540
+    cloud_function_deploy "$project" "${prefix}pick_notification" \
```
spbnick, line 636: Please use plural for the Cloud Function's name, to match the entry point's name and what it's doing. Same for withdrawing below.

cloud
```diff
@@ -702,23 +717,27 @@ function scheduler_job_withdraw() {
...
-    declare -r trigger_topic="$1"; shift
+    declare -r load_queue_trigger_topic="$1"; shift
+    declare -r pick_notifications_trigger_topic="$1"; shift
     # Deploy the jobs
-    scheduler_job_pubsub_deploy "$project" "${prefix}trigger" \
-                                "$trigger_topic" '* * * * *' '{}'
+    scheduler_job_pubsub_deploy "$project" "${prefix}load_queue_trigger" \
+                                "$load_queue_trigger_topic" '* * * * *' '{}'
+    scheduler_job_pubsub_deploy "$project" "${prefix}pick_notifications_trigger" \
+                            "$pick_notifications_trigger_topic" '30 * * * *' '{}'
```
spbnick, line 730: Could you please indent this line the same way the previous ones are indented, and wrap the arguments so that the lines are 79 characters max?

cloud
```diff
@@ -702,23 +717,27 @@ function scheduler_job_withdraw() {
...
 # Withdraw from the scheduler
 # Args: project prefix
 function scheduler_withdraw() {
     declare -r project="$1"; shift
     declare -r prefix="$1"; shift
-    declare -r trigger_topic="$1"; shift
-    scheduler_job_withdraw "$project" "${prefix}trigger" "$trigger_topic"
+    declare -r load_queue_trigger_topic="$1"; shift
+    scheduler_job_withdraw "$project" "${prefix}trigger"\
+                            "$load_queue_trigger_topic"
```
spbnick, line 740: Could you move this line one character to the left so it aligns with the first argument?

cloud
```diff
@@ -965,22 +986,25 @@ function execute_command() {
             --smtp-password-secret="$smtp_password_secret" \
             --database="$database"
         sections_run "$sections" scheduler_deploy \
-            "$project" "$prefix" "$trigger_topic"
+            "$project" "$prefix" "$load_queue_trigger_topic" \
+             "$pick_notifications_trigger_topic"
```
spbnick, line 990: Please move this line one character to the left, so it aligns with the previous one.

spbnick (changes requested):
Thanks a lot for cleaning and organizing your commits, and removing the unnecessary changes!

We're almost there! Only a few minor fixes left, and then, of course, we have to test the functionality itself.

main.py
```diff
@@ -231,6 +231,37 @@ def kcidb_send_notification(data, context):
...
+def kcidb_pick_notifications():
+    """
+    Pick abandoned notifications and send them.
+    """
+    for notification_id in SPOOL_CLIENT.unpicked():
+        # Pick abandoned notification and resend
+        message = SPOOL_CLIENT.pick(notification_id)
+        if not message:
+            continue
+        # log message before sending email
```
spbnick, line 250: Could you remove this comment as well, since you removed the one above?

cloud
```diff
@@ -965,22 +986,25 @@ function execute_command() {
...
         sections_run "$sections" cloud_functions_withdraw "$project" "$prefix"
         sections_run "$sections" firestore_withdraw \
             "$project" "$spool_collection_path"
         sections_run "$sections" secrets_withdraw \
             "$project" "$smtp_password_secret"
         sections_run "$sections" pubsub_withdraw \
             --project="$project" \
-            --trigger-topic="$trigger_topic" \
+            --load-queue-trigger-topic="$load_queue_trigger_topic" \
+            --pick-notifications-trigger-topic= \
```
spbnick, line 1006: Please remove the "equals" character (`=`) from the option's end. It's not generally supposed to be there, if its value is in a separate argument.

cloud
```diff
@@ -702,23 +719,28 @@ function scheduler_job_withdraw() {
...
 # Withdraw from the scheduler
 # Args: project prefix
 function scheduler_withdraw() {
     declare -r project="$1"; shift
     declare -r prefix="$1"; shift
-    declare -r trigger_topic="$1"; shift
-    scheduler_job_withdraw "$project" "${prefix}trigger" "$trigger_topic"
+    declare -r load_queue_trigger_topic="$1"; shift
+    scheduler_job_withdraw "$project" "${prefix}trigger"\
+                           "$load_queue_trigger_topic"
```
spbnick, line 743: Waaaait, I was wrong, we don't really need passing the third argument to `scheduler_job_withdraw` at all. It doesn't care which topic the Cloud Scheduler is sending its messages to, when removing the job, it only needs the project name and the job name to remove it. Thus we don't need the second commit :man_facepalming: Please remove it. Please also remove passing `"$triger_topic"` to `scheduler_job_withdraw` in the first commit.

Sorry, I confused you. Comes from dipping into things for half an hour only and trying to do it quick :grin:

This still has another problem, though: we need to update the jobs we're withdrawing here, to match what we're deploying in `scheduler_deploy`. That is, we need to withdraw both `"${prefix}load_queue_trigger"` and `"${prefix}pick_notifications_trigger"` here.

cloud
```diff
@@ -624,7 +632,15 @@ function cloud_functions_deploy() {
...
-                          --trigger-topic "${trigger_topic}" \
+                          --trigger-topic "${load_queue_trigger_topic}" \
+                          --memory 256MB \
+                          --timeout 540
+    cloud_function_deploy "$project" "${prefix}pick_notifications" \
+                          "$source" \
+                          --entry-point kcidb_pick_notifications\
+                          --env-vars-file "$env_yaml" \
+                          --runtime python37 \
+                          --trigger-topic "${pick_notifications_trigger_topic}" \
```
spbnick, line 643: Could you move this function deployment before the `send_notifications` one? This way we'll be replacing functions back-to-front (pipeline-wise), making it easier to preserve compatibility with updates.

cloud
```diff
@@ -637,6 +653,7 @@ function cloud_functions_withdraw() {
     cloud_function_withdraw "$project" "${prefix}send_notification"
     cloud_function_withdraw "$project" "${prefix}spool_notifications"
     cloud_function_withdraw "$project" "${prefix}load_queue"
+    cloud_function_withdraw "$project" "${prefix}pick_notifications"
```
spbnick, line 656: Same here, could you move the function to before `send_notifications`? Here the order is not critical, I think, but it's better to preserve code uniformity in general.

cloud
```diff
@@ -940,21 +963,24 @@ function execute_command() {
...
             --project="$project" \
-            --trigger-topic="$trigger_topic" \
+            --load-queue-trigger-topic="$load_queue_trigger_topic" \
             --new-topic="$new_topic" \
             --new-load-subscription="$new_load_subscription" \
             --new-debug-subscription="$new_debug_subscription" \
             --updated-topic="$updated_topic" \
-            --updated-debug-subscription="$updated_debug_subscription"
+            --updated-debug-subscription="$updated_debug_subscription" \
+            --pick-notifications-trigger-topic="$pick_notifications_trigger_topic"
```
spbnick, line 972: Could you wrap this line same as the similar line below?

spbnick (changes requested):
Looks great! I only left a few minor requests inline. After those are fixed, please move into testing this.

The (manual for now) test should consists of the following steps:
* Create a fresh deployment, capable of sending email to an address you control and mailbox you can access. Read `main.py` to figure out how to do that. You can modify it temporarily to do what you want and then deploy (we would need to implement configuring that later).
* Delete the `kcidb_send_notification` Cloud Function, so that notifications are not sent immediately. This would let us create "abandoned" notifications.
* Load certain data into the database so that at least one notification is generated. The sample I pointed you at earlier would probably be enough.
* Check that you get the corresponding email notification(s) delivered to your address.

spbnick:
Another test should be the same as the previous one, but without removing the `kcidb_send_notification` Cloud Function, to verify that notifications are still delivered normally.

spbnick (commented):
Looks good to me, now only testing remains! Thank you, Maryam :grin:

spbnick (approved):
LGTM! Thank you, Maryam!

PR #273, opened 2022-01-03
Actions: Do double deploy/withdraw

spbnick:
Check that repeated deployment and withdrawing don't break, in GitHub
actions.


mharyam (approved):
LGTM

PR #275, opened 2022-01-05
Actions: Make test workflow reusable and call it from deploy

### LINKED ISSUE #265: Deploy "main" commits to production

mharyam:
close #265 

1) Change test.yml to a reusable workflow 
2) Make deploy.yml call test.yml workflow 
3) Change triggers call for all the workflows

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,19 +1,15 @@
 name: deploy
 
-
-# Only trigger, when the test workflow succeeded
 "on":
   push:
     branches: [main]
-  workflow_run:
-    workflows: ["test"]
-    types:
-      - completed
-
 
 jobs:
+  test:
+    uses: kernelci/kcidb/.github/workflows/test.yml@main
+
   deploy:
-    # this job will only run if the PR has been merged
+    needs: test
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,10 +1,9 @@
 name: test
 
 "on":
-  push:
-    branches: [main]
   pull_request:
     branches: [main]
+  workflow_call: {}
 
 jobs:
   check_python:


.github/workflows/deploy.yml
```diff
@@ -1,19 +1,15 @@
...
   push:
     branches: [main]
-  workflow_run:
-    workflows: ["test"]
-    types:
-      - completed
-
 
 jobs:
+  call-worflow-test:
```
spbnick, line 8: Typo: `worflow` -> `workflow`. Although, I would simply call this job `test`.

.github/workflows/deploy.yml
```diff
@@ -1,19 +1,15 @@
...
+    uses: kernelci/kernelci-api/.github/workflows/test.yml@main
```
spbnick, line 9: This is pointing to another repository and to a non-existing workflow.

.github/workflows/test.yml
```diff
@@ -1,10 +1,10 @@
 name: test
 
 "on":
-  push:
-    branches: [main]
   pull_request:
     branches: [main]
+  workflow_call:
+    null
```
spbnick, line 7: Could you instead assign an empty dictionary to `workflow_call`. The use of `null` here is somewhat obscure, I think. Like this:
```yaml
  workflow_call: {}
```

spbnick (changes requested):
Apart from the inline comments, the commit subject describes only a part of what you do here. Please try to summarize your whole change instead. Thank you!

.github/workflows/deploy.yml
```diff
@@ -1,19 +1,15 @@
...
-      - completed
-
 
 jobs:
+  test:
+    uses: kernelci/kcidb/.github/workflows/test.yml@main
+
   deploy:
-    # this job will only run if the PR has been merged
+    needs: call-worflow-test
```
spbnick, line 12: This doesn't match the job name anymore.

spbnick (changes requested):
I have just one inline comment this time.

The commit message is better now, it is not telling less than what's happening. However, now it's too generic, and doesn't say enough. It doesn't even say which workflow is updated. Could you come up with something more meaningful? Something about what we're actually changing here? What's the essence of this change, and perhaps also what's the intention (although the latter can go into the body of the commit message instead).

spbnick:
Ah, there's still a shell syntax issue in deploy.yml, at line 55, which you introduced with your previous change. Could you fix it in a separate commit before the main one?

spbnick (approved):
Ah, this looks alright, thank you @mharyam :) Let's try merging it.

PR #280, opened 2022-01-07
Actions: Fix shell syntax issue in deploy workflow 

### LINKED ISSUE #265: Deploy "main" commits to production

mharyam:
close #265
fix shell syntax issue in deploy workflow


spbnick:
Thank you! Looks good to me. Will merge once CI finishes.

However, this wouldn't really close #265, so I'll reopen it. You can still link to the issue with something like "Concerns #265".

spbnick:
We still need to implement secret passing as I mentioned on Slack for deployment to work.

spbnick (approved):

PR #281, opened 2022-01-07
Actions: explicitly pass secrets to the called test workflow

mharyam:
Concerns #265"


spbnick:
Makes sense, thank you, @mharyam!

However, could you please rebase on latest origin/main? This should drop your already-merged commit from the commit list.

spbnick (approved):
Looks great! Thank you, @mharyam :)

PR #287, opened 2022-01-26
monitor: Truncate notification subject and message to char limit

### LINKED ISSUE #276: Sanitize out-of-spec notification messages instead of aborting
Right now we have (somewhat arbitrary) limits on what notification message subject and body should be: 256 and 4096 bytes of UTF-8 respectively. We also have a requirement of no control characters for subjects. The exact lengths don't matter, as long as we fit the maximum Firestore document size of 1MB.

However, we're handling any trespassing by assertion failure, which stops the notifications from being delivered. Instead we need to force compliance on the message by truncating and/or sanitizing it, log a warning, and let it through.

To make it easier we can redefine subject/body length limits to be in characters, assuming that maximum UTF-8-encoded byte length would be 4x character length, but that normally it will be under 1.2x, given we have mostly ASCII in our notifications. We have to keep in mind the message encoding overhead and not get too loose with the limits, though.

mharyam:
close #276

### DIFF
--- a/kcidb/monitor/output.py
+++ b/kcidb/monitor/output.py
@@ -5,6 +5,7 @@ import re
 import base64
 import textwrap
 import html
+import logging
 from email.message import EmailMessage
 import kcidb.oo
 from kcidb.monitor.misc import is_valid_firestore_id
@@ -15,13 +16,16 @@ from kcidb.templates import ENV as TEMPLATE_ENV
 
 # A regex matching permitted notification message subject strings
 NOTIFICATION_MESSAGE_SUBJECT_RE = re.compile(r"[^\x00-\x1f\x7f]*")
-# Maximum length of a UTF-8-encoded notification message subject
+# Maximum character length of a notification message subject
 NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN = 256
-# Maximum length of a UTF-8-encoded notification message body
+# Maximum character length of a notification message body
 NOTIFICATION_MESSAGE_BODY_MAX_LEN = 4096
 
 # A regex matching permitted subscription name strings
 SUBSCRIPTION_RE = re.compile(r"([A-Za-z0-9][A-Za-z0-9_]*)?")
+
+# Module's logger
+LOGGER = logging.getLogger(__name__)
 
 
 class NotificationMessage:
@@ -174,18 +178,22 @@ class Notification:
             self.obj.get_type().name: self.obj,
         }
         subject = TEMPLATE_ENV.from_string(self.message.subject).render(ctx)
-        subject_extra_bytes = \
-            len(subject.encode()) - NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN
-        assert subject_extra_bytes <= 0, \
-            f"Subject is {subject_extra_bytes} bytes too long"
+        subject_extra_characters = \
+            len(subject) - NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN
+        if subject_extra_characters > 0:
+            subject = subject[:NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN - 1] + "✂️"
+            LOGGER.warning("Subject is %s characters too long, truncated",
+                           subject_extra_characters)
         assert NOTIFICATION_MESSAGE_SUBJECT_RE.fullmatch(subject), \
             f"Subject is invalid, must match " \
             f"{NOTIFICATION_MESSAGE_SUBJECT_RE.pattern} regex"
         body = TEMPLATE_ENV.from_string(self.message.body).render(ctx)
-        body_extra_bytes = \
-            len(body.encode()) - NOTIFICATION_MESSAGE_BODY_MAX_LEN
-        assert body_extra_bytes <= 0, \
-            f"Body is {body_extra_bytes} bytes too long"
+        body_extra_characters = \
+            len(body) - NOTIFICATION_MESSAGE_BODY_MAX_LEN
+        if body_extra_characters > 0:
+            body = body[:NOTIFICATION_MESSAGE_BODY_MAX_LEN - 1] + "✂️"
+            LOGGER.warning("Body is %s characters too long, truncated",
+                           body_extra_characters)
 
         # Generate the plain-text message
         email = EmailMessage()


kcidb/monitor/output.py
```diff
@@ -176,16 +179,18 @@ def render(self):
...
             f"Subject is invalid, must match " \
             f"{NOTIFICATION_MESSAGE_SUBJECT_RE.pattern} regex"
         body = TEMPLATE_ENV.from_string(self.message.body).render(ctx)
         body_extra_bytes = \
             len(body.encode()) - NOTIFICATION_MESSAGE_BODY_MAX_LEN
-        assert body_extra_bytes <= 0, \
-            f"Body is {body_extra_bytes} bytes too long"
+        if body_extra_bytes > 0:
+            body = body[:NOTIFICATION_MESSAGE_BODY_MAX_LEN]
+            LOGGER.warning("Body is %s bytes too long", body_extra_bytes)
```
spbnick, line 193: You need to from limiting by bytes to limiting by characters before you can meaningfully truncate messages. See how those differ e.g. in https://en.wikipedia.org/wiki/UTF-8#Encoding

kcidb/monitor/output.py
```diff
@@ -174,18 +177,22 @@ def render(self):
...
         subject = TEMPLATE_ENV.from_string(self.message.subject).render(ctx)
-        subject_extra_bytes = \
-            len(subject.encode()) - NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN
-        assert subject_extra_bytes <= 0, \
-            f"Subject is {subject_extra_bytes} bytes too long"
+        subject_extra_characters = \
+            len(subject) - NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN
+        if subject_extra_characters > 0:
+            subject = subject[:NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN]
+            LOGGER.warning("Subject is %s characters too long",
```
spbnick, line 184: Could you note what's being done about the subject, in the log message? Like this:
```suggestion
            LOGGER.warning("Subject is %s characters too long, truncated",
```
And same for the body message?

kcidb/monitor/output.py
```diff
@@ -174,18 +177,22 @@ def render(self):
...
         }
         subject = TEMPLATE_ENV.from_string(self.message.subject).render(ctx)
-        subject_extra_bytes = \
-            len(subject.encode()) - NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN
-        assert subject_extra_bytes <= 0, \
-            f"Subject is {subject_extra_bytes} bytes too long"
+        subject_extra_characters = \
+            len(subject) - NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN
+        if subject_extra_characters > 0:
+            subject = subject[:NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN]
```
spbnick, line 183: It would be good to leave an indication to the notification recipient, that the subject was truncated. E.g. by adding some sort of character on the end, say "✂️". However, if we add that character after truncating on the limit, the subject will go over the limit. So we need to truncate to the limit, minus one, and then add the character. Could you do that? And same for the body?

spbnick (changes requested):
Looks pretty good, thank you, @mharyam :relaxed:

We're lumping two changes into one commit here: switching to limiting by characters, and truncating the message. However, since the change is so small, we can get away with just one commit, but we have to note both of these in the commit message.

Just `output:` in the commit message is missing the outer context. If you want to mention "output", you gotta mention the "monitor" as well, something like "monitor: output:". However, I think just "monitor:" would be enough. Oh, and please don't put space before the colon.

Overall, something like this commit message would work better: "monitor: Truncate notification subject and message to char limit"

mharyam:
> Looks pretty good, thank you, @mharyam ☺️
> 
> We're lumping two changes into one commit here: switching to limiting by characters, and truncating the message. However, since the change is so small, we can get away with just one commit, but we have to note both of these in the commit message.
> 
> Just `output:` in the commit message is missing the outer context. If you want to mention "output", you gotta mention the "monitor" as well, something like "monitor: output:". However, I think just "monitor:" would be enough. Oh, and please don't put space before the colon.
> 
> Overall, something like this commit message would work better: "monitor: Truncate notification subject and message to char limit"

Okay I will change it 
Thank you

kcidb/monitor/output.py
```diff
@@ -174,18 +177,24 @@ def render(self):
...
         subject = TEMPLATE_ENV.from_string(self.message.subject).render(ctx)
-        subject_extra_bytes = \
-            len(subject.encode()) - NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN
-        assert subject_extra_bytes <= 0, \
-            f"Subject is {subject_extra_bytes} bytes too long"
+        subject_extra_characters = \
+            len(subject) - NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN
+        if subject_extra_characters > 0:
+            subject = subject[:NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN-1]
+            subject += "\u2702"
```
spbnick, line 184: If you just put the literal character in here, instead of the escape sequence, the code will fit into one line. Also, please put spaces around that minus character. Something like this:
```suggestion
            subject = subject[:NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN - 1] + "✂️"
```
It's also easier to understand what's going on, if it's the character directly.

BTW, your code is adding the [scissors character](https://codepoints.net/U+2702) (`\u2702`), not the [scissors emoji](https://emojipedia.org/emoji/%E2%9C%82/) (`\u2702\ufe0f`), which is the same character, but followed by emoji [variation selector](https://en.wikipedia.org/wiki/Variation_Selectors_(Unicode_block)) (`\ufe0f`). GitHub is rendering them the same in comments for some reason, but they are not. Is that on purpose?

Same goes for the body, of course.

spbnick (changes requested):
Almost there, just have one small functional and one style comment. Thank you!

kcidb/monitor/output.py
mharyam: No, it is not 
I will fix it

kcidb/monitor/output.py
```diff
@@ -174,18 +178,22 @@ def render(self):
...
+            subject = subject[:NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN - 1] + "✂️"
+            LOGGER.warning("Subject is %s characters too long, truncated",
+                           subject_extra_characters)
         assert NOTIFICATION_MESSAGE_SUBJECT_RE.fullmatch(subject), \
             f"Subject is invalid, must match " \
             f"{NOTIFICATION_MESSAGE_SUBJECT_RE.pattern} regex"
         body = TEMPLATE_ENV.from_string(self.message.body).render(ctx)
-        body_extra_bytes = \
+        body_extra_characters = \
             len(body.encode()) - NOTIFICATION_MESSAGE_BODY_MAX_LEN
```
spbnick, line 192: This is still calculating byte length, not character length. Please correct.

spbnick (changes requested):
Ah, one functional comment this time, for a spot I missed before. One small style comment as well. Thank you, Maryam!

mharyam:
> Ah, one functional comment this time, for a spot I missed before. One small style comment as well. Thank you, Maryam!

@spbnick 
I have updated the comment 
Kindly review

spbnick (approved):
Looks great, thank you, Maryam! Merging.

There's still one change left to fix #276: replacing invalid characters in the subject, and removing the remaining `assert` in `render()`, but it could go into a separate PR.

PR #288, opened 2022-01-31
monitor: Remove invalid notification subject character

mharyam:
concerns #276

### DIFF
--- a/kcidb/monitor/output.py
+++ b/kcidb/monitor/output.py
@@ -14,8 +14,11 @@ from kcidb.templates import ENV as TEMPLATE_ENV
 # We like the "id" name
 # pylint: disable=invalid-name
 
+# Characters not allowed in notifications message subjects
+NOTIFICATION_MESSAGE_SUBJECT_INVALID_CHARS = r"\x00-\x1f\x7f"
 # A regex matching permitted notification message subject strings
-NOTIFICATION_MESSAGE_SUBJECT_RE = re.compile(r"[^\x00-\x1f\x7f]*")
+NOTIFICATION_MESSAGE_SUBJECT_RE = \
+    re.compile(f"[^{NOTIFICATION_MESSAGE_SUBJECT_INVALID_CHARS}]*")
 # Maximum character length of a notification message subject
 NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN = 256
 # Maximum character length of a notification message body
@@ -184,9 +187,12 @@ class Notification:
             subject = subject[:NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN - 2] + "✂️"
             LOGGER.warning("Subject is %s characters too long, truncated",
                            subject_extra_characters)
-        assert NOTIFICATION_MESSAGE_SUBJECT_RE.fullmatch(subject), \
-            f"Subject is invalid, must match " \
-            f"{NOTIFICATION_MESSAGE_SUBJECT_RE.pattern} regex"
+        if not NOTIFICATION_MESSAGE_SUBJECT_RE.fullmatch(subject):
+            subject = re.sub(f"[{NOTIFICATION_MESSAGE_SUBJECT_INVALID_CHARS}]",
+                             "⯑", subject)
+            LOGGER.warning("Subject is invalid, must match %s"
+                           "regex, invalid characters removed",
+                           NOTIFICATION_MESSAGE_SUBJECT_RE.pattern)
         body = TEMPLATE_ENV.from_string(self.message.body).render(ctx)
         body_extra_characters = \
             len(body) - NOTIFICATION_MESSAGE_BODY_MAX_LEN
--- a/kcidb/monitor/output.py
+++ b/kcidb/monitor/output.py
@@ -181,7 +181,7 @@ class Notification:
         subject_extra_characters = \
             len(subject) - NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN
         if subject_extra_characters > 0:
-            subject = subject[:NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN - 1] + "✂️"
+            subject = subject[:NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN - 2] + "✂️"
             LOGGER.warning("Subject is %s characters too long, truncated",
                            subject_extra_characters)
         assert NOTIFICATION_MESSAGE_SUBJECT_RE.fullmatch(subject), \
@@ -191,7 +191,7 @@ class Notification:
         body_extra_characters = \
             len(body) - NOTIFICATION_MESSAGE_BODY_MAX_LEN
         if body_extra_characters > 0:
-            body = body[:NOTIFICATION_MESSAGE_BODY_MAX_LEN - 1] + "✂️"
+            body = body[:NOTIFICATION_MESSAGE_BODY_MAX_LEN - 2] + "✂️"
             LOGGER.warning("Body is %s characters too long, truncated",
                            body_extra_characters)
 
--- a/kcidb/monitor/test_output.py
+++ b/kcidb/monitor/test_output.py
@@ -202,3 +202,83 @@ class NotificationTestCase(unittest.TestCase):
         self.assertEqual(message.get("Subject")[-2:], "✂️")
         self.assertEqual(len(message.get_body('plain').get_content()[:-1]), NOTIFICATION_MESSAGE_BODY_MAX_LEN)
         self.assertEqual((message.get_body('plain').get_content())[-3:-1], "✂️")
+
+    def test_subject_invalid_character(self):
+        """Check subject invalid character of Notification """
+
+        db_client = db.Client("sqlite::memory:")
+        db_client.init()
+        oo_client = oo.Client(db_client)
+        db_client.load({
+            "version": self.version,
+            "checkouts": [
+                {
+                    "contacts": [
+                        "rdma-dev-team@redhat.com"
+                    ],
+                    "start_time": "2020-03-02T15:16:15.790000+00:00",
+                    "git_repository_branch": "wip/jgg-for-next",
+                    "git_commit_hash": "5e29d1443c46b6ca70a4c940a67e8c09f05dcb7e",
+                    "patchset_hash": "",
+                    "git_repository_url": "https://git.kernel.org/pub/scm/linux/kernel/git/arm64/linux.git",
+                    "misc": {
+                        "pipeline_id": 467715
+                    },
+                    "id": "origin:1",
+                    "origin": "origin",
+                    "patchset_files": [],
+                    "valid": True,
+                },
+            ],
+            "builds": [
+                                {
+                    "id": "kernelci:kernelci.org:619fecfc6d932a49b5f2efb0",
+                    "checkout_id": "origin:1",
+                    "origin": "origin",
+                    "comment": "v5.16-rc2-19-g383a44aec91c",
+                    "start_time": "2021-11-25T20:07:24.499000+00:00",
+                    "duration": 486.695294857,
+                    "architecture": "x86_64",
+                    "config_name": "x86_64_defconfig+x86-chromebook+kselftest",
+                    "valid": True
+                },
+            ]
+        })
+        revision = oo_client.query(orm.Pattern.parse(">revision#"))["revision"][0]
+
+        # test invalid character in subject are replaced
+        # by the uncertainty sign character in the email
+        notification_message = NotificationMessage(
+            ["kernelci-results-staging@groups.io"],
+            subject="Test completed for linux.git:\n@commit20-\x7f\t",
+            body="Below is the summary of results Kernel CI database has recorded for this revision so far."
+        )
+        notification = Notification(revision, "test_subscription", notification_message)
+        message = notification.render()
+        self.assertIsInstance(message, email.message.EmailMessage)
+        self.assertEqual(len(message.get("Subject")), 42)
+        self.assertEqual(message.get("Subject")[29], "⯑")
+        self.assertEqual(message.get("Subject")[-2:], "⯑⯑")
+
+        # test empty subject passes unchanged in the email
+        notification_message = NotificationMessage(
+            ["kernelci-results-staging@groups.io"],
+            subject="",
+            body="Below is the summary of results."
+        )
+        notification = Notification(revision, "test_subscription2", notification_message)
+        message = notification.render()
+        self.assertIsInstance(message, email.message.EmailMessage)
+        self.assertEqual(len(message.get("Subject")), 0)
+
+        # test subject with all valid characters passes unchanged in the email
+        notification_message = NotificationMessage(
+            ["kernelci-results-staging@groups.io"],
+            subject="Testing for this Subject.😀😀",
+            body="Below is the summary of results."
+        )
+        notification = Notification(revision, "test_subscription2", notification_message)
+        message = notification.render()
+        self.assertIsInstance(message, email.message.EmailMessage)
+        self.assertEqual(len(message.get("Subject")), 27)
+        self.assertNotIn(message.get("Subject"), "⯑")
--- a/kcidb/monitor/test_output.py
+++ b/kcidb/monitor/test_output.py
@@ -4,7 +4,9 @@ import email
 import unittest
 from kcidb_io import schema
 from kcidb import orm, db, oo
-from kcidb.monitor.output import NotificationMessage, Notification
+from kcidb.monitor.output import NotificationMessage, Notification,\
+    NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN, NOTIFICATION_MESSAGE_BODY_MAX_LEN
+
 
 # Disable long line checking for JSON data
 # flake8: noqa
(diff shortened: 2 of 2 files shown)


kcidb/monitor/output.py
```diff
@@ -184,9 +185,11 @@ def render(self):
             subject = subject[:NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN - 1] + "✂️"
             LOGGER.warning("Subject is %s characters too long, truncated",
                            subject_extra_characters)
-        assert NOTIFICATION_MESSAGE_SUBJECT_RE.fullmatch(subject), \
-            f"Subject is invalid, must match " \
-            f"{NOTIFICATION_MESSAGE_SUBJECT_RE.pattern} regex"
+        if not NOTIFICATION_MESSAGE_SUBJECT_RE.fullmatch(subject):
```
mharyam, line 188: @spbnick 
I am thinking of making this come before truncating the subject. 
What do you think?

kcidb/monitor/output.py
```diff
@@ -184,9 +185,11 @@ def render(self):
...
+            subject = re.sub('[^A-Za-z0-9]+', ' ', subject).strip()
```
spbnick, line 189: This would replace any number of consecutive non-Latin and non-digit characters with a single space. This means that text in any language not using Latin characters will be replaced with a single space, which we don't want to do. This would also make it difficult to determine how many characters were there, and make debugging harder.

Instead, we should replace exactly the inverse of permitted characters, and replace them one-to-one.

Also, it's better to use a character which would clearly signify that something was replaced. There's the old "[substitute character](https://en.wikipedia.org/wiki/Substitute_character)" in ASCII and Unicode, but it's also a control character and could be interpreted differently. There's the Unicode "[replacement character](https://en.wikipedia.org/wiki/Specials_(Unicode_block)#Replacement_character)", but it normally signifies that there was something wrong in the Unicode encoding. I couldn't find anything better than the "[uncertainty sign](https://codepoints.net/U+2BD1?lang=en)" with an uncertain (ha-ha) purpose, which looks like the "replacement character", but is white rather than black. It will have to do. Could you use that instead as the replacement?

To make sure our "correct" and "incorrect" character patterns are always exactly the inversion of each other, we can generate one from another.

We can e.g. define a new variable called `NOTIFICATION_MESSAGE_SUBJECT_INVALID_CHARS` specifying invalid characters, and set it to `\x00-\x1f\x7f`. Then we can generate both regular expressions from that, to make sure they are always the opposites of each other. Can you do that?

spbnick (changes requested):

kcidb/monitor/output.py
spbnick: That would be correct if we were changing the length of the string. However, as I suggest in my review we shouldn't be doing that. Still conceptually it would make better sense to the reader if we did that first, so please go ahead in any case :smile: Thanks!

mharyam: Thanks @spbnick 
Since we are replacing the character I will just leave it there 
I thought we will be removing the unwanted character

kcidb/monitor/output.py
```diff
@@ -32,6 +35,7 @@ class NotificationMessage:
     """
     Message for a notification about a report object state.
     """
+
```
spbnick, line 38: Is adding this new line necessary? If not, please remove.

spbnick (changes requested):
Looks great, just one empty line comment.

However, do you think you could write a test for this feature?

And perhaps one more test for the previous change you did, in a separate commit before this one?

Thank you.

mharyam:
> Looks great, just one empty line comment.
> 
> However, do you think you could write a test for this feature?
> 
> And perhaps one more test for the previous change you did, in a separate commit before this one?
> 
> Thank you.

Okay, I will add them

kcidb/monitor/test_output.py
```diff
@@ -85,3 +85,57 @@ def test_min(self):
...
+        notification.obj.git_commit_name = "Set an invalid data with length greater " \
+                                           "than notification subject max length" * 40
+        notification.obj.git_commit_hash = "Set an invalid commit hash with length greater " \
+                                           "than notification body max length" * 250
+        self.assertEqual(notification.id,
+                         'mark_brown:revision:'
+                         'KCc1ZTI5ZDE0NDNjNDZiNmNhNzBhNGM5NDBhNj'
+                         'dlOGMwOWYwNWRjYjdlJywgJycp:')
+        message = notification.render()
+        self.assertIsInstance(message, email.message.EmailMessage)
```
spbnick, line 141: This is a pretty good start. However, you don't actually check that either the subject or the body length was (not)limited.

To test this well, you would need to check these for both subject and body:

- [ ] Zero-length NotificationMessage subject/body produces zero length subject and body in the email, without failing
- [ ] Under-limit length subject/body are left intact in the email
- [ ] Exact-limit length subject/body are left intact in the email, with all characters being single-byte in UTF-8
- [ ] Exact-limit length subject/body are left intact in the email, with at least one character being multi-byte in UTF-8 (verifying that we're limiting by characters, not by bytes)
- [ ] Over-limit length subject/body are trimmed to the exact limit with the last character replaced with the scissors emoji in the email

Also, you don't really need to involve subscriptions to create a notification. Having them involved makes it harder to craft the notifications you need for testing. Instead you can create notification objects directly. Something like this:
```python
        db_client = db.Client("sqlite::memory:")
        db_client.init()
        db_client.load(io_data)
        oo_client = oo.Client(db_client)
        revision = oo_client.query(orm.Pattern.parse(">revision#"))["revision"][0]
        message = NotificationMessage(
            ["kernelci-results-staging@groups.io"],
            subject="JINJA2_TEMPLATE_FOR_YOUR_SUBJECT",
            body="JINJA2_TEMPLATE_FOR_YOUR_BODY"
        )
        notification = Notification(revision, "test_subscription", message)
        email = notification.email()
```

spbnick (changes requested):

kcidb/monitor/test_output.py
```diff
@@ -85,3 +85,89 @@ def test_min(self):
...
+                    "config_name": "x86_64_defconfig+x86-chromebook+kselftest",
+                    "valid": True
+                },
+            ]
+        })
+        revision = oo_client.query(orm.Pattern.parse(">revision#"))["revision"][0]
+        message = NotificationMessage(
+            ["kernelci-results-staging@groups.io"],
+            subject="JINJA2_TEMPLATE_FOR_YOUR_SUBJECT",
+            body="JINJA2_TEMPLATE_FOR_YOUR_BODY"
```
spbnick, line 134: Please don't use the literal placeholders I put into the example :joy: I was only trying to illustrate the point for you, pointing out where subject/body go, and that these strings are in fact Jinja2 templates. Please put something less confusing there.

I find it useful to use strings of digits as samples when testing for length. Something like `"1234567890"` or `"123456789012345678901234"`. This way it's easier to see that the first string is ten characters long and the second is 24 characters. But pick whatever feels best for you.

kcidb/monitor/test_output.py
```diff
@@ -85,3 +85,89 @@ def test_min(self):
...
+        self.assertIsInstance(message, email.message.EmailMessage)
+        self.assertEqual(len(message.get("Subject", "")), 32)
+        self.assertEqual(len(message.get_body('plain').get_content()[:-1]), 29)
+
+        empty_message = NotificationMessage(
+            ["kernelci-results-staging@groups.io"],
+            subject="", body="")
+        empty_notification = Notification(revision, "test_subscription_2", empty_message)
+        empty_message = empty_notification.render()
+        self.assertIsInstance(empty_message, email.message.EmailMessage)
```
spbnick, line 147: I don't think we need to check that `render()` returns an `EmailMessage` in every test. We could save a few lines of code, if we do it only once.

kcidb/monitor/test_output.py
```diff
@@ -85,3 +85,89 @@ def test_min(self):
...
+                    "comment": "v5.16-rc2-19-g383a44aec91c",
+                    "start_time": "2021-11-25T20:07:24.499000+00:00",
+                    "duration": 486.695294857,
+                    "architecture": "x86_64",
+                    "config_name": "x86_64_defconfig+x86-chromebook+kselftest",
+                    "valid": True
+                },
+            ]
+        })
+        revision = oo_client.query(orm.Pattern.parse(">revision#"))["revision"][0]
```
spbnick, line 130: Since you're putting multiple tests in the same function, could you add an empty line after this one, to separate the common setup from the tests, and then add comments before creation of each message and its block of assertions, saying what you're testing there?

kcidb/monitor/test_output.py
```diff
@@ -85,3 +85,89 @@ def test_min(self):
...
+        self.assertEqual(len(message.get_body('plain').get_content()[:-1]), 29)
+
+        empty_message = NotificationMessage(
+            ["kernelci-results-staging@groups.io"],
+            subject="", body="")
+        empty_notification = Notification(revision, "test_subscription_2", empty_message)
+        empty_message = empty_notification.render()
+        self.assertIsInstance(empty_message, email.message.EmailMessage)
+        self.assertEqual(len(empty_message.get("Subject", "")), 0)
+        self.assertEqual(len(empty_message.get_body('plain').get_content()[:-1]), 0)
```
spbnick, line 149: Hmm, curious, why do you need to cut off the last character? Something to do with the way messages are encoded? What is that character?

kcidb/monitor/test_output.py
```diff
@@ -85,3 +85,89 @@ def test_min(self):
...
+
+        exact_message = NotificationMessage(
+            ["kernelci-results-staging@groups.io"],
+            subject="JINJA2_TEMPLATE_FOR_YOUR_SUBJECT" * 8,
+            body="JINJA2_TEMPLATE_FOR_YOUR_BODY___" * 128
```
spbnick, line 154: Could you use the actual variables defining the limits here and below, instead of hard-coding them? This way we won't  have to update the tests when/if we change them.

kcidb/monitor/test_output.py
```diff
@@ -85,3 +85,89 @@ def test_min(self):
...
+        )
+        exact_notification = Notification(revision, "test_subscription_3", exact_message)
+        exact_message = exact_notification.render()
+        self.assertIsInstance(exact_message, email.message.EmailMessage)
+        self.assertEqual(len(exact_message.get("Subject", "")), 256)
```
spbnick, line 159: Hmm, why do you need to provide a default for the Subject here?

kcidb/monitor/test_output.py
```diff
@@ -85,3 +85,89 @@ def test_min(self):
...
+        self.assertEqual(len(exact_message.get_body('plain').get_content()[:-1]), 4096)
```
spbnick, line 160: Apart from the hardcoded limits, and my questions on the subject's default and cutting off the last character, this one is good, but we also need the same test with multi-byte characters.

spbnick (changes requested):
Looking good, but still needs some improvements, one more test, and I have a couple questions.

Thank you :smiley:

kcidb/monitor/test_output.py
mharyam: Yes, it is 
It always adds a new line character at the end of each message

kcidb/monitor/test_output.py
mharyam: I will remove this, wanted to avoid if the subject comes back empty 😃 
But it is not necessary since I am providing it in the test

kcidb/monitor/test_output.py
spbnick: Hmm... interesting. Keep removing it for now, but it would be good to find out where that happens.

kcidb/monitor/test_output.py
spbnick: Yeah, it's better to remove it, or it could mask a bug where Subject would not be set.

kcidb/monitor/test_output.py
mharyam: Yh,  I tried checking and noticed that it gets added by default when python returns contents from the get_content function

kcidb/monitor/test_output.py
```diff
@@ -85,3 +87,123 @@ def test_min(self):
...
+        empty_notification = Notification(revision, "test_subscription_2", empty_message)
+        empty_message = empty_notification.render()
+        self.assertEqual(len(empty_message.get("Subject")), 0)
+        self.assertEqual(len(empty_message.get_body('plain').get_content()[:-1]), 0)
+
+        # test exact-limit length subject/body are left intact in the email,
+        exact_message = NotificationMessage(
+            ["kernelci-results-staging@groups.io"],
+            subject="Test completed for linux.git:123" * 8,
+            body="Below is the summary of results Kernel CI database has recorded " * 64
```
spbnick, line 159: Could you generate the input strings using the `*_MAX_LEN` constants as well, here and below? Something like `subject="S" * NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN, body="B" * NOTIFICATION_MESSAGE_BODY_MAX_LEN`?

kcidb/monitor/test_output.py
```diff
@@ -85,3 +87,123 @@ def test_min(self):
...
+        self.assertEqual(len(exact_message.get("Subject")), NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN)
+        self.assertEqual(len(exact_message.get_body('plain').get_content()[:-1]),
+                         NOTIFICATION_MESSAGE_BODY_MAX_LEN)
+
+        # test exact-limit length with a multibyte character in UTF-8
+        # subject/body are left intact in the email,
+        exact_message_with_multibyte = NotificationMessage(
+            ["kernelci-results-staging@groups.io"],
+            subject="Test completed for linux.git:©12" * 8,
+            body="Below is the summary of results Kernel CI database has recorded•" * 64
```
spbnick, line 172: Here you can just use a pair of multibyte characters instead of `"S"` and `"B"`.

kcidb/monitor/test_output.py
```diff
@@ -4,11 +4,13 @@
...
 from kcidb import orm, db, oo
-from kcidb.monitor.output import NotificationMessage, Notification
+from kcidb.monitor.output import NotificationMessage, Notification,\
+    NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN, NOTIFICATION_MESSAGE_BODY_MAX_LEN
+
 
 # Disable long line checking for JSON data
 # flake8: noqa
-# pylint: disable=line-too-long
+# pylint: disable=line-too-long : disable-msg=too-many-locals
```
spbnick, line 13: You can just use one variable each for `NotificationMessage`, `Notification`, and `EmailMessage` throughout the code, now that you have comments explaining what each setup/assertion block does, and reuse them between blocks. This way you'll likely not need to disable this warning. But please don't use the same variable to store both `NotificationMessage` and `EmailMessage` objects, like you do now, that could be confusing.

spbnick (changes requested):
Hey, this is pretty good, and I think we already got most of the functionality right. Only a few minor comments remain. Thank you!

kcidb/monitor/test_output.py
```diff
@@ -207,3 +207,57 @@ def test_subject_and_body_length(self):
...
+            ["kernelci-results-staging@groups.io"],
+            subject="Test completed for linux.git:\n@commit20\t",
+            body="Below is the summary of results Kernel CI database has recorded for this revision so far."
+        )
+        notification = Notification(revision, "test_subscription", message)
+        message = notification.render()
+        self.assertIsInstance(message, email.message.EmailMessage)
+        self.assertEqual(len(message.get("Subject")), 40)
+        self.assertEqual(message.get("Subject")[29], "�")
+        self.assertEqual(message.get("Subject")[-1], "�")
```
spbnick, line 263: We also need the following tests to do this well:

- [ ] Empty subject passes unchanged (I know we have this in length-limit test, but this one needs a copy).
- [ ] Non-empty subject without characters valid (including non-Latin/non-ASCII characters) passes unchanged.

Plus it would be good to add `\x7f` to the sample with invalid characters, as that should be removed too.

kcidb/monitor/output.py
```diff
@@ -184,9 +187,12 @@ def render(self):
             subject = subject[:NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN - 2] + "✂️"
             LOGGER.warning("Subject is %s characters too long, truncated",
                            subject_extra_characters)
-        assert NOTIFICATION_MESSAGE_SUBJECT_RE.fullmatch(subject), \
-            f"Subject is invalid, must match " \
-            f"{NOTIFICATION_MESSAGE_SUBJECT_RE.pattern} regex"
+        if not NOTIFICATION_MESSAGE_SUBJECT_RE.fullmatch(subject):
+            subject = re.sub(f"[{NOTIFICATION_MESSAGE_SUBJECT_INVALID_CHARS}]",
+                             "�", subject)
```
spbnick, line 192: Ah, wait, this is the "replacement character". We need to use the "uncertainty sign" character instead: `⯑`. The former indicates something wrong with Unicode encoding, and would be confusing.

mharyam:
@spbnick 
All done 😄 
Kindly review

kcidb/monitor/test_output.py
```diff
@@ -202,3 +202,83 @@ def test_subject_and_body_length(self):
...
+        )
+        notification = Notification(revision, "test_subscription2", notification_message)
+        message = notification.render()
+        self.assertIsInstance(message, email.message.EmailMessage)
+        self.assertEqual(len(message.get("Subject")), 0)
+
+        # test subject with all valid characters passes unchanged in the email
+        notification_message = NotificationMessage(
+            ["kernelci-results-staging@groups.io"],
+            subject="Testing for this Subject.",
```
spbnick, line 277: Could you add some national characters here, or an emoji, something outside the Latin/ASCII character set?

spbnick (changes requested):
This is great, thank you, Maryam :) I have just one small request left inline which you probably missed in my yesterday's review.

kcidb/monitor/test_output.py
mharyam: oh, yes I missed it 
But I have added it now.

spbnick (approved):
Looks wonderful :heart_eyes: Thank you very much, Maryam! Merging :smile:

PR #289, opened 2022-01-31
Tie KCIDB to V4 i/o schema

### LINKED ISSUE #155: Consider tying KCIDB to a particular I/O schema
At the moment a big part of KCIDB code, as well as tests is simply bound to the latest I/O schema. This makes updating KCIDB-IO rather involved, and precludes integration testing of KCIDB-IO and KCIDB if a new schema is introduced. Instead, explicitly specify and use a particular I/O schema throughout the code, and update it whenever ready.

This would allow using and testing KCIDB with KCIDB-IO, regardless of whether a new I/O schema version is being worked on.

mharyam:
Close #155

### DIFF
--- a/kcidb/__init__.py
+++ b/kcidb/__init__.py
@@ -41,7 +41,7 @@ class Client:
             `kcidb.DatabaseNotInitialized` if the database is not
             initialized.
             `kcidb.db.IncompatibleSchema` if the database schema
-            is incompatible with the latest I/O schema.
+            is incompatible with the current I/O schema.
         """
         assert database is None or \
             isinstance(database, str) and database
@@ -65,7 +65,7 @@ class Client:
 
         Args:
             data:   The JSON report data to submit.
-                    Must adhere to the latest version of I/O schema.
+                    Must adhere to the current version of I/O schema.
 
         Returns:
             Submission ID string.
@@ -74,7 +74,7 @@ class Client:
             `NotImplementedError`, if not supplied with a project ID or an MQ
             topic name at initialization time.
         """
-        assert LIGHT_ASSERTS or io.schema.is_valid_latest(data)
+        assert LIGHT_ASSERTS or io.SCHEMA.is_valid_exactly(data)
         if not self.mq_publisher:
             raise NotImplementedError
         return self.mq_publisher.publish(data)
@@ -85,7 +85,7 @@ class Client:
 
         Args:
             data_iter:  An iterator returning the JSON report data to submit.
-                        Each must adhere to the latest version of I/O schema.
+                        Each must adhere to the current version of I/O schema.
             done_cb:    A function to call when a report is successfully
                         submitted. Will be called with the submission ID of
                         each report returned by the iterator, in order.
@@ -117,7 +117,7 @@ class Client:
                                 returned report, or zero for no limit.
 
         Returns:
-            An iterator returning report JSON data adhering to the latest I/O
+            An iterator returning report JSON data adhering to the current I/O
             schema version, each containing at most the specified number of
             objects.
 
@@ -155,7 +155,7 @@ class Client:
                         as well.
 
         Returns:
-            The fetched JSON data adhering to the latest I/O schema version.
+            The fetched JSON data adhering to the current I/O schema version.
 
         Raises:
             `NotImplementedError`, if not supplied with a dataset name at
@@ -174,7 +174,7 @@ class Client:
         else:
             raise NotImplementedError
 
-        assert LIGHT_ASSERTS or io.schema.is_valid_latest(data)
+        assert LIGHT_ASSERTS or io.SCHEMA.is_valid_exactly(data)
         return data
 
 
@@ -202,7 +202,7 @@ def submit_main():
         sys.stdout.flush()
 
     client.submit_iter(
-        (io.schema.upgrade(io.schema.validate(data), copy=False)
+        (io.SCHEMA.upgrade(io.SCHEMA.validate(data), copy=False)
          for data in misc.json_load_stream_fd(sys.stdin.fileno())),
         done_cb=print_submission_id
     )
@@ -232,7 +232,7 @@ def query_main():
 def schema_main():
     """Execute the kcidb-schema command-line tool"""
     sys.excepthook = misc.log_and_print_excepthook
-    description = 'kcidb-schema - Output latest or older I/O JSON schema'
+    description = 'kcidb-schema - Output current or older I/O JSON schema'
     parser = misc.OutputArgumentParser(description=description)
     misc.argparse_schema_add_args(parser, "output")
     args = parser.parse_args()
@@ -259,13 +259,13 @@ def validate_main():
 def upgrade_main():
     """Execute the kcidb-upgrade command-line tool"""
     sys.excepthook = misc.log_and_print_excepthook
-    description = 'kcidb-upgrade - Upgrade I/O JSON data to latest schema'
+    description = 'kcidb-upgrade - Upgrade I/O JSON data to current schema'
     parser = misc.OutputArgumentParser(description=description)
     misc.argparse_schema_add_args(parser, "upgrade")
     args = parser.parse_args()
     misc.json_dump_stream(
         (
-            args.schema_version.upgrade(io.schema.validate(data), copy=False)
+            args.schema_version.upgrade(io.SCHEMA.validate(data), copy=False)
             for data in misc.json_load_stream_fd(sys.stdin.fileno())
         ),
         sys.stdout, indent=args.indent, seq=args.seq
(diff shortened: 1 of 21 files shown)


kcidb/__init__.py
```diff
@@ -14,6 +14,9 @@
 # Compatibility alias
 io_schema = io.schema
 
+# latest kcidb-io version
+io_latest_version = io_schema.V4
```
spbnick, line 18: Now that we've merged the base work both in kcidb-io and kcidb, please rebase your PR and use `kcidb.io.SCHEMA` to refer to the currently-used schema instead.

test_kcidb.py
```diff
@@ -242,10 +244,10 @@ def test_upgrade_main(self):
                             stderr_re=".*Invalid major version number: '0'\n")
         self.assertExecutes(latest_version,
                             "kcidb.upgrade_main",
-                            str(kcidb_io.schema.LATEST.major + 1),
+                            str(io_latest_version.major + 1),
                             status=2,
                             stderr_re=".*No schema version found for major "
-                            f"number {kcidb_io.schema.LATEST.major + 1}\n")
+                            f"number {io_latest_version.major + 1}\n")
```
spbnick, line 250: There's plenty more references to the latest schema in the kcidb code, both direct (through `kcidb_io.schema.LATEST`) and indirect (e.g. through calling functions in `kcidb_io` and `kcidb_io.schema`, which all refer to the latest schema). We gotta update all these cases.

spbnick (changes requested):

test_kcidb.py
mharyam: Okay, I will check them and update

mharyam: @spbnick 
Kindly review my changes

kcidb/db/bigquery/__init__.py
```diff
@@ -133,8 +134,8 @@ def init(self):
                 f" FROM `{table_ref}` GROUP BY id"
             self.client.create_table(view)
         # Set dataset schema version
-        self.dataset.labels["version_major"] = str(io.schema.LATEST.major)
-        self.dataset.labels["version_minor"] = str(io.schema.LATEST.minor)
+        self.dataset.labels["version_major"] = str(IO_LATEST_VERSION.major)
+        self.dataset.labels["version_minor"] = str(IO_LATEST_VERSION.minor)
```
spbnick, line 138: These and similar edits in this file would be incorrect, because we will not necessarily be using the latest schema in the driver, just the one all of kcidb is currently using. So, these should be changed to just `io.SCHEMA` directly.

kcidb/db/sqlite/__init__.py
```diff
@@ -122,8 +123,8 @@ def init(self):
                         raise Exception(
                             f"Failed creating table {name!r}"
                         ) from exc
-                version = int(io.schema.LATEST.major * 1000 +
-                              io.schema.LATEST.minor % 1000)
+                version = int(IO_LATEST_VERSION.major * 1000 +
+                              IO_LATEST_VERSION.minor % 1000)
```
spbnick, line 127: Same here regarding calling it `IO_LATEST_VERSION`. Besides, it's not the version of "I/O", it's the version of "I/O schema".

kcidb/misc.py
```diff
@@ -263,7 +263,7 @@ def schema_version(string):
                 f"Invalid major version number: {string!r}"
             )
         major = int(string)
-        version = io.schema.LATEST
+        version = SCHEMA
```
spbnick, line 266: Let's keep this file similar to others and spell out `io.SCHEMA`, instead of having a direct `SCHEMA` reference.

Besides, this file is particularly low-cohesion due to its purpose (collecting bits and pieces too small for their own modules), so we have to be careful to not pollute the namespace.

kcidb/orm/__init__.py
```diff
@@ -15,7 +15,7 @@
 
 # We'll get to it, pylint: disable=too-many-lines
 
-
+IO_LATEST_VERSION = io.SCHEMA
```
spbnick, line 18: Same here about calling it `LATEST`. Please just use `io.SCHEMA`.

requirements.txt
```diff
@@ -12,4 +12,4 @@ jinja2
 python-dateutil
 cached-property
 jq@git+https://github.com/kernelci/jq.py.git@1.2.1.post1
-kcidb-io@git+https://github.com/kernelci/kcidb-io.git@v4_pre1
+kcidb-io@git+https://github.com/kernelci/kcidb-io.git
```
spbnick, line 15: Could you please make unpinning a separate commit?

spbnick (changes requested):
This is pretty good! Thank you, @mharyam.

Apart from the inline comments, we still need to update references to the functions from kcidb-io which are using the latest version of the schema implicitly. These are functions defined in [`kcidb_io`](https://github.com/kernelci/kcidb-io/blob/main/kcidb_io/__init__.py) and [`kcidb_io.schema`](https://github.com/kernelci/kcidb-io/blob/main/kcidb_io/schema/__init__.py) modules.

One test you can do locally to find out if you have something left to update is to set kcidb_io.schema.LATEST to None temporarily and run the tests. If something fails, then we certainly have something to update. Reverse might not be true, but it would still be a useful check.

kcidb/misc.py
```diff
@@ -36,7 +36,7 @@
 }
 
 # Check light assertions only, if True
-LIGHT_ASSERTS = not os.environ.get("KCIDB_HEAVY_ASSERTS", "")
+LIGHT_ASSERTS = not os.environ.get("KCIDB_HEAVY_ASSERTS", "Not empty")
```
spbnick, line 39: Could you please revert this change? This would make heavy assertions always enabled and our production deployment **very** slow. Please be careful to avoid submitting such changes. I could've easily overlooked it.

If you want to run tests, just execute `export KCIDB_HEAVY_ASSERTS=1` in your shell before that.

spbnick (changes requested):
Wow, wonderful work, @mharyam! That must've took a lot of effort and concentration to hunt all of these down.

Apart from the (important) inline comment, we still need to change "latest" to "current" in all docstrings and messages referring to the schema. E.g.:
```diff
diff --git a/kcidb/__init__.py b/kcidb/__init__.py
index 26dd4c2..c9fa454 100644
--- a/kcidb/__init__.py
+++ b/kcidb/__init__.py
@@ -65,7 +65,7 @@ class Client:
 
         Args:
             data:   The JSON report data to submit.
-                    Must adhere to the latest version of I/O schema.
+                    Must adhere to the current version of I/O schema.
 
         Returns:
             Submission ID string.
```

You can find all instances of the word "latest" in all source files using this shell command, for example:
```bash
grep -rInw --include '*.py' latest *
```

spbnick:
Oh, and we need to rebase this, now that your other PR is merged.

kcidb/misc.py
mharyam: Oh, This was a mistake 
I added it because the test was failing  😃 
WIll revert back, thanks.

mharyam:
> LIGHT_ASSERTS = not os.environ.get("KCIDB_HEAVY_ASSERTS", "Not empty")

@

> Oh, and we need to rebase this, now that your other PR is merged.

@spbnick All done, you can review it now

kcidb/orm/__init__.py
```diff
@@ -15,7 +15,6 @@
 
 # We'll get to it, pylint: disable=too-many-lines
 
-
```
spbnick, line 18: Did you need this line removed for some reason? If not, please revert.

spbnick (changes requested):
This is great! Thank you, Maryam :smiley:
I only found one more sneaky empty line getting removed (comment inline), and there's one more case of "latest" instead of "current" in `doc/submitter_guide.md`:
```patch
diff --git a/doc/submitter_guide.md b/doc/submitter_guide.md
index 7b410c1..b6002ce 100644
--- a/doc/submitter_guide.md
+++ b/doc/submitter_guide.md
@@ -104,7 +104,7 @@ Both should execute without errors and finish with zero exit status.
 -----------------------------
 
 KCIDB accepts data in JSON, described by a versioned schema. The
-`kcidb-schema` tool will output the latest schema version. However, all tools
+`kcidb-schema` tool will output the current schema version. However, all tools
 will accept data complying with older schema versions, as a rule. Any
 exceptions to that will be mentioned in KCIDB release notes. Pipe your data
 into `kcidb-validate` tool to check if it will be accepted.
```

kcidb/monitor/test_output.py
```diff
@@ -2,11 +2,10 @@
 
 import email
 import unittest
-from kcidb.io import schema
 from kcidb import orm, db, oo
 from kcidb.monitor.output import NotificationMessage, Notification,\
     NOTIFICATION_MESSAGE_SUBJECT_MAX_LEN, NOTIFICATION_MESSAGE_BODY_MAX_LEN
-
+from kcidb.io import SCHEMA
```
spbnick, line 8: Ah, there's another line being removed unnecessarily, but let's get this merged already :joy: Otherwise it's perfect :smile:

spbnick (approved):
Wonderful work! Thank you, Maryam :smiley: Merging!

kcidb/monitor/test_output.py
mharyam: Ah, I managed to skip that 😃 

Thank you

PR #291, opened 2022-02-01
Record currently-used schema version in kcidb.io module

spbnick:
Record the currently-used schema version in the new `kcidb.io` module.

Concerns #155
Provides support to #289, which can now use `kcidb.io.SCHEMA` to refer to the object for the currently-used schema version.

@mharyam, please review.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -14,7 +14,7 @@ extension-pkg-whitelist=jq
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use "--disable=all --enable=classes
 # --disable=W".
-disable=duplicate-code,fixme,too-few-public-methods,redefined-builtin
+disable=duplicate-code,fixme,too-few-public-methods,redefined-builtin,consider-using-from-import
 
 [REPORTS]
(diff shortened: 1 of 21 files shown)


mharyam (commented):
@spbnick 
I thought the update was to allow us not to specify versions when we install kcidb-io 

For example, if we do not specify any version, when we install kcdib-io it installs the latest one 

Then we have a place where we specify which version to use in kcidb(part of the update I worked on [here](https://github.com/kernelci/kcidb/pull/289)

spbnick:
> @spbnick
> I thought the update was to allow us not to specify versions when we install kcidb-io
> 
> For example, if we do not specify any version, when we install kcdib-io it installs the latest one
> 
> Then we have a place where we specify which version to use in kcidb(part of the update I worked on here

Yes, that's still the target. What do you see as contradicting that in this PR?

spbnick:
@mharyam, yes, we update the kcidb-io version here, because we're still pinned, and we need the new changes we just merged there.

mharyam (approved):
LGTM

spbnick:
Thank you, @mharyam, merged!

Issue #293, opened 2022-02-14
Expand kcidb.oo module tests


spbnick:
We have [just merged](https://github.com/kernelci/kcidb/pull/247) the first basic tests of the `kcidb.oo` module. It would be good to add more tests.

As a start, we could use the following setup:

- [x] Load the following data for a revision into an SQLite database:
  - [x] Two valid checkouts, and one invalid checkout, using the same commit hash, and with repositories "https://repo_valid", "https://repo_valid", and "https://repo_invalid", and branches "branch_valid1", "branch_valid2", and "branch_invalid" respectively.
  - [x] Two valid, and one invalid build for each valid checkout, using architectures "valid1", "valid2", and "invalid" respectively
  - [x] Two passed, and one failed tests for each valid build, using paths "pass1", "pass2", and "fail" respectively

Do the following when accessing the database through the OO client:

- [ ] Check that database objects can be traversed using the OO representation:
  - [x] Check that the revision's links can be successfully traversed, including:
    - [x] The `repo_branch_checkouts` property has all checkouts loaded into the database, indexed by their repositories and branches
    - [x] The `architecture_valid_builds` property has all six builds, indexed by their architectures and valid status
    - [x] The `builds` property has all six loaded builds
    - [x] The `tests` property has all twelve loaded tests
    - [x] The `tests_root` property returns the root test node (`Node` instance)
  - [x] Check that links of a valid checkout can be successfully traversed, including:
    - [x] The `revision` property points to the correct revision object
    - [x] The `builds` property has all three of the loaded builds
    - [x] The `tests` property has all six loaded tests
    - [x] The `tests_root` property returns the root test node (`Node` instance)
  - [x] Check that links of a valid build can be successfully traversed, including:
    - [x] The `checkout` property points to the correct checkout object
    - [x] The `tests` property has all three loaded tests
    - [x] The `tests_root` property returns the root test node (`Node` instance)
  - [x] Check that links of a test can be successfully traversed, including:
    - [x] The `build` property points to the correct build object
  - [x] Check that links of the revision's root test node (`Node` instance from revision's `tests_root`) can be successfully traversed, including:
    - [x] The `parent` property points to the revision object
    - [x] The `tests` property contains an empty array
    - [x] The `nodes` property contains three sub-nodes, with paths "pass1", "pass2", and "fail", and each of those nodes:
      - [x] Has `parent` property point to the containing node
      - [x] Has four tests with the same corresponding `path` in `tests` property
[...]

PR #296, opened 2022-03-01
tests: Add more kcidb.oo test

mrbazzan:
Part of #293

### DIFF
--- a/kcidb/oo/__init__.py
+++ b/kcidb/oo/__init__.py
@@ -419,6 +419,9 @@ class Build(Object, TestContainer):
 class Test(Object):
     """An OO-representation of a test"""
 
+    # prevent class from being collected by unittest.
+    __test__ = False
+
 
 # A map of object type names and Object-derived classes handling their data
 CLASSES = dict(
--- a/kcidb/test_oo.py
+++ b/kcidb/test_oo.py
@@ -2,6 +2,7 @@
 
 import kcidb
 from kcidb.unittest import local_only
+from kcidb.oo import Checkout, Build, Test
 
 
 @local_only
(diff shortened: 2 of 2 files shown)


kcidb/test_oo.py
```diff
@@ -203,3 +203,169 @@ def test_two_object(self):
             '>checkout#>build#>test#',
             {"test", "build", "checkout", "revision"}, 2
         )
+
+
+class KCIDBMore00ClientTestCase(kcidb.unittest.TestCase):
```
mrbazzan, line 208: I couldn't come up with variable names for the new class, so I just added *More*. That probably needs to be changed.

kcidb/test_oo.py
```diff
@@ -203,3 +203,169 @@ def test_two_object(self):
...
+                        "origin": "redhat", "status": "FAIL", "path": "fail",
+                    },
+                ]
+            }
+        )
+        self.client = kcidb.oo.Client(source)
+
+    def test_traversing_revision_link(self):
+        """Check that revision's links are successfully traversed"""
+        self.assertTrue([1])
```
mrbazzan, line 371: This is just to get the test passing.

kcidb/test_oo.py
spbnick: Since all our next tests are about traversing the objects, we can call it something like `KCIDBOOTraversingTestCase`.

kcidb/test_oo.py
```diff
@@ -203,3 +203,169 @@ def test_two_object(self):
...
+    def setUp(self):
+        source = kcidb.db.Client("sqlite::memory:")
+        source.init()
+        source.load(
+            {
+                "version": {"major": 4, "minor": 0},
+                "checkouts": [
+                    {
+                        "id": "_:kernelci:5acb9c2a7bc836e"
+                              "9e5172bbcd2311499c5b4e5f1",
```
spbnick, line 220: I haven't looked at the data details, we'll just see how the tests go, but you can have minimal IDs, or IDs actually describing what the object is about. You only need them to comply with the schema. E.g. something like `_:valid1`, `_:valid2`, `_:invalid` and so on. That should make the data easier to read and understand.

spbnick (changes requested):
Nice small start! Let's continue in small steps. Perhaps merge a few tests at a time? Thank you :relaxed:

mrbazzan:
> Nice small start! Let's continue in small steps. Perhaps merge a few tests at a time? Thank you relaxed

Ready!

kcidb/test_oo.py
```diff
@@ -203,3 +204,225 @@ def test_two_object(self):
...
+            oo_object_list: A list of kcidb.oo.Object instances.
+            num_instances:  The number of instance of object contained
+                            in the oo_object_list.
+        """
+        assert isinstance(oo_object_list, list)
+        assert isinstance(num_instances, int)
+
+        self.assertEqual(
+            len(oo_object_list), num_instances,
+            msg=f"Invalid number of {oo_object_list[0]} instances"
```
spbnick, line 234: This would fail with another exception if the `oo_object_list` is empty, masking the assertion failure. I.e. you can't really find out the expected type of objects in an empty list. Additionally, strictly speaking, the type of one object doesn't guarantee that the rest of them are the same type. To make this work reliably and make sense you'd need to pass the (name of the) expected type to the function explicitly.

kcidb/test_oo.py
```diff
@@ -203,3 +204,225 @@ def test_two_object(self):
...
+            oo_dict:    A dictionary of objects indexed by two indices.
+            item:       A list of first indexing items.
+       """
+        for key in oo_dict.keys():
+            self.assertIn(key, item)
+            self.assertEqual(oo_dict[key].keys(), item[key].keys())
+            for type_objs in oo_dict[key]:
+                self.assertEqual(
+                    len(oo_dict[key][type_objs]), item[key][type_objs]
+                )
```
spbnick, line 251: Having and using this kind of function for our tests seems like a good idea.

However, it needs better documentation, explaining what exactly the last argument should be. It can also be made more elegant and more universal with recursion, allowing us to validate any number of nested dictionaries, or even arrays. We also need to verify that none of them contain anything more than what we expect.

We can shorten `Dictionary` to just `Dict` in the name, and we also need to be explicit that it's not checking actual contents, but only the counts. A nice addition could be checking the types of objects too, but we can start with just the counts.

Can you figure out how to do that?

kcidb/test_oo.py
```diff
@@ -203,3 +204,225 @@ def test_two_object(self):
...
+
+    # Conform to unittest conventions, pylint: disable=invalid-name
+
+    def query_str(self, pattern_string):
+        """Run OO client query with a parsed string."""
+        return self.client.query(
+            kcidb.orm.Pattern.parse(pattern_string)
+        )
+
+    def assertNumberOfInstances(self, oo_object_list, num_instances):
```
spbnick, line 220: This function doesn't actually care whether "instances", or anything else (like numbers, strings, or types) is contained in the lists. So we can have more generic names for the function and the arguments. This would make the function more useful and the code using it easier to understand.

spbnick (changes requested):
Nice start for the first bunch of tests! I left a few style and functionality comments inline.

mrbazzan:
> Nice start for the first bunch of tests! I left a few style and functionality comments inline.

@spbnick
I have implemented the necessary changes. Please review.

mrbazzan:
> Nice start for the first bunch of tests! I left a few style and functionality comments inline.

@spbnick
I have implemented the necessary changes. Please review.

kcidb/test_oo.py
mrbazzan: Yes. of course. Check out how I went about it.

mrbazzan: Yes. of course. Check out how I went about it.

kcidb/test_oo.py
```diff
@@ -203,3 +203,240 @@ def test_two_object(self):
...
+
+        self.assertEqual(
+            len(list_of_items), num_elements,
+            msg=f"Invalid number of {type_name} instances"
+        )
+        for i, obj in enumerate(list_of_items):
+            self.assertIsInstance(obj, kcidb.oo.CLASSES[type_name],
+                                  msg=f"Invalid type of {type_name} #{i}")
+
+    def assertDictContains(self, dict_of_items, data, type_name):
```
spbnick, line 241: I think we can get away with just one function, for both kinds of containers. E.g.:

```python
def assertContains(self, container, contents):
    """
    Check that a container has the specified contents.

    Args:
        container:  The container to check the contents of.
        contents:   The contents pattern to check against. The pattern matches
                    either a list of certain number of objects of particular
                    type, or a dictionary with string keys and values matching
                    specified patterns.

                    A list pattern is a tuple containing the type of each
                    contained instance, and the number of those instances.

                    A dictionary pattern is a dictionary containing string
                    keys for the matched dictionary to have, with values being
                    patterns for the corresponding values in the matched
                    dictionary.

    Returns:
        True if the container matched the contents pattern, False otherwise.
    """
```

If we make the patterns contain types directly, instead of their names in kcidb.oo.CLASSES, then the invocations will be more direct and obvious, and we can make the function work with more types.

You can, of course, import the types you need directly, so you can e.g. write `Checkout` instead of `kcidb.oo.Checkout`, and make the code shorter.

BTW, we also need to check that dictionaries don't contain keys we did not specify.

kcidb/test_oo.py
```diff
@@ -203,3 +203,240 @@ def test_two_object(self):
...
+    def test_traversing_revision_links(self):
+        """Check that revision's links are successfully traversed."""
+
+        revision = self.query_str(
+            '>revision["5acb9c2a7bc836e9e5172bbcd2311499c5b4e5f1", ""]#'
+        )['revision'][0]
+
+        self.assertListContains(revision.checkouts, 3, "checkout")
+        self.assertListContains(revision.builds, 6, "build")
+        self.assertListContains(revision.tests, 12, "test")
```
spbnick, line 426: If we implement the function above, then we can write:
```python
        self.assertContains(revision.checkouts, (kcidb.oo.Checkout, 3))
        self.assertContains(revision.checkouts, (kcidb.oo.Build, 6))
        self.assertContains(revision.checkouts, (kcidb.oo.Test, 12))
```

kcidb/test_oo.py
```diff
@@ -203,3 +203,240 @@ def test_two_object(self):
...
+
+        self.assertFalse(revision.checkouts_valid)
+        self.assertFalse(revision.builds_valid)
+
+        self.assertIsInstance(revision.tests_root, kcidb.oo.Node)
+
+        self.assertDictContains(revision.repo_branch_checkouts, {
+            "https://repo_valid": {"valid1": 1, "valid2": 1},
+            "https://repo_invalid": {"invalid": 1}
+        }, type_name="checkout")
```
spbnick, line 436: With the function above this would be something like this:
```python
        self.assertContains(repo.revision_branch_checkout, {
            "https://repo_valid": {
                "valid1": (kcidb.oo.Checkout, 1),
                "valid2": (kcidb.oo.Checkout, 1),
            },
            "https://repo_invalid": {
                "invalid": (kcidb.oo.Checkout, 1),
            },
        })
```

spbnick (changes requested):
This looks much nicer, thank you, Abdulwasiu! I added one more suggestion, however.

kcidb/test_oo.py
mrbazzan: > You can, of course, import the types you need directly, so you can 

Importing directly works perfectly with `Checkout` and `Build`, but gives warning with `Test` because **unittest** tries to collect the Test class but it couldn't because it has a ` __init__ ` constructor. 

I added `__test__= False` to prevent `Test` from being collected by **unittest**.

mrbazzan:
> This looks much nicer, thank you, Abdulwasiu! I added one more suggestion, however.

No problem. Suggestion implemented.
Please review when you can.

kcidb/oo/__init__.py
```diff
@@ -418,6 +418,9 @@ def tests(self):
 class Test(Object):
     """An OO-representation of a test"""
 
+    # prevent class from being collected by unittest.
+    __test__ = False
+
```
spbnick, line 423: Nice one!

kcidb/test_oo.py
```diff
@@ -203,3 +204,246 @@ def test_two_object(self):
...
+            for i, obj in enumerate(container):
+                self.assertIsInstance(
+                    obj, contents[0],
+                    msg=f"Invalid type of {contents[0]} #{i}")
+
+        elif isinstance(contents, dict):
+            for key in contents:
+                self.assertIn(key, container,
+                              msg=f"Invalid key({key}) found in container.")
+                self.assertContains(container[key], contents[key])
```
spbnick, line 256: We also need to check that the `container` doesn't contain any keys beside the ones in `contents`.

kcidb/test_oo.py
```diff
@@ -203,3 +204,246 @@ def test_two_object(self):
...
+            )
+            for i, obj in enumerate(container):
+                self.assertIsInstance(
+                    obj, contents[0],
+                    msg=f"Invalid type of {contents[0]} #{i}")
+
+        elif isinstance(contents, dict):
+            for key in contents:
+                self.assertIn(key, container,
+                              msg=f"Invalid key({key}) found in container.")
```
spbnick, line 255: Shouldn't this be complaining about the `key` not being found in the `container` instead?

spbnick (changes requested):
Just two minor comments, otherwise looks awesome! Thank you, Abdulwasiu!

mrbazzan:
> Just two minor comments, otherwise looks awesome! Thank you, Abdulwasiu!
Good to go.

kcidb/test_oo.py
```diff
@@ -203,3 +204,250 @@ def test_two_object(self):
...
+                msg=f"Invalid number of {contents[0]} instances"
+            )
+            for i, obj in enumerate(container):
+                self.assertIsInstance(
+                    obj, contents[0],
+                    msg=f"Invalid type of {contents[0]} #{i}")
+
+        elif isinstance(contents, dict):
+            key = contents.keys() ^ container.keys()
+            self.assertFalse(key, msg=f"{key} found.")
```
spbnick, line 258: This is a little too terse. Could you make a separate assertion for missing keys, and another for extra keys, each with a clear message to avoid head scratching in case of failure?

Otherwise looks great, thank you :blush:

mrbazzan:
> Just two minor comments

Good to go.

spbnick:
Looks great! Thank you, Abdulwasiu! Merging.

PR #300, opened 2022-04-02
Error in the CI

### LINKED ISSUE #299: Actions: Error in the CI
Due to a recent update in **pylint** ( >= v2.13.0 ), some linting errors popped up in the Actions checker.

```
************* Module kcidb.mq
mq/__init__.py:552:4: W0221: Number of parameters was 1 in 'Subscriber.decode_data' and is now 2 in overridden 'EmailSubscriber.decode_data' method (arguments-differ)
************* Module kcidb.oo
oo/__init__.py:349:12: E4702: Iterated dict 'repo_branch_checkouts' is being modified inside for loop body, iterate through a copy of it instead. (modified-iterating-dict)
```

Possible options to solve this issue
- Set a specific version of pylint(<2.13.0)
- We can also disable the pylint checker
- To solve the `modified-iterating-dict`, use `repo_branch_checkout.items()` instead of `for repo in repo_branch_checkout`.

or others...

mrbazzan:
Closes #299

### DIFF
--- a/kcidb/oo/__init__.py
+++ b/kcidb/oo/__init__.py
@@ -344,6 +344,7 @@ class Revision(Object, BuildContainer, TestContainer):
             if branch not in branch_checkouts:
                 branch_checkouts[branch] = []
             branch_checkouts[branch].append(checkout)
+        # pylint: disable=modified-iterating-dict
         # Sort branches by number of checkouts
         for repo in repo_branch_checkouts:
             repo_branch_checkouts[repo] = dict(sorted(
--- a/kcidb/mq/__init__.py
+++ b/kcidb/mq/__init__.py
@@ -549,6 +549,7 @@ class EmailPublisher(Publisher):
 class EmailSubscriber(Subscriber):
     """Email queue subscriber"""
 
+    # We'll be OK, pylint: disable=arguments-differ
     def decode_data(self, message_data):
         """
         Decode email from the message data.


kcidb/oo/__init__.py
```diff
@@ -345,9 +345,9 @@ def repo_branch_checkouts(self):
                 branch_checkouts[branch] = []
             branch_checkouts[branch].append(checkout)
         # Sort branches by number of checkouts
-        for repo in repo_branch_checkouts:
+        for repo, obj in repo_branch_checkouts.items():
             repo_branch_checkouts[repo] = dict(sorted(
-                repo_branch_checkouts[repo].items(),
+                obj.items(),
```
spbnick, line 350: Hmm, we don't actually prevent modifying the `repo_branch_checkouts` here. The only difference is that we're accessing its elements via `items()`, and are using an intermediate variable for the values.

Does that really avoid the pylint warning? Is that because we simply outsmarted pylint, or does `items()` actually make a copy of the dictionary keys and values for iteration, and thus avoids the "modify in a loop" problem? Could you check that please?

Thanks!

spbnick: Also, is it really unsafe to change values when iterating over keys?

mrbazzan: > Does that really avoid the pylint warning? Is that because we simply outsmarted pylint, or does `items()` actually make a copy of the dictionary keys and values for iteration, and thus avoids the "modify in a loop" problem? Could you check that please?
> 
Yes, `.items()` returns an independent copy of the container.  [link(https://stackoverflow.com/a/2315529/13712347). It returns a dictionary view object which is a read-only proxy(an interface to the object in memory), when the dictionary changes, the view also reflect these changes.

mrbazzan: > Also, is it really unsafe to change values when iterating over keys?

No
https://stackoverflow.com/a/6777632/13712347

`Pylint` probably assumes that any changes to the original container might mean modification, and hence the need to iterate over a copy, then change the original container.

spbnick: Looking at it and reading some more, I think it's just fine to leave as is, and ask pylint to not worry about this :grin:

We're not change the dictionary keys, only values, so that's OK, and the original code is simpler.

Thank you!

mrbazzan: Okay.

spbnick:
Thank you, merged! Next time let's add an explanation of why that particular dict-modifying warning could be waived safely, but this time let's unblock other PRs faster :)

mrbazzan:
> Next time let's add an explanation of why that particular dict-modifying warning could be waived safely

I added an extra comment on the commit. Perhaps it wasn't concise enough?

PR #302, opened 2022-04-20
More test cases for oo module

mrbazzan:
Part of #293

### DIFF
--- a/kcidb/test_oo.py
+++ b/kcidb/test_oo.py
@@ -2,7 +2,7 @@
 
 import kcidb
 from kcidb.unittest import local_only
-from kcidb.oo import Checkout, Build, Test
+from kcidb.oo import Checkout, Build, Test, Node
 
 
 @local_only
@@ -455,4 +455,4 @@ class KCIDBTraversingTestCase(kcidb.unittest.TestCase):
         self.assertFalse(revision.checkouts_valid)
         self.assertFalse(revision.builds_valid)
 
-        self.assertIsInstance(revision.tests_root, kcidb.oo.Node)
+        self.assertIsInstance(revision.tests_root, Node)
--- a/kcidb/test_oo.py
+++ b/kcidb/test_oo.py
@@ -217,6 +217,23 @@ class KCIDBTraversingTestCase(kcidb.unittest.TestCase):
             kcidb.orm.Pattern.parse(pattern_string)
         )
 
+    @staticmethod
+    def filter_valid(container):
+        """
+        Filter out the valid items in a container.
+        Args:
+            container:  A list whose items are object(s) that
+                        have "valid" property.
+
+        Returns:
+            The list of object(s) with valid property true.
+        """
+        assert all(hasattr(i, "valid") for i in container)
+        return list(filter(
+            lambda item: item.valid is True,
+            container
+        ))
+
     def assertContains(self, container, contents):
         """
         Check that a container has the specified contents.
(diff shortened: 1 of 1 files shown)


mrbazzan:
@spbnick Ready for review!

kcidb/test_oo.py
```diff
@@ -264,6 +284,18 @@ def assertContains(self, container, contents):
 
                 self.assertContains(container[key], contents[key])
 
+    def get_valid_item(self, pattern_str, type):
+        """
+        Get a list of object(s) that has property, valid, to be true.
+        Args:
+            pattern_str:    A string which depicts the object to be retrieved.
+            types:          A specified schema type name.
```
spbnick, line 292: Please document the return value.

kcidb/test_oo.py
```diff
@@ -456,3 +488,67 @@ def test_traversing_revision_links(self):
         self.assertFalse(revision.builds_valid)
 
         self.assertIsInstance(revision.tests_root, Node)
+        self._traversing_revision_root_test_node(revision)
```
spbnick, line 491: This doesn't have to be a part of this test case. We can have that as a separate, explicit test case. So please remove this call, and make that function a test case.

kcidb/test_oo.py
```diff
@@ -264,6 +284,18 @@ def assertContains(self, container, contents):
 
                 self.assertContains(container[key], contents[key])
 
+    def get_valid_item(self, pattern_str, type):
```
spbnick, line 287: This mostly does `query_str` and then filters out the valid items. So we should either make sure the querying is obvious from the function name, and call it something like `query_str_valid()`, or just make a static method for filtering the result of invoking `query_str()`, call it something like `filter_valid()`, and change the invocation appropriately.

kcidb/test_oo.py
```diff
@@ -253,6 +263,16 @@ def assertContains(self, container, contents):
...
+        elif isinstance(contents, list):
+            self.assertItemIsTwo(contents)
+            self.assertIsInstance(container, contents[0])
+            self.assertEqual(container.parent, contents[1])
+            self.assertContains(container.tests, (Test, 4))
+
+            # Check that each tests have the same path.
+            self.assertEqual(len({obj.path for obj in container.tests}), 1)
+            self.assertEqual(container.nodes, {})
+
```
spbnick, line 275: This is a very specific change for such a generic function, and it's not documented. Did it end up here by accident? Could you explain what you're trying to do here?

spbnick (changes requested):
Thank you, Abdulwasiu! You're moving fast, but I have a few questions/requests :)

kcidb/test_oo.py
mrbazzan: I was trying to use `assertContain` to cater to `Node` conditions. I shouldn't have done it that way.

mrbazzan:
```python
self.assertEqual(len({obj.path for obj in container.tests}), 1)
```
? 
I was checking that the `Test` returned by each sub-nodes test property has the same path.

mrbazzan:
@spbnick
Ready for review.

kcidb/test_oo.py
```diff
@@ -217,6 +217,25 @@ def query_str(self, pattern_string):
...
+    def filter_valid(container, type):
+        """
+        Filter out the valid items in a container.
+        Args:
+            container:  A dictionary whose values are object(s) that
+                        have "valid" property.
+            type:       A key in the container.
+
+        Returns:
+            The list of object(s) with valid property which is true.
```
spbnick, line 230: This would be easier to read, I think:
```suggestion
            The list of object(s) with valid property true.
```

kcidb/test_oo.py
```diff
@@ -217,6 +217,25 @@ def query_str(self, pattern_string):
...
         )
 
+    @staticmethod
+    def filter_valid(container, type):
+        """
+        Filter out the valid items in a container.
+        Args:
+            container:  A dictionary whose values are object(s) that
+                        have "valid" property.
+            type:       A key in the container.
```
spbnick, line 227: Indexing a dictionary is a fundamental operation, and doesn't have anything to do with purely filtering a list. So, please, remove this argument, and explicitly retrieve the corresponding list from the container at the invocation point. Hiding such a trivial operation inside a function doesn't save us code, and makes it harder to understand. Let's make this function as simple as possible, while still being useful.

kcidb/test_oo.py
```diff
@@ -217,6 +217,25 @@ def query_str(self, pattern_string):
...
+
+        Returns:
+            The list of object(s) with valid property which is true.
+        """
+        assert type in container
+        assert (hasattr(i, "valid") for i in container.values())
```
spbnick, line 233: I think this will never fail. Did you forget to call `all()` here?

spbnick (changes requested):
Thanks a lot, @mrbazzan, this looks pretty good :relaxed: I only have two comments.

spbnick:
> Although check_auxillary_files is failing because of 504 error from this [url](https://wiki.ubuntu.com/FirmwareTestSuite/)

Yeah, probably just some proxy failure at Ubuntu. I retried it and it passes now.

kcidb/test_oo.py
mrbazzan: Okay.

spbnick:
Looks good to me! Thank you, Abdulwasiu!

Merging.

Do we have any more OO tests planned in that issue?

mrbazzan:
> Do we have any more OO tests planned in that issue?
No, we don't.

PR #305, opened 2022-05-04
Add PostgreSQL driver

spbnick:

PR #308, opened 2022-05-06
postgresql: Fix errors in docs

mrbazzan:
Concerns #305

### DIFF
--- a/kcidb/db/postgresql/schema.py
+++ b/kcidb/db/postgresql/schema.py
@@ -19,7 +19,7 @@ class Column:
     @staticmethod
     def pack(value):
         """
-        Pack the JSON representation of the column value into the SQLite
+        Pack the JSON representation of the column value into the PostgreSQL
         representation.
         """
         return value
@@ -27,7 +27,7 @@ class Column:
     @staticmethod
     def unpack(value):
         """
-        Unpack the SQLite representation of the column value into the JSON
+        Unpack the PostgreSQL representation of the column value into the JSON
         representation.
         """
         return value
@@ -37,8 +37,8 @@ class Column:
         Initialize the column schema.
 
         Args:
-            type:           The name of the SQLite type (affinity) to use for
-                            this column.
+            type:           The name of the PostgreSQL type (affinity) to use
+                            for this column.
             constraint:     The column's constraint.
                             A member of the Constraint enum, or None,
                             meaning no constraint.
@@ -143,7 +143,7 @@ class TimestampColumn(Column):
     @staticmethod
     def unpack(value):
         """
-        Unpack the SQLite representation of the column value into the JSON
+        Unpack the PostgreSQL representation of the column value into the JSON
         representation.
         """
         return value.isoformat()


mrbazzan:
CC: @spbnick 
Some trivial things I also noticed:
1) These [lines](https://github.com/kernelci/kcidb/blob/a6fdb0b47138fdaed0b6847c06fb149199c1966e/kcidb/db/postgresql/schema.py#L239-L241) seem incoherent. It should be simplified.

2) Why did you use this [format ](https://github.com/kernelci/kcidb/blob/a6fdb0b47138fdaed0b6847c06fb149199c1966e/kcidb/db/postgresql/__init__.py#L83-L90) and not something like the one below?
```python
        with self.conn, self.conn.cursor() as cursor:
             try:
                cursor.execute("SELECT get_version()")
                version = cursor.fetchone()[0]
                return int(version / 1000), int(version % 1000)
             # It's auto-generated, pylint: disable=no-member
             except psycopg2.errors.UndefinedFunction:
                return 0, 0
```

spbnick:
Thanks a lot for the fixes, @mrbazzan! Merging :relaxed:

spbnick (approved):

spbnick:
> CC: @spbnick Some trivial things I also noticed:
> 
>     1. These [lines](https://github.com/kernelci/kcidb/blob/a6fdb0b47138fdaed0b6847c06fb149199c1966e/kcidb/db/postgresql/schema.py#L239-L241) seem incoherent. It should be simplified.

Do you mean it's hard to understand? If yes, then how about this:

```python
        """
        Format the "INSERT/UPDATE" command template for loading a row into a
        database, observing deduplication logic.

        Args:
            name:       The name of the target table of the command.
            prio_db:    If true, format the UPDATE part of the command so that
                        the values already in the database take priority over
                        the loaded ones, and vice versa otherwise.
        Returns:
            The formatted "INSERT/UPDATE" command template, expecting
            parameters packed by the pack() method.
        """
```

>     2. Why did you use this [format ](https://github.com/kernelci/kcidb/blob/a6fdb0b47138fdaed0b6847c06fb149199c1966e/kcidb/db/postgresql/__init__.py#L83-L90) and not something like the one below?
> 
> 
> ```python
>         with self.conn, self.conn.cursor() as cursor:
>              try:
>                 cursor.execute("SELECT get_version()")
>                 version = cursor.fetchone()[0]
>                 return int(version / 1000), int(version % 1000)
>              # It's auto-generated, pylint: disable=no-member
>              except psycopg2.errors.UndefinedFunction:
>                 return 0, 0
> ```

Hmm, no reason in particular. I would say they do mostly the same thing.
Do you think one is better than the other?

mrbazzan:
> > CC: @spbnick Some trivial things I also noticed:
> > ```
> > 1. These [lines](https://github.com/kernelci/kcidb/blob/a6fdb0b47138fdaed0b6847c06fb149199c1966e/kcidb/db/postgresql/schema.py#L239-L241) seem incoherent. It should be simplified.
> > ```
> 
> Do you mean it's hard to understand? If yes, then how about this:
> 
> ```python
>         """
>         Format the "INSERT/UPDATE" command template for loading a row into a
>         database, observing deduplication logic.
> 
>         Args:
>             name:       The name of the target table of the command.
>             prio_db:    If true, format the UPDATE part of the command so that
>                         the values already in the database take priority over
>                         the loaded ones, and vice versa otherwise.
>         Returns:
>             The formatted "INSERT/UPDATE" command template, expecting
>             parameters packed by the pack() method.
>         """
> ```
> 

Perfect! I'll create a PR to that effect.

> > ```
> > 2. Why did you use this [format ](https://github.com/kernelci/kcidb/blob/a6fdb0b47138fdaed0b6847c06fb149199c1966e/kcidb/db/postgresql/__init__.py#L83-L90) and not something like the one below?
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```python
> >         with self.conn, self.conn.cursor() as cursor:
> >              try:
> >                 cursor.execute("SELECT get_version()")
> >                 version = cursor.fetchone()[0]
> >                 return int(version / 1000), int(version % 1000)
> >              # It's auto-generated, pylint: disable=no-member
> >              except psycopg2.errors.UndefinedFunction:
> >                 return 0, 0
> > ```
> 
> Hmm, no reason in particular. I would say they do mostly the same thing. Do you think one is better than the other?

No, all the other instance of doing something like this in `postgresql` db used something like the format I specified. So, I just wanted to know if there was a reason for changing it.

spbnick:
> No, all the other instance of doing something like this in postgresql db used something like the format I specified. So, I just wanted to know if there was a reason for changing it.

Ah, I see. Yeah makes sense to have the same approach everywhere. Perhaps on the next refactoring we'll clean that up. Meanwhile it should be good enough. Thanks!

PR #309, opened 2022-05-09
postgresql: Update docs

mrbazzan:


mrbazzan:
CC @spbnick

spbnick:
Thanks for submitting this, Abdulwasiu! Merged :relaxed: 

It's no big deal, but it's better to be a bit more specific in commit messages. We're not really updating anything here, since the code and the comment didn't go out of sync, but we could say that we're improving readability, or clarifying the docs. So something like `postgresql: Clarify schema Table.format_insert() docs` could've been better.

PR #322, opened 2022-10-06
broonie: Add more repos

spbnick:


spbnick:
@broonie, I'm adding your personal repos to KCIDB monitoring. Do those look right?

spbnick:
Alright, will keep all of them, then, and remove the CI one, if it becomes a problem. Thanks!

PR #354, opened 2023-02-17
submit: Clarify that we are submitting a JSON object

broonie:
The documentation for submit() says that it needs to be passed JSON, making a bit ambiguous if this means a Python JSON object or the rendered JSON.  Explicitly say that the former is required.


spbnick (approved):

spbnick:
Thanks, Mark :relaxed:

PR #356, opened 2023-02-20
Switch to new draft7 format checker of jsonschema

spbnick:
Switch to using the new format checker object of the jsonschema package. This fixes the deprecation warning.


spbnick:
Merged as part of #357

PR #369, opened 2023-03-07
Doc: Removing release mentions

### LINKED ISSUE #353: Update documentation to instruct using latest kcidb
Releases are going away in favor of continuous integration and deployment, so update the documentation to direct users to downloading the latest `main` branch.

This is done, when all files in the `doc` directory stop directing readers to download the latest or specific release, and instead tell them to download/install the latest `main` branch.

octonawish-akcodes:
These changes will allow users to download the main branch instead of latest version.
@spbnick Need reviews.

Fixes #353

### DIFF
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -15,10 +15,6 @@ Where `<SOURCE>` is the location of the package source, e.g. a git repo:
 
     pip3 install --user git+https://github.com/kernelci/kcidb.git
 
-a (release) tag in a git repo:
-
-    pip3 install --user git+https://github.com/kernelci/kcidb.git@v9
-
 or a directory path:
 
     pip3 install --user .
--- a/doc/submitter_guide.md
+++ b/doc/submitter_guide.md
@@ -60,12 +60,15 @@ dashboard](https://kcidb.kernelci.org/).
 2\. Install KCIDB
 -----------------
 
-Pick a KCIDB [release][releases]. Latest would normally be best.
 
-Let's say you picked v9. Run this:
+KCIDB employs continuous integration and delivery, and aims to keep
+the code working at all times.
+
+Please install the latest version from GitHub:
+
 
 ```bash
-pip3 install --user 'git+https://git@github.com/kernelci/kcidb.git@v9'
+pip3 install --user 'git+https://git@github.com/kernelci/kcidb.git'
 ```
 
 Then make sure your PATH includes the `~/.local/bin` directory, e.g. with:
@@ -95,11 +98,10 @@ exit status.
 3\. Generate some report data
 -----------------------------
 
-KCIDB accepts data in JSON, described by a versioned schema. The
-`kcidb-schema` tool will output the current schema version. However, all tools
-will accept data complying with older schema versions, as a rule. Any
-exceptions to that will be mentioned in KCIDB release notes. Pipe your data
-into `kcidb-validate` tool to check if it will be accepted.
+`kcidb-schema` tool will output the current schema version.
+
+However, all tools will accept data complying with older schema versions. Pipe
+your data into `kcidb-validate` tool to check if it will be accepted.
 
 Here's a minimal report, containing no data:
 
@@ -406,7 +408,6 @@ let you, kernel developers, and KCIDB developers see how it works, what
 changes/additions might be needed to both your data and KCIDB, and improve our
 reporting faster!
 
-[releases]: https://github.com/kernelci/kcidb/releases
 [datetime_format]: https://tools.ietf.org/html/rfc3339#section-5.6
 [tests]: https://github.com/kernelci/kcidb/blob/master/tests.yaml
 [dashboard]: https://kcidb.kernelci.org/


doc/submitter_guide.md
```diff
@@ -60,19 +60,18 @@ dashboard](https://kcidb.kernelci.org/).
...
-pip3 install --user 'git+https://git@github.com/kernelci/kcidb.git@v9'
+pip3 install --user 'git+https://git@github.com/kernelci/kcidb.git@main'
 ```
 
 Then make sure your PATH includes the `~/.local/bin` directory, e.g. with:
 
     export PATH="$PATH":~/.local/bin
 
-See [Installation](../installation) for alternatives, and if you know your
+See [Installation](installation.md) for alternatives, and if you know your
```
spbnick, line 74: Unfortunately, this won't work when deployed to https://kernelci.org/docs/kcidb/submitter_guide/ Please revert.

doc/submitter_guide.md
```diff
@@ -60,19 +60,18 @@ dashboard](https://kcidb.kernelci.org/).
 2\. Install KCIDB
 -----------------
 
-Pick a KCIDB [release][releases]. Latest would normally be best.
```
spbnick, line 63: Please remove the `releases` link reference at the bottom of the file as well.

doc/installation.md
```diff
@@ -13,8 +13,7 @@ To install the package for the current user, run this command:
 
 Where `<SOURCE>` is the location of the package source, e.g. a git repo:
 
-    pip3 install --user git+https://github.com/kernelci/kcidb.git
-
+    pip3 install --user git+https://git@github.com/kernelci/kcidb.git@main
```
spbnick, line 16: This doesn't really need changing, as it would pick the `main` branch by default. It's actually better to keep it this way in case the main branch name changes.

doc/installation.md
```diff
@@ -13,8 +13,7 @@ To install the package for the current user, run this command:
...
 a (release) tag in a git repo:
 
     pip3 install --user git+https://github.com/kernelci/kcidb.git@v9
```
spbnick, line 19: This, however, needs to be removed, as it mentions releases.

doc/submitter_guide.md
```diff
@@ -60,19 +60,18 @@ dashboard](https://kcidb.kernelci.org/).
...
 
-Let's say you picked v9. Run this:
+This will install the latest version of the package from the main branch of the kcidb repository.
```
spbnick, line 64: We need to be clear that picking up latest `main` is OK here, and explain the development philosophy. Something like this would do, but feel free to expand/modify, if you have better ideas:

> KCIDB employs continuous integration and delivery, and aims to keep the code working at all times. Please install the latest version from GitHub:

spbnick (changes requested):
A good take, thank you, Abhishek.

I have a few requests/comments, below and inline.

We need to remove all mentions of releases.

Please also change this:
> However, all tools will accept data complying with older schema versions, as a rule. Any
exceptions to that will be mentioned in KCIDB release notes.

to just this:
> However, all tools will accept data complying with older schema versions.

We'll figure out later how to notify about any future incompatibilities, if any (and that shouldn't really happen anyway).

doc/submitter_guide.md
octonawish-akcodes: Okay done

doc/submitter_guide.md
octonawish-akcodes: done

doc/installation.md
octonawish-akcodes: done

doc/submitter_guide.md
octonawish-akcodes: Okay

doc/submitter_guide.md
```diff
@@ -60,12 +60,11 @@ dashboard](https://kcidb.kernelci.org/).
 2\. Install KCIDB
 -----------------
 
-Pick a KCIDB [release][releases]. Latest would normally be best.
 
-Let's say you picked v9. Run this:
+KCIDB employs continuous integration and delivery, and aims to keep the code working at all times. Please install the latest version from GitHub:
```
spbnick, line 64: Please wrap the text at the 78 character boundary. This makes the raw documentation easier to read (and Markdown is intended to be read raw as well).

doc/submitter_guide.md
```diff
@@ -96,9 +95,7 @@ exit status.
 -----------------------------
 
 KCIDB accepts data in JSON, described by a versioned schema. The
-`kcidb-schema` tool will output the current schema version. However, all tools
-will accept data complying with older schema versions, as a rule. Any
-exceptions to that will be mentioned in KCIDB release notes. Pipe your data
+`kcidb-schema` tool will output the current schema version. However, all tools will accept data complying with older schema versions. Pipe your data
```
spbnick, line 98: Please also wrap this paragraph at 78 character boundary.

spbnick (changes requested):
Almost perfect, thank you, Abhishek!

I left a couple comments about text wrapping, inline.

Also please keep the commits organized in the way they should go into the repo, when merged. That is commits that each make sense by themselves, and don't contain more than one logical change. Please also use [imperative mood](https://en.wikipedia.org/wiki/Imperative_mood) in commit messages, i.e. as if giving a command for something to change. Use "tag prefixes" when appropriate, see commit history of the particular area you change to find which ones you should use. E.g. for your change something like "doc:" would be appropriate.

In the case of this PR, your two commits should be merged into one (I don't need to know the history of how you developed this PR, only how it will go into the repo in the end), and the commit subject should be something like "doc: Remove mentions of releases".

doc/submitter_guide.md
octonawish-akcodes: I have combined all commits into one single commit with the changes you suggested above sir.

doc/submitter_guide.md
octonawish-akcodes: okay

doc/submitter_guide.md
```diff
@@ -60,12 +60,13 @@ dashboard](https://kcidb.kernelci.org/).
...
-Pick a KCIDB [release][releases]. Latest would normally be best.
 
-Let's say you picked v9. Run this:
+KCIDB employs continuous integration and delivery, and aims to keep the code working at all times.
+
+Please install the latest version from GitHub:
 
 ```bash
-pip3 install --user 'git+https://git@github.com/kernelci/kcidb.git@v9'
+pip3 install --user 'git+https://git@github.com/kernelci/kcidb.git@main'
```
spbnick, line 69: Could you please remove `@main` from the end? It will work without it, would look simpler, and wouldn't tie us to the particular name.

doc/submitter_guide.md
```diff
@@ -95,11 +96,10 @@ exit status.
 3\. Generate some report data
 -----------------------------
 
-KCIDB accepts data in JSON, described by a versioned schema. The
-`kcidb-schema` tool will output the current schema version. However, all tools
-will accept data complying with older schema versions, as a rule. Any
-exceptions to that will be mentioned in KCIDB release notes. Pipe your data
-into `kcidb-validate` tool to check if it will be accepted.
+`kcidb-schema` tool will output the current schema version. However, all tools will accept data complying with older schema versions.
```
spbnick, line 99: Please wrap this paragraph to 78 characters.

spbnick (changes requested):
Looks good, just two minor requests left. Thank you!

doc/submitter_guide.md
octonawish-akcodes: Done

doc/submitter_guide.md
octonawish-akcodes: Done, but I am a bit confused will it work?

spbnick: > Done, but I am a bit confused will it work?

It should, IIRC, the remote repository provides the default branch name to check out, and that's what pip uses. Regardless, you could always try it yourself :grin:

spbnick (changes requested):
Thank you! The code looks good to me now, but please have the commits in the PR the same way as they would go into the repository, by the time of review. Do not leave amendments on top. I've described commit requirements in one of the previous reviews. In your case it should be just one commit. Thanks :relaxed:

spbnick:
Thank you, @octonawish-akcodes, LGTM! Merging.

One nitpick (merging regardless): the commit says "improve formatting", but you're not actually improving formatting with your change, it stays the same. However, thanks for remembering to keep the commit message updated with your changes!

Issue #277, opened 2022-01-06
Add pycodestyle run to CI


spbnick:
Add a run of pycodestyle on kcidb code to GitHub "test" workflow.

mrbazzan:
Can I pick this up?

spbnick:
Sure! Would be great, Abdulwasiu :) Note that the docs will need updating, though, to comply with pydocstyle requirements.

octonawish-akcodes:
Can you assign this to me sir, I am very close to fix this :)

spbnick:
Sure thing :grin:

PR #370, opened 2023-03-07
pydocstyle run added in CI

octonawish-akcodes:
These changes will fix issue #277 
Needed reviews @spbnick

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,6 +60,8 @@ jobs:
         run: "flake8 kcidb *.py"
       - name: Check python sources with pylint
         run: "pylint kcidb *.py"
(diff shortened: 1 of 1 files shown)


spbnick:
Thanks a lot, Abhishek. However, I made a mistake in the original issue, and it should be `pydocstyle`, not `pycodestyle` :see_no_evil: Fixed that now.

Resolving the issue should also include modifying the code/docstrings to comply with the tool's requirements where possible/makes sense, so CI keeps passing. So it's OK if you find that too much work for this contribution period (I know you're trying to make as many contributions as possible).

spbnick:
Closing this for now to keep the list of PRs cleaner.

PR #383, opened 2023-03-09
doc: Re-created administrator guide workflow diagram using mermaid js

### LINKED ISSUE #240: Consider redrawing architecture diagram using Mermaid
@gctucker [said earlier](https://github.com/kernelci/kcidb/pull/238#discussion_r753338087):
> ASCII art is cool, but for this static website there's also Mermaid which is easier on the eyes.
>
> See this example:
 > https://github.com/kernelci/kernelci-project/blame/main/kernelci.org/content/en/docs/architecture.md#L53
> and how it's rendered:
> https://kernelci.org/docs/architecture/

Consider using that to make the architecture diagram nicer in `doc/administrator_guide.md`. One drawback, is that it becomes unreadable in the raw Markdown file, kinda defeating its original idea, but perhaps we could live with that for the sake of a prettier website.

octonawish-akcodes:
fix #240 

Recreated the administrator guide workflow using mermaid js.

@spbnick Need reviews

### DIFF
--- a/doc/administrator_guide.md
+++ b/doc/administrator_guide.md
@@ -5,52 +5,47 @@ draft: false
 wieght: 40
 description: "Deploying, maintaining, and upgrading a KCIDB service installation"
 ---
-Kcidb infrastructure is mostly based on Google Cloud services at the moment:
-
-    === Hosts ===  ======================= Google Cloud Project ========================
-
-    ~ Webserver ~                                                    ~~~~ BigQuery ~~~~~
-    kcidb-grafana <-------------------------------------------------  .. kcidb_XX ..
-                                                                     :   checkouts  :
-    ~~ Client ~~~                                                    :   builds     :
-    kcidb-query <--------------------------------------------------- :   tests      :
-                                                                      ''''''''''''''
-                    ~~ Pub/Sub ~~       ~~~~ Cloud Functions ~~~~            ^
-                    kcidb_trigger ----------.                                |
-                                            v                                |
-    kcidb-submit -> kcidb_new --------> kcidb_load_queue --------------------'
-                                            |
-                          .-----------------'
-                          v                                          ~~~~ Firestore ~~~~
-                    kcidb_updated ----> kcidb_spool_notifications -> notifications
-                                                                           |
-                                                   .-----------------------'
-                                                   |
-                                                   v                 ~ Secret Manager ~~
-                                        kcidb_send_notification <--- kcidb_smtp_password
-                                                   |
-                                                   |                 ~~~~~~ GMail ~~~~~~
-                                                   `---------------> bot@kernelci.org
-
-BigQuery stores the report dataset and serves it to Grafana dashboards hosted
-on kcidb.kernelci.org, as well as to any clients invoking `kcidb-query` or
-using the kcidb library to query the database.
-
-Whenever a client submits reports, either via `kcidb-submit` or the kcidb
-library, they go to a Pub/Sub message queue topic named `kcidb_new`, then to
-the `kcidb_load_queue` "Cloud Function", which loads the data to the BigQuery
-dataset, and then pushes the list of updated objects to `kcidb_updated` topic.
-The `kcidb_load_queue` function is triggered periodically via messages to
-`kcidb_trigger` topic, pushed there by the Cloud Scheduler service.
-
-That topic is watched by `kcidb_spool_notifications` function, which picks up
-the data, generates report notifications, and stores them in a Firestore
-collection named `notifications`.
-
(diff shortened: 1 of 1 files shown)


doc/administrator_guide.md
```diff
@@ -5,32 +5,24 @@ draft: false
...
-                                                   .-----------------------'
-                                                   |
-                                                   v                 ~ Secret Manager ~~
-                                        kcidb_send_notification <--- kcidb_smtp_password
-                                                   |
-                                                   |                 ~~~~~~ GMail ~~~~~~
-                                                   `---------------> bot@kernelci.org
+```mermaid
+graph TD
+  Webserver["Webserver"] --> BigQuery["BigQuery"]
```
spbnick, line 10: We don't need separate labels here, since they're the same as the node name, I suppose?

doc/administrator_guide.md
```diff
@@ -5,32 +5,24 @@ draft: false
...
+  Client["~~ Client ~~"] --> BigQuery
```
spbnick, line 11: The `~~` around the `Client` were just part of ASCII art :joy: We can drop the label here as well.

doc/administrator_guide.md
```diff
@@ -5,32 +5,24 @@ draft: false
...
+  kcidb-query -->|Queries| BigQuery
+  BigQuery -->|Reports| kcidb-grafana["kcidb-grafana"]
```
spbnick, line 13: Same here, no need for a label

doc/administrator_guide.md
```diff
@@ -5,32 +5,24 @@ draft: false
...
+  
+  kcidb-submit --> kcidb_new["kcidb_new Pub/Sub topic"]
```
spbnick, line 15: Is there a way to have different boxes for different things? Like have Pub/Sub topics one color, Cloud Functions another, and so on? Having to spell out the object type each time would be cumbersome and make reading the diagram harder. What expressive options do we have with Mermaid in general?

doc/administrator_guide.md
```diff
@@ -5,32 +5,24 @@ draft: false
...
+  kcidb_new --> kcidb_load_queue["kcidb_load_queue Cloud Function"]
+  kcidb_load_queue -->|Loads data| BigQuery
+  BigQuery --> kcidb_updated["kcidb_updated Pub/Sub topic"]
+  kcidb_updated --> kcidb_spool_notifications["kcidb_spool_notifications Cloud Function"]
+  kcidb_spool_notifications -->|Generates notifications| Firestore["Firestore"]
```
spbnick, line 20: I would say the edge here should have something like "Stores generated notifications", instead of "Generates notifications".

doc/administrator_guide.md
```diff
@@ -5,32 +5,24 @@ draft: false
...
+  Firestore --> kcidb_send_notification["kcidb_send_notification Cloud Function"]
+  kcidb_send_notification -->|Sends notifications| GMail["GMail"]
+  SecretManager["Secret Manager"] --> kcidb_send_notification
+  kcidb_smtp_password["kcidb_smtp_password secret"] --> SecretManager
```
spbnick, line 24: In the original diagram `kcidb_smtp_password` is simply inside the `Secret Manager` (the former is a resource, and the latter is the service hosting it). Is there a way for us to have containers in a Mermaid diagram?

spbnick (changes requested):
As I feared, so far there doesn't seem to be much expressive power here, but perhaps there are some options, or we could get creative with whatever we have? I left some comments/questions inline. Thanks!

doc/administrator_guide.md
spbnick: We could also have "Hosts" or "Serves" edges from `Secret Manager` to the secrets like `kcidb_smtp_password`, of course. That could get complicated fast, though, if we want to describe all containers this way.

doc/administrator_guide.md
octonawish-akcodes: ok

doc/administrator_guide.md
octonawish-akcodes: ok

doc/administrator_guide.md
octonawish-akcodes: ok

doc/administrator_guide.md
octonawish-akcodes: Yes there are ways make containers but I am not sure about the color coding, have to look.

doc/administrator_guide.md
octonawish-akcodes: Okay

PR #389, opened 2023-03-13
tests: Ignore SSL errors when validating

spbnick:
Ignore SSL errors (such as invalid or expired certificate) when validating the test catalog, since we're only verifying that the resource is there, and because that happens sometimes due to neglect, and not because the test's website went away. Otherwise our CI breaks.


spbnick:
Thanks, Tales, merging!

PR #392, opened 2023-03-14
switch from mocking to fixtures

NikkyXO:
switching from using mocks to pytests.fixtures

### DIFF
--- a/kcidb/db/__init__.py
+++ b/kcidb/db/__init__.py
@@ -74,7 +74,6 @@ class Client(kcidb.orm.Source):
                                   driver
         """
         assert isinstance(database, str)
-        self.database = database
         self.driver = misc.instantiate_spec(DRIVER_TYPES, database)
         assert all(io_schema <= io.SCHEMA
                    for io_schema in self.driver.get_schemas().values()), \
--- a/kcidb/db/__init__.py
+++ b/kcidb/db/__init__.py
@@ -74,6 +74,7 @@ class Client(kcidb.orm.Source):
                                   driver
         """
         assert isinstance(database, str)
+        self.database = database
         self.driver = misc.instantiate_spec(DRIVER_TYPES, database)
         assert all(io_schema <= io.SCHEMA
                    for io_schema in self.driver.get_schemas().values()), \
--- a/kcidb/test_db.py
+++ b/kcidb/test_db.py
@@ -1,7 +1,6 @@
 """kcdib.db module tests"""
 
 import re
-import textwrap
 import datetime
 import json
 from itertools import permutations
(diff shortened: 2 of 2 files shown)


NikkyXO:
@spbnick  i will like you to review the code work i did in kcidb.test_db.test_init_main before i continue with the rest of the changes

kcidb/test_db.py
```diff
@@ -19,25 +20,19 @@ def test_schemas_main():
...
-        client.init = Mock()
-        client.is_initialized = Mock(return_value=False)
-        with patch("kcidb.db.Client", return_value=client) as \
-                Client:
-            status = function()
-        Client.assert_called_once_with("bigquery:project.dataset")
-        client.init.assert_called_once()
+        if not empty_database:
+            pytest.skip("Empty database not configured")
+        status = kcidb.db.init_main()
```
spbnick, line 29: If the pytest configuration doesn't have the fixture available, then the test won't be executed. We don't need to check for it here.

There is a bit of a problem before we can run those tests with these fixtures, actually.

The entry points for the command-line tools we're testing here (i.e. `kcidb.db.init_main`, which is an entry point for the `kcidb-db-init` tool) expect a database specification string (run `kcidb-db-init --database-help` to get a description of those), while the `empty_database` and `clean_database` fixtures are objects of type `kcidb.db.Client` representing an already-established connection to a database. One cannot be used in place of another as is. However, the `kcidb.db.Client` objects are created with a database specification string as the input. You could modify that class to store the specification string, say in `self.database`, and then extract it from the `empty_database` or `clean_database` fixture in the test, to pass to the tool's entry point. I don't think you would need to pass **any** driver_source to `assert_executes` here, as we're not going to be doing any "mocking".

If you'd like we can have a call tomorrow, and I'll walk you through what's going on. Reach out on Slack and tell me which times you have available, if you'd like to do that.

NikkyXO: implemented changes @spbnick

kcidb/test_db.py
```diff
@@ -19,194 +19,77 @@ def test_schemas_main():
...
-                Client:
-            status = function()
-        Client.assert_called_once_with("bigquery:project.dataset")
-        client.dump_iter.assert_called_once()
-        return status
-    """)
-    assert_executes("", *argv, driver_source=driver_source,
-                    stdout_re=re.escape(json.dumps(empty) + "\n" +
-                                        json.dumps(empty) + "\n"))
+    assert_executes("", *argv, stdout_re=re.escape(json.dumps(empty) + "\n"))
```
spbnick, line 50: Only this will succeed for an empty database.

Next test would be loading something into the database e.g. using `empty_database.load(...)`, then dumping, and verifying that the data matches. There are plenty of test I/O data samples in other tests, you could borrow for this one. Look for `load()` calls.

kcidb/test_db.py
```diff
@@ -19,194 +19,77 @@ def test_schemas_main():
...
+    assert_executes("", *argv, stdout_re=re.escape(json.dumps(empty) + "\n" +
+                                                   json.dumps(empty) + "\n"))
```
spbnick, line 52: This would only work before, because we mocked the client returning the data. We don't need this exact test anymore. The one I describe above would be better.

kcidb/test_db.py
```diff
@@ -19,194 +19,77 @@ def test_schemas_main():
...
-    argv = ["kcidb.db.query_main", "-d", "bigquery:project.dataset"]
-    assert_executes("", *argv, driver_source=driver_source)
+
+    argv = ["kcidb.db.query_main", "-d", empty_database.database]
+    empty_database.is_initialized()
+    assert_executes("", *argv)
 
     argv = [
-        "kcidb.db.query_main", "-d", "bigquery:project.dataset",
+        "kcidb.db.query_main", "-d", empty_database.database,
```
spbnick, line 63: Here too, you would need to load the data you're querying. Look to other tests calling `query()` for samples.

kcidb/test_db.py
```diff
@@ -1059,15 +942,15 @@ def test_query(empty_database):
...
-    ]
+               {
+                   **unambigious_result,
+                   "incidents": [incident_x, incident_y],
+               },
+               {
+                   **unambigious_result,
+                   "incidents": [incident_y, incident_x],
+               },
+           ]
```
spbnick, line 953: Please don't post unrelated whitespace changes. Here, and above. If you need to change this for something, e.g. to make CI work, make a separate commit doing it, and provide explanation in the commit message. Thanks!

spbnick (changes requested):
I left some comments inline about the obvious things I could see.

Regarding the problems you're seeing, let's go one at a time: post the command you're executing (e.g. `pytest`), the output you're getting (e.g. pytest output with the error message/backtrace), and what you're expecting to see instead. Please also make sure the code in the PR is up to date with what you're running, so that I could reproduce your problem and investigate.

Thank you!

NikkyXO:
> I left some comments inline about the obvious things I could see.
> 
> 
> 
> Regarding the problems you're seeing, let's go one at a time: post the command you're executing (e.g. `pytest`), the output you're getting (e.g. pytest output with the error message/backtrace), and what you're expecting to see instead. Please also make sure the code in the PR is up to date with what you're running, so that I could reproduce your problem and investigate.
> 
> 
> 
> Thank you!

I made some changes, the tests are passing now

kcidb/.test_db.py.swp
spbnick, line 1: Please remove this file from the commit.

kcidb/db/.schematic.py.swp
spbnick, line 1: Please remove this file from the commit.

how_to.md
spbnick, line 1: Did you want to add this file to the commit? If yes, could you explain what you're trying to do?

kcidb/test_db.py
```diff
@@ -1,7 +1,5 @@
 """kcdib.db module tests"""
-
```
spbnick, line 2: Please do not include unnecessary whitespace changes into unrelated commits. If e.g. CI requires it, please make a separate commit for it.

kcidb/test_db.py
```diff
@@ -19,194 +17,201 @@ def test_schemas_main():
...
     """Check kcidb-db-query works"""
-    driver_source = textwrap.dedent("""
-        from unittest.mock import patch
-        with patch("kcidb.db.Client"):
-            return function()
-    """)
-    argv = ["kcidb.db.query_main", "-d", "bigquery:project.dataset"]
-    assert_executes("", *argv, driver_source=driver_source)
+
+    empty_database.is_initialized()
```
spbnick, line 147: This line doesn't seem to have any effect. Please remove, unless you want to take it somewhere further and just forgot to do that.

kcidb/test_db.py
```diff
@@ -19,194 +17,201 @@ def test_schemas_main():
...
+            dict(id="_:2", origin="_", issue_id="_:2", issue_version=1,
+                 test_id="_:1"),
+            dict(id="_:3", origin="_", issue_id="_:3", issue_version=1,
+                 build_id="_:2", test_id="_:2"),
+            dict(id="_:4", origin="_", issue_id="_:4", issue_version=1,
+                 test_id="_:1"),
+            dict(id="_:5", origin="_", issue_id="_:4", issue_version=1,
+                 test_id="_:2"),
+        ],
+    ))
```
spbnick, line 191: Please also reduce the data to just one object per type. Then you wouldn't have a problem comparing it to the query result due to order changing. You can also just create a single fixture with the data, shared by several tests and use it here.

kcidb/test_db.py
```diff
@@ -19,194 +17,201 @@ def test_schemas_main():
...
+
     assert_executes(
-        json.dumps(empty), *argv,
-        driver_source=driver_source,
-        stdout_re=re.escape(
-            json.dumps(empty) + "\n" +
-            json.dumps(empty) + "\n"
-        )
-    )
+        json.dumps(empty), *argv)
```
spbnick, line 196: OK, this obviously works, but you haven't actually verified that you get the data you just loaded, so it's missing the point of the test. You will have to specify the IDs of the objects you load above to the query command in your `argv` list. Also, the query command doesn't take (ignores) the input you're specifying in the first argument, please remove.

kcidb/test_db.py
```diff
@@ -19,194 +17,201 @@ def test_schemas_main():
...
-        client = Mock()
-        client.get_schema = Mock(return_value=((1, 0), V4_1))
-        with patch("kcidb.db.Client", return_value=client):
-            return function()
-    """)
-    argv = ["kcidb.db.load_main", "-d", "bigquery:project.dataset"]
 
-    assert_executes("", *argv, driver_source=driver_source)
-    assert_executes('{', *argv, driver_source=driver_source,
+    empty_database.is_initialized()
```
spbnick, line 202: Same here, this line has no effect.

kcidb/test_db.py
```diff
@@ -19,194 +17,201 @@ def test_schemas_main():
...
-            status = function()
-        Client.assert_called_once_with("bigquery:project.dataset")
-        assert client.load.call_count == 2
-        client.load.assert_has_calls([call({repr(empty)}),
-                                      call({repr(empty)})])
-        return status
-    """)
-    assert_executes(json.dumps(empty) + json.dumps(empty), *argv,
-                    driver_source=driver_source)
+    assert_executes(json.dumps(empty) + json.dumps(empty), *argv)
```
spbnick, line 214: Would be good to also load some actual data (e.g. the same data used by other tests), and verify it gets loaded, but we can get it one step at a time, and do that in the next PR.

.swp
spbnick, line 1: Please remove this file from the commit. These files are created by `vim` and are normally removed on exit, but I probably forgot to exit it before our shared session terminated, so they were left around. You can remove them from your disk as well, not only from the commit.

spbnick (changes requested):
Hey, this is pretty good, @NikkyXO! I left a bunch of comments inline, but we're getting there. Let's try to go easy and slow and not do too many things at once.

Apart from those comments, you need to boost your git commit manipulation game to be able to post neat PRs. See if you could follow https://learngitbranching.js.org/ for a bit to learn more about that, and then we can have a video call and a shared session to show you some things.

Oh, and please clean up that commit message. We don't need it to say the same thing three times :grin: Thanks!

NikkyXO:
Alright. I will go through the git commit tutorial, and make recommended changes on my PT.  Thank you.  
I will reach out to you on slack later on, for  a date for the meet call

how_to.md
NikkyXO: No, i'm removing it

NikkyXO:
@spbnick i made corrections

kcidb/test_db.py
```diff
@@ -19,194 +18,170 @@ def test_schemas_main():
...
-            status = function()
-        Client.assert_called_once_with("bigquery:project.dataset")
-        client.dump_iter.assert_called_once()
-        return status
-    """)
-    assert_executes("", *argv, driver_source=driver_source,
-                    stdout_re=re.escape(json.dumps(empty) + "\n"))
+    argv = ["kcidb.db.dump_main", "-d", empty_database.database,
+            "--indent=0"]
+    empty_database.is_initialized()
```
spbnick, line 46: This line has no effect, please remove.

spbnick (changes requested):
Looking pretty good, we're close! Left you a few comments/requests inline. Thank you, @NikkyXO!

kcidb/db/__init__.py
```diff
@@ -74,7 +74,10 @@ def __init__(self, database):
                                   driver
         """
         assert isinstance(database, str)
+
+        self.database = database
         self.driver = misc.instantiate_spec(DRIVER_TYPES, database)
+        self.database = database
```
spbnick, line 80: I think doing this once is enough, and this line has no effect. Please remove. Please also remove the extra blank line you're adding above.

spbnick (changes requested):
The test code seems to be good now, thank you, @NikkyXO.

However, both commits contain changes to tests. Please make the first commit only add the storage of the database string, and the second commit only the test changes. Please remove all unrelated formatting and whitespace changes.

Thank you!

PR #394, opened 2023-03-16
cloud: schema feature added

### LINKED ISSUE #372: cloud: Support specifying the database schema to deploy
At the moment the [cloud tool](https://github.com/kernelci/kcidb/blob/main/cloud) doesn't specify the version of the schema to initialize the databases to, when deploying, so the databases get initialized to the latest version. Add support for specifying the schema version using a command-line option with the `cloud deploy` command. E.g. `--schema VERSION`, similar to the `kcidb-db-init` tool used for actual initialization.

This will make it easier to make deployments with older schema versions, for testing upgrades, and reproducing issues with existing (older-schema) deployments.

octonawish-akcodes:
Support for schema version for initializing the db with versions specified.

This PR will closes #372 

@spbnick need reviews

Am I on the right track? I would appreciate some guidance. Also how can I test it.

### DIFF
--- a/kcidb/db/__init__.py
+++ b/kcidb/db/__init__.py
@@ -590,10 +590,16 @@ def query_main():
     description = \
         "kcidb-db-query - Query objects from Kernel CI report database"
     parser = QueryArgumentParser(description=description)
+    parser.add_argument(
+        '--schema', type=str, default='latest',
+        help='I/O schema version to use (in the format io:<major>.<minor>) '
+             'or "latest" to use the latest schema version'
+    )
     args = parser.parse_args()
-    client = Client(args.database)
+    client = Client(args.database, io_version=args.schema)
     if not client.is_initialized():
         raise Exception(f"Database {args.database!r} is not initialized")
+    io_schema = client.get_schema()[1]
     query_iter = client.query_iter(
         ids=dict(checkouts=args.checkout_ids,
                  builds=args.build_ids,
@@ -609,17 +615,24 @@ def query_main():
     )
 
 
+
 def load_main():
     """Execute the kcidb-db-load command-line tool"""
     sys.excepthook = kcidb.misc.log_and_print_excepthook
     description = \
         'kcidb-db-load - Load reports into Kernel CI report database'
     parser = ArgumentParser(description=description)
+    parser.add_argument('--schema', help='I/O schema version in the format "io:<major>.<minor>"')
     args = parser.parse_args()
     client = Client(args.database)
     if not client.is_initialized():
         raise Exception(f"Database {args.database!r} is not initialized")
-    io_schema = client.get_schema()[1]
+    io_schema = io.latest_schema() if args.schema == 'latest' else None
+    if args.schema and args.schema.startswith('io:'):
+        major, minor = map(int, args.schema[3:].split('.'))
+        io_schema = io.find(major, minor)
+    if io_schema is None:
+        io_schema = client.get_schema()[1]
     for data in kcidb.misc.json_load_stream_fd(sys.stdin.fileno()):
         data = io_schema.upgrade(io_schema.validate(data), copy=False)
         client.load(data)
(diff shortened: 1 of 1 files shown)


cloud
```diff
@@ -1260,6 +1265,7 @@ function cloud_functions_env() {
                           database \
                           clean_test_databases \
                           empty_test_databases \
+                          schema \
```
spbnick, line 1268: I don't think we need to pass the schema version to the environment. We could discover it from the databases themselves, if needed.

cloud
```diff
@@ -1675,6 +1681,7 @@ function execute_command() {
                           updated_publish \
                           submitters \
                           argv \
+                          schema \
```
spbnick, line 1684: Let's keep the variable name consistent throughout the code, if we can. I think `schema_version` works well for that.

spbnick (changes requested):
This is a good start, now you need to write the code to accept (and validate) the version number from the command line, pass it around where it's needed, and apply it to other database types as well.

Regarding testing, I've created a Google Cloud project for you to test your changes. It's called `kcidb-abhishek` and you should've received an invitation to join it via email. Next you need to install the Google Cloud SDK, create a service account with administrator permissions, and create and download a key for it. Then you could export its location via the `GOOGLE_APPLICATION_CREDENTIALS` environment variable, and after that you would be able to run the cloud tool to make a test deployment. Try with an unmodified `cloud` tool first. The command to deploy to your project would look something like `./cloud deploy kcidb-abhishek "" 1 -v --smtp-mocked --test --heavy-asserts`. Use the [Administrator Guide](https://kernelci.org/docs/kcidb/administrator_guide/) as a reference.

cloud
octonawish-akcodes: Okay

cloud
octonawish-akcodes: okay

octonawish-akcodes:
@spbnick Need reviews

spbnick (changes requested):
This is starting to look good. Thank you, @octonawish-akcodes. I left a few comments inline.

However, I just realized that we can't just pass the same schema version to each database type (BigQuery/PostgreSQL/sqlite), as they could be totally different, because the schema development for different databases has different needs and moves at different speeds.

However, we can specify the I/O schema version, which the databases should support instead, and that has to be the same. The `kcidb.db.Client` class [supports](https://github.com/kernelci/kcidb/blob/cfcfe8f2d7940f6b0ae7ebe15d6a19ffaca17ea2/kcidb/db/__init__.py#L96-L101) that, but none of the database-managing command-line tools do. So we would need a commit before this one adding support for that.

We could make this work if we e.g. allow the `--schema` options for these tools to accept a string in the shape of `io:<major>.<minor>` for the I/O schemas. While we're at it, we could also allow it to accept the string `latest`, meaning the latest database schema version, and that would help you get rid of those "specified/unspecified schema version" branches you're adding to database initialization functions in `cloud`.

octonawish-akcodes:
Tried to implement I/O versions to the db command-line tools, am I on the right track? Need reviews.

cloud
```diff
@@ -2218,6 +2241,7 @@ function execute() {
             --heavy-asserts) heavy_asserts="true"; shift;;
             --mute-updates) updated_publish="false"; shift;;
             --test) test="true"; shift;;
+            --schema) schema_version="$2"; shift 2;;
```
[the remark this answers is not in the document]
octonawish-akcodes: Okay

cloud
```diff
@@ -2164,7 +2186,7 @@ function execute() {
         getopt_longopts+=",smtp-mocked,test"
         if [[ $command == @(deploy|withdraw) ]]; then
             getopt_shortopts+="vs:"
-            getopt_longopts+=",verbose,sections:,submitter:"
+            getopt_longopts+=",verbose,sections:,submitter:,schema:"
```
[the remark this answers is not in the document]
octonawish-akcodes: But like in the withdrawing stage doesnt it need the schema values?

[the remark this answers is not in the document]
spbnick: The "withdraw" and all other stages would be able to retrieve the schema version from the databases themselves, so only "deploy" needs it to know which schema to initialize the databases to.

spbnick (changes requested):
This is a good start.

Note that only two tools use the `--schema` option for now: `kcidb-db-init` and `kcidb-db-upgrade`, no other tool needs it (including `kcidb-db-query`).

Instead of copying the option-parsing code around, create another "argparse type" function, similar to `kcidb.misc.version()` and `schema_version()` inside `kcidb.misc.argparse_schema_add_args()`. Then both the tools would be able to use it. The function should produce a value which would be accepted by both `kcidb.db.Client.init()` and `kcidb.db.Client.upgrade()`.

Actually, this might need an extraction of a method looking up a database schema corresponding to an I/O schema, from both `init()` and `upgrade()`, in `kcidb.db.Client`, and then use of that method in both of them. This would make it possible to use that new method to check and report an error whenever the resulting database version is older than the current one, in `kcidb-db-upgrade` (instead of just failing an assertion in `init()` and `upgrade()`).

PR #395, opened 2023-03-16
Add pre-commit hook script

### LINKED ISSUE #401: Feature : pre-commit scripts
It would be nice to have [pre-commit ](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) hooks for testing the changes before commit to the branch. 
We currently use pylint, flake8, and yamllint (possibly thinking of adding some more).  The way to do this will be to create an executable in the top directory and symlink into `.git/hooks/pre-commit` .
It could be great if we could just call that script from the GitHub test workflow, instead of having to copy the commands between them.

shivam-Purohit:
Added feature: pre-commit 
closes #401
checks against flake8 and black for now. 
@spbnick Is this the correct method? Which one should we add and (possibly remove ) in this?
Also, we will require to make changes in documentation and specify the user to use the pre-commit install cmd after installation.
PS: pre-commit black changed the formatting. If it is not required we can remove that.

### DIFF
--- a/.github-workflow-script
+++ b/.github-workflow-script
@@ -1,0 +1,30 @@
+#!/usr/bin/env python3
+"""Extract steps containing special env variable KCIDB_HOOKS from Github workflow"""
+import yaml
+import argparse
+
+# Parse command-line arguments
+parser = argparse.ArgumentParser()
+parser.add_argument("workflow_file",
+                    help="Path to the GitHub Actions workflow file")
+
+args = parser.parse_args()
+
+# Load the GitHub Actions workflow file as YAML
+with open(args.workflow_file, "r") as file:
+    data = yaml.safe_load(file)
+
+print("#!/bin/bash")
+# Loop over all jobs
+for job_name, job in data["jobs"].items():
+    # Loop over all steps in the job
+    for step in job["steps"]:
+        # Check if the step has a KCIDB_HOOKS environment variable
+        if "env" in step and step["env"].get("KCIDB_HOOKS") == "pre-commit":
+            # Print the name of the step
+            print(f"\n# {step['name']}")
+            # Extract the command(s) from the step
+            command = step.get("run", [])
+            # Print the command(s)
+            for line in command if isinstance(command, list) else [command]:
+                print(line)
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -63,7 +63,7 @@ jobs:
       - name: Test with pytest
         run: KCIDB_IO_HEAVY_ASSERTS=1 KCIDB_HEAVY_ASSERTS=1 pytest --tb=native
 
-  check_auxilary_files:
+  check_auxiliary_files:
 
     runs-on: ubuntu-latest
     strategy:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,10 +58,16 @@ jobs:
           pip3 install --upgrade '.[dev]'
       - name: Check python sources with flake8
         run: "flake8 kcidb *.py"
+        env:
+          KCIDB_HOOKS: "pre-commit"
       - name: Check python sources with pylint
         run: "pylint kcidb *.py"
+        env:
+          KCIDB_HOOKS: "pre-commit"
       - name: Test with pytest
         run: KCIDB_IO_HEAVY_ASSERTS=1 KCIDB_HEAVY_ASSERTS=1 pytest --tb=native
+        env:
+          KCIDB_HOOKS: "pre-commit"
 
   check_auxiliary_files:
 
@@ -83,6 +89,8 @@ jobs:
           pip3 install --upgrade '.[dev]'
       - name: Check YAML files with yamllint
         run: find -name '*.yaml' -or -name '*.yml' | xargs yamllint
+        env:
+          KCIDB_HOOKS: "pre-commit"
       - name: Check test catalog is ordered alphabetically
         run: |
           test_list=$(
@@ -100,6 +108,8 @@ jobs:
                   echo "$test_list_diff" >&2
                   false
               }
+        env:
+          KCIDB_HOOKS: "pre-commit"
       - name: Validate test catalog
         run: kcidb-tests-validate --urls < tests.yaml
 
--- a/.pre-commit
+++ b/.pre-commit
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+set -e -u -o pipefail
+
+if git rev-parse --verify HEAD >/dev/null 2>&1
+then
+	against=HEAD
+else
+	# Initial commit: diff against an empty tree object
+	against=$(git hash-object -t tree /dev/null)
+fi
+
+# If you want to allow non-ASCII filenames set this variable to true.
+allownonascii=$(git config --bool hooks.allownonascii || echo "false")
+
+# Redirect output to stderr.
+exec 1>&2
+
+# Cross platform projects tend to avoid non-ASCII filenames; prevent
+# them from being added to the repository. We exploit the fact that the
+# printable range starts at the space character and ends with tilde.
+if [ "$allownonascii" != "true" ] &&
+	# Note that the use of brackets around a tr range is ok here, (it's
+	# even required, for portability to Solaris 10's /usr/bin/tr), since
+	# the square bracket bytes happen to fall in the designated range.
+	test $(git diff --cached --name-only --diff-filter=A -z $against |
+	  LC_ALL=C tr -d '[ -~]\0' | wc -c) != 0
+then
+	cat <<\EOF
+Error: Attempt to add a non-ASCII file name.
+
+This can cause problems if you want to work with people on other platforms.
+
+To be portable it is advisable to rename the file.
+
+If you know what you are doing you can disable this check using:
+
+  git config hooks.allownonascii true
+EOF
+	exit 1
+fi
+
+# If there are whitespace errors, print the offending file names and fail.
+git diff-index --check --cached $against --
+
+# Execute steps extracted from GitHub Actions
+. <(./.github-workflow-script .github/workflows/test.yml)
--- a/doc/developer_guide.md
+++ b/doc/developer_guide.md
@@ -22,6 +22,30 @@ Then make sure your PATH includes the `~/.local/bin` directory, e.g. with:
 
     export PATH="$PATH":~/.local/bin
 
+### Pre-commit hooks
+
+Git hooks are scripts that can be triggered automatically by Git when
+certain events occur, such as committing changes to a repository.
+They are typically used to enforce coding standards, perform automated
+tests, or perform other checks on code changes before they are committed.
+
+There are two main components in the setup 1) `.pre-commit` script and
+2) `.github-workflow-script`. The `.github-workflow-script` extracts the
+steps directly from the GitHub workflow, while a `.pre-commit `script will
+run the extracted commands.
+
+To leverage the setup, symlink the `.pre-commit` script into the git
+pre-commit hook (`.git/hooks/pre-commit`). We can do that using a
+simple command in the terminal. Make sure you are in the root
+directory of the project.
+
+    ln -s ../../.pre-commit .git/hooks/pre-commit
+
+Now, every time you run git commit, the pre-commit hook script will be
+executed before the commit is created. If any of the pre-commit checks
+fail, the commit will be aborted and you will see an error message
+explaining the issue.
+
 Guidelines
 ----------
 


shivam-Purohit:
First, I will work on making the pre-commit script including pytest, pylint and flake8. We will move from there then. I have not done much scripting work before I will refer to your code for help.

spbnick:
Thank you for your continued work on this, Shivam.

We shouldn't have anything under the `.git` directory stored in the repo, because that's where git stores its own files, and we better not interfere with that using git itself. Instead let's put our hook in the `.pre-commit` file in the root of the repository tree (beside `.gitignore`, `setup.py`, etc.), and add a note to the `doc/developer_guide.md` to link it into the `.git/hooks/pre-commit` file with a relative symlink. E.g. with the `ln -s ../../.pre-commit .git/hooks/pre-commit` command. Note that you're using an absolute symlink, which contains the full path to your repository, and that is unlikely to work anywhere else except your machine.

Please remove all code changes unrelated to the change you're working on, from this PR. Please also make sure the commits are in [shape](https://kernelci.org/docs/kcidb/developer_guide/#commits) before the review (that is there is only the required number of logically-atomic commits and nothing else in the PR). Try going through the [interactive git branching tutorial](https://learngitbranching.js.org/) if you need to brush up on history rewriting and commit amending.

shivam-Purohit:
> However, it would be nice to add a paragraph about this to the `docs/developer_guide.md` at the same time, as that increases usefulness of the change quite a bit. You'd need to recommend using the pre-commit hook script from the repo for development, and put the symlinking command I wrote in a previous comment here. Could you do that?

yes I am preparing to do that .



> Sorry, I'm not sure what you're asking here. Could you elaborate?
I am not sure how can we grab the steps from the workflow yaml and run them in the pre-commit. I am searching for a way

spbnick:
> yes I am preparing to do that .

Thanks a lot!

> I am not sure how can we grab the steps from the workflow yaml and run them in the pre-commit. I am searching for a way

Well, I guess first write a little Python program using the `yaml` module to read the GitHub workflow file, get down to the objects describing the steps and the shell commands to run, and outputting them as a shell script.

Then we can work on integrating it into the pre-commit hook.

We can do that in a follow-up PR, though.

.pre-commit
```diff
@@ -0,0 +1,49 @@
...
+	exit 1
+fi
+
+# If there are whitespace errors, print the offending file names and fail.
+git diff-index --check --cached $against --
+flake8 kcidb *.py
+pylint kcidb *.py
+find -name '*.yaml' -or -name '*.yml' | xargs yamllint
+kcidb-tests-validate < tests.yaml
+KCIDB_IO_HEAVY_ASSERTS=1 KCIDB_HEAVY_ASSERTS=1 pytest --tb=native --disable-warnings
```
spbnick, line 49: The last line is missing a newline on the end. Some editors do that (remove trailing newlines), but we don't normally have files like this in the repo. Could you change your editor settings to not do this, and add the newline? Otherwise those will keep sneaking into PRs and create unnecessary churn.

shivam-Purohit:
> Could you rewrite it in Python, so that we don't have to require `yq`? Just using the `yaml` module should be enough.

I first wrote it in python only 😂 and tried converting it to bash script file. I will use my original file

shivam-Purohit:
@spbnick I think this should be good ?

spbnick:
Please reorganize the commits to match [commit guidelines](https://kernelci.org/docs/kcidb/developer_guide/#commits), that is one commit per independent change, before review.

shivam-Purohit:
> Please reorganize the commits to match [commit guidelines](https://kernelci.org/docs/kcidb/developer_guide/#commits), that is one commit per independent change, before review.

should i squash them into 3 commits
3 commits 
one defining .pre-commit script
one .github-workflow-script
one is the documentation change

shivam-Purohit:
I could not get the exact changes in the commit. If you want I could delete this pr and open another with only 3 commits that I mentioned above.

.github-workflow-script
```diff
@@ -0,0 +1,51 @@
...
+                        for name in ['Check python sources with flake8',
+                                     'Check python sources with pylint',
+                                     'Test with pytest'])
+                      ]
+
+check_aux_files_commands = [step['run'] for step in workflow['jobs']
+                                          ['check_auxiliary_files']['steps']
+                            if 'run' in step and any(name in step['name']
+                              for name in ['Check YAML files with yamllint'])
+                            ]
```
spbnick, line 33: This is a bit too tightly coupled to the workflow YAML we have, and so is brittle (will break easily with future changes).

How about we turn this around: we mark the steps that should be used in pre-commit hooks in the workflow YAML somehow, and then pick them out in this script. So you could just go over each job and step in order, pick the ones that are marked, and output their commands.

We could mark the steps by assigning a special environment variable for each of them. E.g.:
```yaml
      - name: Check python sources with flake8
        run: "flake8 kcidb *.py"
        env:
          KCIDB_HOOKS: "pre-commit"
```

spbnick: Another thing: I don't think we need to store the steps in arrays first. Just output them as you find them. It would also be nice to have the name of each step output as a comment in front of its command. So the example above would be output like this:
```bash
# Check python sources with flake8
flake8 kcidb *.py
```

spbnick (changes requested):
Thank you, Shivam, this starts to look very nice. Also, you write good docs!

I left you a few requests/comments inline.

Regarding commits, this really doesn't need to be multiple commits, as there was no pre-commit hook before, and so a change adding it, along with the requisite script and documentation could be considered atomic, and done in one commit.

However, please separate the fix for "auxiliary" into its own commit preceding your changes.

doc/developer_guide.md
```diff
@@ -22,6 +22,30 @@ Then make sure your PATH includes the `~/.local/bin` directory, e.g. with:
...
+commands in your terminal. Make sure you are in the root directory of the
+project.
+
+    chmod +x .pre-commit
+    chmod +x .github-workflow-script
+
+Now, every time you run git commit, the pre-commit hook script will be
+executed before the commit is created. If any of the pre-commit checks
+fail, the commit will be aborted and you will see an error message
+explaining the issue.
```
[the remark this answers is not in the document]
shivam-Purohit: okay I taught we were not working on this right now like linking with the git hook. I did not do that in my code. Will that be a problem? I will replace the make exec commands with this one.
another thing is that the git does not show any pre-commit file on its own. It has the pre-commit.sample 
would we need to like specify the user to do so or our commit will save and git will remember the change?

.github-workflow-script
shivam-Purohit: > So you could just go over each job and step in order, pick the ones that are marked, and output their commands.

That seems interesting. I will try it tonight.

shivam-Purohit: > ```shell
> # Check python sources with flake8
> ```

```
#Check python sources with flake8:
flake8 kcidb *.py

#Check python sources with pylint:
pylint kcidb *.py

#Test with pytest:
KCIDB_IO_HEAVY_ASSERTS=1 KCIDB_HEAVY_ASSERTS=1 pytest --tb=native

#Check YAML files with yamllint:
find -name '*.yaml' -or -name '*.yml' | xargs yamllint
```
that # names will be ignored right?
this is the terminal result

.github/workflows/test.yml
```diff
@@ -63,7 +63,7 @@ jobs:
       - name: Test with pytest
         run: KCIDB_IO_HEAVY_ASSERTS=1 KCIDB_HEAVY_ASSERTS=1 pytest --tb=native
 
-  check_auxilary_files:
+  check_auxiliary_files:
```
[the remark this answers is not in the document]
shivam-Purohit: Once the changes are done. I will club the commits or try that other way of doing

doc/developer_guide.md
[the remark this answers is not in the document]
spbnick: We don't need any code for that. Just add something like this to the docs:
```markdown
To activate the pre-commit hook symlink the `.pre-commit` file into `.git/hooks/pre-commit`.
E.g. with this command:

    ln -s ../../.pre-commit .git/hooks/pre-commit
```
We cannot commit this symlink anyway, because it goes into Git's data for the repository. So developers must create the link themselves. It should also be left up to developers to decide if they want to run anything on commit.

Yes, git doesn't have any `pre-commit` script by default, on purpose. It's up to the user to install what they need. The `pre-commit.sample` is exactly what it says - an example of such a hook with some commonly-useful checks (which we took into our `.pre-commit` script, BTW).

.github-workflow-script
spbnick: The hash character `#` starts a comment in shell (same as in Python). Please put a space after it, to make the comments easier to read, and please don't add the colon `:` on the end. They're fine without them. Everyone expects a comment on a line by itself to describe the following line(s), and the colon just creates unnecessary visual noise.

spbnick: Otherwise looks good!

shivam-Purohit:
```
# Check python sources with flake8
# Check python sources with pylint
# Test with pytest
================================================================= test session starts =================================================================
platform linux -- Python 3.10.6, pytest-7.2.2, pluggy-1.0.0
rootdir: /home/shivam/Desktop/pre-commit/kcidb
collected 72 items                                                                                                                                    

test_kcidb.py .....^C
```

is this good?  Can you check the PR I used the printed commands in the output of github-workflow-script and checked if the output starts with # and printed that out in the pre-commit file And executed the command below it

.pre-commit
```diff
@@ -0,0 +1,57 @@
...
+
+  git config hooks.allownonascii true
+EOF
+	exit 1
+fi
+
+# If there are whitespace errors, print the offending file names and fail.
+git diff-index --check --cached $against --
+
+#!/bin/bash
```
spbnick, line 46: You don't need this at all. Moreover, [such line](https://en.wikipedia.org/wiki/Shebang_(Unix)) is only effective as the very first line of an executable.

Just imagine that `. <(COMMAND)` *literally* *pastes* the output of the `COMMAND` in its place.

However, it would be nice if the output of `.github-workflow-script` included that as the first line, in case anyone wants to save it and use it as an independent script.

.github-workflow-script
```diff
@@ -0,0 +1,36 @@
...
+# Extract steps with an `env` tag defined in the `check_auxiliary_files` job
+for step in data["jobs"]["check_auxiliary_files"]["steps"]:
+    if "env" in step and "KCIDB_HOOKS" in step["env"]:
+        print(f"\n# {step['name']}")
+        if "run" in step:
+            command = step["run"]
+            # If the command is a list, join the items into a string
+            if isinstance(command, list):
+                command = "\n".join(command)
+            print(command)
```
spbnick, line 36: You don't need to do these separately, and it actually hurts the flexibility by picking only these two specific jobs. Just have two nested loops: the outer going over all the jobs, and the inner going over all the steps in the job, and rely fully on the `KCIDB_HOOKS` env variable to tell you what to execute.

.github-workflow-script
```diff
@@ -0,0 +1,36 @@
...
+                    help="Path to the GitHub Actions workflow file")
+
+args = parser.parse_args()
+
+# Load the GitHub Actions workflow file as YAML
+with open(args.workflow_file, "r") as file:
+    data = yaml.safe_load(file)
+
+for step in data["jobs"]["check_python"]["steps"]:
+    if "env" in step and "KCIDB_HOOKS" in step["env"]:
```
spbnick, line 18: Could you also check that the value of `KCIDB_HOOKS` is `pre-commit`?

spbnick (changes requested):
Getting there! Please see the inline comments.

.pre-commit
```diff
@@ -0,0 +1,57 @@
...
+# Call the GitHub Actions YAML workflow file processing script and capture the output
+output=$(./.github-workflow-script .github/workflows/test.yml)
+
+while IFS= read -r line; do
+  if [[ $line == \#* ]]; then
+    echo "$line"
+  else
+    eval "$line"
+  fi
+done <<< "$output"
```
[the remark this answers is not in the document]
spbnick: No, the `.pre-commit` script shouldn't output anything unless something broke, and the commands we execute there will take care of that (well, except pytest, but OK, it takes a while).

The `.github-workflow-script` should output the code to be executed in the `.pre-commit` script. We only add the comments to that generated code, so it's easier to read and debug, not to output them during the execution.

Sorry, if that wasn't clear before.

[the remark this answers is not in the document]
shivam-Purohit: I think I misunderstood the statement. I made the changes you request. Do let me know if it is okay.

spbnick (changes requested):
Hey, this looks pretty good! Thank you, @shivam-Purohit. I have only a bunch of cosmetic/grammar comments inline. And please make it into two commits: one fixing "auxiliary", and another adding the scripts and the documentation.

Thank you :heart:

doc/developer_guide.md
```diff
@@ -22,6 +22,30 @@ Then make sure your PATH includes the `~/.local/bin` directory, e.g. with:
...
+### Pre-commit hooks
+
+GitHub pre-commit hooks are scripts that can be triggered automatically by
+Git when certain events occur, such as committing changes to a repository.
+They are typically used to enforce coding standards, perform automated
+tests, or perform other checks on code changes before they are committed.
+
+There are two main components in the setup 1) .pre-commit script and
+2) .github-workflow-script. The github-workflow-script extracts the steps
+directly from the github-workflow,while the .pre-commit script will run
```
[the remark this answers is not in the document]
shivam-Purohit: .pre-commit here should also have the backtick right?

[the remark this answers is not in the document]
spbnick: Yeah, because we're talking about a literal file, not a concept.

shivam-Purohit:
@spbnick I squashed the commits.  The CI test arent running for some reason

doc/developer_guide.md
```diff
@@ -22,6 +22,30 @@ Then make sure your PATH includes the `~/.local/bin` directory, e.g. with:
 
     export PATH="$PATH":~/.local/bin
 
+### Pre-commit hooks
+
+GitHub pre-commit hooks are scripts that can be triggered automatically by
```
spbnick, line 27: I suppose you wanted to talk about git hooks in general, not only "pre-commit" hook, here. If so, please remove the word "pre-commit".

Also, it should be "Git", not "GitHub".

.github/workflows/test.yml
```diff
@@ -83,6 +91,8 @@ jobs:
           pip3 install --upgrade '.[dev]'
       - name: Check YAML files with yamllint
         run: find -name '*.yaml' -or -name '*.yml' | xargs yamllint
+        env:
+          KCIDB_HOOKS: "pre-commit"
       - name: Check test catalog is ordered alphabetically
         run: |
```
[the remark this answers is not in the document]
shivam-Purohit: I dont get it. Add it to the step `Check test catalog is ordered alphabetically` ?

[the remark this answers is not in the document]
spbnick: Yes, please.

spbnick:
Looks great! Thank you, @shivam-Purohit. Merging!

PR #379, opened 2023-03-08
Move argparse code into separate argparse modules 

shivam-Purohit:
Added module argparse to separate the functions and classes using the command-line argument parsing. 
closes #367

PR #396, opened 2023-03-17
Break down kcidb.orm module

### LINKED ISSUE #386: Break down kcidb.orm module
The [`kcidb.orm`](https://github.com/kernelci/kcidb/blob/main/kcidb/orm/__init__.py) module is now 1623 lines long, which is a bit much. 

We can extract parts of it into separate modules to make it more readable and maintainable.

Extract the `SCHEMA` variable and its dependencies and assertions into a new module called `kcidb.orm.data`.

Extract the `Pattern` class and its dependencies into a new module called `kcidb.orm.query`.

Update all references as necessary.

octonawish-akcodes:
Extracted `Schema` and `Pattern` class and their dependencies and made them as a separate module from kcidb.orm module.
This PR closes #386 
@spbnick Need reviews.

### DIFF
--- a/kcidb/db/abstract.py
+++ b/kcidb/db/abstract.py
@@ -191,14 +191,15 @@ class Driver(ABC):
         The database must be initialized.
 
         Args:
-            pattern_set:    A set of patterns ("kcidb.orm.Pattern"
+            pattern_set:    A set of patterns ("kcidb.orm.query.Pattern"
                             instances) matching objects to fetch.
         Returns:
             A dictionary of object type names and lists containing retrieved
             objects of the corresponding type.
         """
         assert isinstance(pattern_set, set)
-        assert all(isinstance(r, orm.Pattern) for r in pattern_set)
+        assert all(isinstance(r, orm.query.Pattern)
+                   for r in pattern_set)
         assert self.is_initialized()
 
     @abstractmethod
--- a/kcidb/orm/__init__.py
+++ b/kcidb/orm/__init__.py
@@ -3,1286 +3,14 @@ Kernel CI report object-relational mapping (ORM) - report data organized into
 objects, but without the object-oriented interface.
 """
 
-import re
-import textwrap
 import logging
-import argparse
 from abc import ABC, abstractmethod
-import jsonschema
-import kcidb.io as io
 import kcidb.misc
 from kcidb.misc import LIGHT_ASSERTS
-
-# We'll get to it, pylint: disable=too-many-lines
-
+from kcidb.orm import query, data
 
 # Module's logger
 LOGGER = logging.getLogger(__name__)
-
-
-class Relation:
-    """A parent/child relation between object types"""
-
-    def __init__(self, parent, child, ref_fields):
-        """
-        Initialize a relation.
-
-        Args:
-            parent:     The parent type.
-            child:      The child type.
-            ref_fields: A tuple containing the list of names of child fields,
-                        containing values of parent's globally-identifying
-                        fields (in the same order), and linking the two.
-        """
-        assert isinstance(parent, Type)
-        assert isinstance(child, Type)
-        assert isinstance(ref_fields, tuple)
-        assert all(isinstance(ref_field, str) for ref_field in ref_fields)
-        assert len(ref_fields) == len(parent.id_fields)
-        self.parent = parent
-        self.child = child
-        self.ref_fields = ref_fields
-
-
-class Type:
-    """An object type"""
-
-    # It's OK, pylint: disable=too-many-instance-attributes,too-many-arguments
-
-    def __init__(self, name, json_schema, id_fields):
-        """
-        Initialize an object type.
-
-        Args:
-            name:                   The type name.
-            json_schema:            The JSON schema describing the raw object
-                                    data.
-            id_fields:              A tuple containing the names of object
-                                    fields identifying it globally.
-        """
-        assert isinstance(name, str)
-        assert isinstance(json_schema, dict)
-        assert isinstance(id_fields, tuple) and \
-               all(isinstance(f, str) for f in id_fields)
-
-        # The name of this type
-        self.name = name
-        # The JSON schema for this type's raw data
-        self.json_schema = json_schema
-        # List of ID field names
-        self.id_fields = id_fields
-        # A list of all relations
-        self.relations = []
-        # A map of parent type names and their relations
-        self.parents = {}
-        # A map of child type names and their relations
-        self.children = {}
-
-    def add_relation(self, relation):
-        """
-        Add a relation of this object to another.
-
-        Args:
-            relation:   The relation to add.
-                        Must have this type as either the parent or the child.
-                        Must not have been added before.
-        """
-        assert isinstance(relation, Relation)
-        assert self is relation.parent or self is relation.child
-        assert relation not in self.relations
-        self.relations.append(relation)
-        if self is relation.parent:
-            self.children[relation.child.name] = relation
-        if self is relation.child:
-            self.parents[relation.parent.name] = relation
-
-    def validate(self, data):
-        """
-        Validate a type's data against its JSON schema.
-
-        Args:
-            data:   The data to validate.
-
-        Returns:
-            The validated (but unmodified) data.
-
-        Raises:
-            `jsonschema.exceptions.ValidationError`, if the data did not
-            adhere to this type's JSON schema.
-        """
-        try:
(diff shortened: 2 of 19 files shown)


kcidb/orm/data.py
```diff
@@ -0,0 +1,302 @@
+from kcidb.orm.__init__ import Type, Relation,_DEFS, _CHECKOUT, _BUILD, _TEST, _INCIDENT, _ISSUE, _ISSUE_CULPRIT, _TEST_ENVIRONMENT
```
spbnick, line 1: This would create a circular dependency, and shouldn't happen.

We need to bring the definitions of these dependencies here as well (as I pointed out in the issue).

Additionally, please do not import from `__init__` directly. This creates a very tight "coupling" and makes the code more fragile.

kcidb/orm/query.py
```diff
@@ -0,0 +1,1461 @@
+from kcidb.orm.__init__ import re, Type, LIGHT_ASSERTS, textwrap
```
spbnick, line 1: Same here: do not import anything from `__init__` explicitly. Also, definitely do not import common definitions available elsewhere (such as `LIGHT_ASSERTS`, or `textwrap`) via a third-party module.

spbnick (changes requested):
Nice start, Abhishek! Left you a few comments/requests inline. Thanks!

spbnick:
Don't hesitate to ask any questions you might have about this, here, or on Slack!

kcidb/orm/query.py
```diff
@@ -0,0 +1,1461 @@
...
+from kcidb.orm.data import Schema, SCHEMA
+import kcidb.io as io
+import kcidb
```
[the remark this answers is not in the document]
octonawish-akcodes: got it.

kcidb/orm/query.py
octonawish-akcodes: Okay sir.

kcidb/orm/data.py
octonawish-akcodes: Okay, I will do that.

octonawish-akcodes:
After making the suggested changes, Below is the output of the test cases. All are passing as they were passing before creating the separate modules for `Schema` and ` Pattern`.

image

octonawish-akcodes:
In updated the PR and removed the reformatting, I'll make a separate PR for that if necessary. Now this PR is purely the break down changes to be reviewed @spbnick

spbnick (changes requested):
Very nice! We're close to the finish line here!

Please deal with the CI errors, and from now on please make sure CI passes before review, as you're now more experienced with the project (please keep ignoring the deployment failures for now).

Please also don't forget to keep rebasing your changes on latest `main` as you work, and before reviews.

Finally, please explain a little what you're doing in the commit message body (take the issue description as a reference).

octonawish-akcodes:
Test cases passed, fixes the cyclic imports, and updated the references where necessary. I also added a description of what I am doing and what I have done have a look, sir. This PR is intended to work as we wanted it to, need reviews @spbnick

kcidb/db/abstract.py
```diff
@@ -1,7 +1,7 @@
 """Kernel CI reporting database - abstract database definitions"""
 
 from abc import ABC, abstractmethod
-import kcidb.orm as orm
+from kcidb.orm.query import Pattern
```
spbnick, line 4: Could you instead write:
```python
import kcidb.orm.query
```
And refer to `Pattern` as `kcidb.orm.query.Pattern`? We only have one reference here (so there's not much typing/reading overhead) and it would be good to be clear where it's coming from exactly.

kcidb/orm/data.py
```diff
@@ -0,0 +1,478 @@
+"""
+The kcidb.orm.data module defines the data structures and constants used
+by the KCIDB object-relational mapping system to organize Kernel CI report
+data into objects.
```
spbnick, line 4: Let's not put the module name into its description, so we don't forget to update it if we ever rename it. Besides, it's redundant. You know that it's a module, and which module it is when you're reading this. Otherwise this is a pretty good description, thanks! But let's start it with just "The data structures and constants ..." directly.

kcidb/orm/query.py
```diff
@@ -0,0 +1,830 @@
+"""
+The kcidb.orm.query module defines the query classes and functions used
+by the KCIDB object-relational mapping system to retrieve Kernel CI report
+data from the database.
```
spbnick, line 4: Same here, let's just write "The query classes ..." and so on. Otherwise, another nice one!

spbnick (changes requested):
Ah, very neat and tidy :relaxed: Thank you, Abhishek!

I only have a few minor requests left.

The commit message looks very good too, thanks!

It would be totally perfect if you used the imperative mood in the commit body as well (i.e. wrote "Extract parts ...", instead of "We extracted parts ...", and so on), but the most important part is the mood of the commit subject, and that's fine.

kcidb/db/abstract.py
octonawish-akcodes: Oh!!! Okay. Done.

kcidb/orm/query.py
octonawish-akcodes: Done!!!

octonawish-akcodes:
@spbnick Done all the requested changes, Need reviews.

kcidb/orm/__init__.py
```diff
@@ -3,1287 +3,27 @@
...
-import textwrap
 import logging
-import argparse
 from abc import ABC, abstractmethod
-import jsonschema
 import kcidb.io as io
 import kcidb.misc
 from kcidb.misc import LIGHT_ASSERTS
+from kcidb.orm.query import Pattern, PatternHelpAction
+from kcidb.orm.data import SCHEMA
```
spbnick, line 12: Could you please import the `query` and `data` modules here, instead of the symbols directly? And refer to the symbols through them, e.g. `query.Pattern` and `data.SCHEMA`. First, this would satisfy #404, and then it would stop these symbols appearing like they're actually still defined here. It's best to always refer to the symbols by their canonical name and definition location, and not use third-party references (e.g. like the `Pattern` imported here would be).

In other words, the symbols you **moved** out of the `kcidb.orm` module should no longer be available there.

This will also require updating more code in other modules, including `main.py` in the top directory, containing the Cloud Function entry points.

kcidb/orm/__init__.py
```diff
@@ -3,1287 +3,27 @@
...
-        ),
-    )
-)
-
 assert all(k.endswith("s") for k in io.SCHEMA.graph if k), \
     "Not all I/O object list names end with 's'"
 
 assert set(SCHEMA.types) >= \
     set(k[:-1] for k in io.SCHEMA.graph if k), \
     "OO types are not a superset of I/O types"
```
spbnick, line 26: These assertions should go into the `data` submodule, together with the `SCHEMA`, as they're closely tied to it, and are verifying either its pre-conditions, or the `SCHEMA` itself.

spbnick (changes requested):
Thank you very much Abhishek! This looks very good and satisfies all my previous requests.

However, now that your PR is much neater I was able to notice one small, and one larger problem with this. I left you a couple requests inline.

Thank you.

kcidb/orm/__init__.py
octonawish-akcodes: Okay

kcidb/orm/__init__.py
octonawish-akcodes: done!!!

spbnick:
@octonawish-akcodes, I tried reproducing your import problems, and it would go smoother if instead of things like this:
```python
import kcidb.orm.query as query
import kcidb.orm.query as data
```
you would write
```python
from kcidb.orm import query, data
```

spbnick:
We don't have much time left for you to clean this up, and we're blocking #379, so I went ahead and did it for you. Please review the patch I added on top of yours, then squash them together, push to PR, and then review the result again.

Main changes I did:
* Remove all unnecessary changes, including whitespace and import changes. It's generally better to minimize churn and do only what's necessary for a particular task (provided it's done well). We can work on unifying and refining imports later.
* Update documentation so it refers to the new location of the moved symbols. Always remember to update the docs.
* Remove the `LOGGER` global from modules that don't need it. Do not add globals unnecessarily.

spbnick (approved):
LGTM. Thank you, @octonawish-akcodes. Merging!

PR #400, opened 2023-03-18
Reformatting and addition of pydocstyle run in CI

### LINKED ISSUE #277: Add pycodestyle run to CI

octonawish-akcodes:
This closes  #277 
Needed reviews @spbnick 

The following changes added the Reformatting in the whole kcidb repo and followed the conventions stated. 
I also added a pydocstyle CI job in the workflows.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,6 +64,10 @@ jobs:
         run: "pylint kcidb *.py"
         env:
           KCIDB_HOOKS: "pre-commit"
+      - name: Check python sources with pydocstyle
+        run: "pydocstyle kcidb *.py"
+        env:
+          KCIDB_HOOKS: "pre-commit"
       - name: Test with pytest
         run: KCIDB_IO_HEAVY_ASSERTS=1 KCIDB_HEAVY_ASSERTS=1 pytest --tb=native
         env:
--- a/conftest.py
+++ b/conftest.py
@@ -1,4 +1,4 @@
-"""Kernel CI reporting - shared test fixtures"""
+"""Kernel CI reporting - shared test fixtures."""
 
 import os
 import atexit
@@ -52,7 +52,7 @@ EMPTY_DATABASES = {
     else []
 )
 def empty_deployment():
-    """Reusable empty (no-data) deployment"""
+    """Reusable empty (no-data) deployment."""
     yield None
     project = os.environ["GCP_PROJECT"]
     # Empty the load queue subscription
@@ -78,7 +78,7 @@ def empty_deployment():
 
 @pytest.fixture
 def clean_database(_clean_database):
-    """Reusable clean (uninitialized) database"""
+    """Reusable clean (uninitialized) database."""
     yield _clean_database
     if _clean_database.is_initialized():
         _clean_database.cleanup()
@@ -86,13 +86,13 @@ def clean_database(_clean_database):
 
 @pytest.fixture
 def empty_database(_empty_database):
-    """Reusable empty (no-data) database"""
+    """Reusable empty (no-data) database."""
     yield _empty_database
     _empty_database.empty()
 
 
 def pytest_generate_tests(metafunc):
-    """Process a collected pytest test function"""
+    """Process a collected pytest test function."""
     if "empty_database" in metafunc.fixturenames:
         metafunc.parametrize("_empty_database",
                              EMPTY_DATABASES.values(),
--- a/kcidb/__init__.py
+++ b/kcidb/__init__.py
@@ -1,4 +1,4 @@
-"""Kernel CI reporting"""
+"""Kernel CI reporting."""
 
 import sys
 import email
@@ -13,17 +13,19 @@ LOGGER = logging.getLogger(__name__)
 
 
 class DatabaseNotInitialized(Exception):
-    """Database is not initialized exception"""
+    """Database is not initialized exception."""
+
     def __init__(self):
+        """Database not initialized."""
         super().__init__("Database is not initialized")
 
 
 class Client:
-    """Kernel CI reporting client"""
+    """Kernel CI reporting client."""
 
     def __init__(self, database=None, project_id=None, topic_name=None):
         """
-        Initialize a reporting client
+        Initialize a reporting client.
 
         Args:
             database:       The database specification string to use for
@@ -180,7 +182,7 @@ class Client:
 
 
 def submit_main():
-    """Execute the kcidb-submit command-line tool"""
+    """Execute the kcidb-submit command-line tool."""
     sys.excepthook = misc.log_and_print_excepthook
     description = \
         'kcidb-submit - Submit Kernel CI reports, print submission IDs'
@@ -210,7 +212,7 @@ def submit_main():
 
 
 def query_main():
-    """Execute the kcidb-query command-line tool"""
+    """Execute the kcidb-query command-line tool."""
     sys.excepthook = misc.log_and_print_excepthook
     description = \
         "kcidb-query - Query Kernel CI reports"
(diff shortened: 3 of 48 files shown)


.github/workflows/test.yml
```diff
@@ -60,6 +60,8 @@ jobs:
         run: "flake8 kcidb *.py"
       - name: Check python sources with pylint
         run: "pylint kcidb *.py"
+      - name: Check python sources with pycodestyle
```
spbnick, line 63: The name is still referring to `pycodestyle`, not `pydocstyle`. Please fix.

conftest.py
```diff
@@ -8,9 +8,9 @@
 
 
 # Specs of "clean" (uninitialized) databases
-CLEAN_DATABASE_SPECS = list(kcidb.db.misc.parse_spec_list(
-    os.environ.get("KCIDB_CLEAN_TEST_DATABASES", "")
-))
+CLEAN_DATABASE_SPECS = list(
+    kcidb.db.misc.parse_spec_list(os.environ.get("KCIDB_CLEAN_TEST_DATABASES", ""))
+)
```
spbnick, line 13: Hmm, does `pydocstyle` actually request **code** changes? If so, please explain that in the commit message. If not, and you think this should be done for some other reason, please move this and all other code changes to a separate commit, and explain the purpose and reason for the change in the commit message.

kcidb/__init__.py
```diff
@@ -1,29 +1,30 @@
...
 
 
 class Client:
-    """Kernel CI reporting client"""
+    """Kernel CI reporting client."""
 
     def __init__(self, database=None, project_id=None, topic_name=None):
-        """
-        Initialize a reporting client
+        """Initialize a reporting client.
```
spbnick, line 27: Hmm, what was the `pydocstyle` error message that prompted this and other similar changes?

I don't particularly like it, as it makes the description harder to notice in a multi-line docstring. Perhaps there's another way?

kcidb/db/__init__.py
```diff
@@ -8,32 +8,39 @@
...
+        return super().get_doc() + "\n            Example: postgresql bigquery:kcidb_01"
 
     @classmethod
     def get_drivers(cls):
-        """
-        Retrieve a dictionary of driver names and types available for driver's
-        control.
+        """Retrieve a dictionary of driver names and types available for
+        
+        driver's control.
```
spbnick, line 43: Uh-oh, I don't think this would do. The first line is [supposed](https://peps.python.org/pep-0257/#multi-line-docstrings) to be a summary and the lines after it and the blank line - a more elaborate description. Just spilling the original summary over to the elaborate description misses the point. The right solution here should be at least shortening the original summary until it fits the new limit, and then posting the original summary as the elaborate description. Still nicer would be elaborating more there, but that's optional.

Same goes for similar solutions everywhere else.

spbnick (changes requested):
Wooow, that's a lot of changes and work done! Thank you, Abhishek!

However, I thought pydocstyle wouldn't need **any** code changes. Could you explain why you did them? Please see the inline comments as well.

Also, please get rid of the merge commit in your PR.

Thank you :heart:

conftest.py
octonawish-akcodes: No, it doesnt. Actually the large amount changes are of the Reformatting done by me as an extra to make it look more good and readable using a formatting extension in vs code.
Should I move all these formatting changes to separate commit or raise a separate issue reg it and then make a PR for that?

kcidb/__init__.py
octonawish-akcodes: In particular I get these kinds of error 

```
/home/abhishek/Downloads/kernelci/kcidb/kcidb/__init__.py:17 in public method `__init__`:
        D107: Missing docstring in __init__
/home/abhishek/Downloads/kernelci/kcidb/kcidb/__init__.py:22 in public class `Client`:
        D400: First line should end with a period (not 't')

```

conftest.py
spbnick: I appreciate the effort to make the code look nicer. However, please keep **all** reformatting of otherwise-unchanged code in separate commits, and explain why you're doing them in the commit messages. Then we can discuss them. Otherwise they make reviewing harder. 

Please also note that some of your changes take lines over the 79-column limit, and so shouldn't be done, at least in that shape.

Thank you :heart:

kcidb/__init__.py
spbnick: Ah, I meant putting the first line on the same line with the opening `"""`.

It seems it's not necessary (from https://peps.python.org/pep-0257/#multi-line-docstrings):
>  The summary line may be on the same line as the opening quotes or on the next line.

So could you please keep the summary line on the next line, after the opening `"""`?

kcidb/db/__init__.py
octonawish-akcodes: Actually here the pydodstyle was giving me error for adding a blank link after each description first line that's why I have to add it too and I also didn't liked it much.

conftest.py
octonawish-akcodes: > I appreciate the effort to make the code look nicer. However, please keep **all** reformatting of otherwise-unchanged code in separate commits, and explain why you're doing them in the commit messages. Then we can discuss them. Otherwise, they make reviewing harder.
> 
> Please also note that some of your changes take lines over the 79-column limit, and so shouldn't be done, at least in that shape.
> 
> Thank you heart

Yes they are, actually the thing is I'm a bit lost cuz if I try to maintain them under the 78 limit then the pydoc creates the issue like the sentence should end with '.' not with other alphabets and then sometimes gets an error message of blank space needed after first-line description. So this is happening.

kcidb/__init__.py
octonawish-akcodes: Then it throws an error single line comments should end in one line ://

spbnick: Ah, forgot to mention that. Yes. A single-line docstring is fine to be on the same line with **both** the opening **and** closing `"""`.

However, when you have multi-line ones, you tend to miss the first line.

I.e. we want this:
```python
"""Single-line docstring."""
```
and this:
```python
"""
Multi-line docstring summary.

Multi-line docstring elaboration, if any.

Args:
    x: An argument
"""

conftest.py
spbnick: Let's treat the docstrings and the code separately. This PR should only concern itself with docstrings, since it's adding the `pydocstyle` to CI. We can discuss and deal with code formatting separately.

Pydocstyle demands that a multi-line docstring should have a summary line, and it **must** be a single line. Currently the code often has first paragraphs (or several) just containing the description of the function/class, and there is **no** summary line as such. Some (most?) of those are short enough to be on a single line, and pydocstyle would count them as "summaries". Some of the others are longer. So, if you want to do this change, you have to come up with a short summary line for those, and put it above the existing description. However, consider these rules, and use your wits to do what makes best sense, it might not always work out like this. This is a documentation-writing exercise to a considerable extent, not only code and text manipulation.

Lastly, read the [PEP 257](https://peps.python.org/pep-0257/) to understand the complete requirements pydocstyle is trying to enforce.

If you'd like we could have a video call about this and other issues, e.g. tomorrow. Just tell me which times you're available, on Slack, and I'll try to find 30-60 minutes.

kcidb/__init__.py
octonawish-akcodes: Okay

conftest.py
octonawish-akcodes: Okay

octonawish-akcodes:
@spbnick I have done all the changes needed for `pydocstyle` and added a run of it too in the CI and have made changes the pdocstyle was giving me in the project and I have fixed them all. Need reviews.

octonawish-akcodes:
image

What to do now sir?

spbnick:
> What to do now sir?

All development package dependencies are listed in `setup.py`, under `extras_require`. They are installed during the setup of the GitHub Actions workflow. You need to add `pydocstyle` there, so it would get installed.

Regarding changes, please make sure there are no merge commits or unrelated changes (like adding the `.github/workflows/script` file) in there, and also rebase your changes onto latest `main`.

I would recommend cloning another copy of the repo and experimenting with cleaning this up there, so you don't lose your changes.

octonawish-akcodes:
@spbnick I have done all the necessary changes, Pls review

PR #403, opened 2023-03-21
Switch to new BigQuery dataset access API

### LINKED ISSUE #378: Switch to new BigQuery dataset access API
We started receiving the following warning from Google Cloud APIs:
> /home/nkondras/projects/github.com/kernelci/kcidb/kcidb/db/bigquery/v04_00.py:71: PendingDeprecationWarning: Client.dataset is deprecated and will be removed in a future version. Use a string like 'my_project.my_dataset' or a cloud.google.bigquery.DatasetReference object, instead.
> self.dataset_ref = self.client.dataset(dataset_name)

Update the code to silence the warning and prepare for the removal of the old API.

This will very likely require us to make the name of the Google Cloud project mandatory in "database specifications" in the `kcidb.db` package, and all the code using it. In that case, notify the package users of the impeding change (or ask @spbnick to do that), change the code to require it, and then switch to the new API, getting rid of the warning.

shivam-Purohit:
closes #378 

Silence the deprecation warnings and switch to the new BigQuery dataset access API

### DIFF
--- a/kcidb/db/bigquery/v04_00.py
+++ b/kcidb/db/bigquery/v04_00.py
@@ -70,7 +70,9 @@ class Connection(AbstractConnection):
             project_id = None
             dataset_name = params
         self.client = bigquery.Client(project=project_id)
-        self.dataset_ref = self.client.dataset(dataset_name)
+        self.dataset_ref = bigquery.DatasetReference(
+            self.client.project, dataset_name
+        )
         try:
             self.client.get_dataset(self.dataset_ref)
         except GoogleNotFound as exc:


shivam-Purohit:
somehow deprecation warning is silenced?

ubuntu  Running  - Oracle VM VirtualBox 21-03-2023 10_32_31

kcidb/db/bigquery/v04_00.py
```diff
@@ -70,7 +71,7 @@ def __init__(self, params):
             project_id = None
             dataset_name = params
         self.client = bigquery.Client(project=project_id)
-        self.dataset_ref = self.client.dataset(dataset_name)
+        self.dataset_ref = DatasetReference(self.client.project, dataset_name)
```
spbnick, line 74: What happens if the parameter string lacks the project ID?

spbnick (changes requested):
Thank you, @shivam-Purohit.

I left you a crucial question inline.

Apart from that, this change is not just about "silencing the deprecation warning". Just having a warning is not such a big deal. What's worse is that deprecated interfaces get removed in future versions, and that would be a much bigger issue - the code would just break. So the subject of this commit should be something like "bigquery: Update the dataset access interface", and the commit message should explain that the old interface is deprecated and will be removed in the future (apart from any other explanations the change might require).

Finally, as you're now quite acquainted with the project, please make sure CI is passing before review from now on (keep ignoring the deployment failures for now).

kcidb/db/bigquery/v04_00.py
shivam-Purohit: self.dataset_ref = DatasetReference(dataset_name)
E   TypeError: DatasetReference.__init__() missing 1 required positional argument: 'dataset_id'

spbnick: I was talking about the database parameter string supplied to the KCIDB database client. Read the `_PARAMS_DOC` string above.

With the code currently in main, this command should succeed: `kcidb-db-schema -d bigquery:kcidb_01` (provided you have your `GOOGLE_APPLICATION_CREDENTIALS` set correctly). However, after your change it looks like only this command, specifying the project name explicitly, could work `kcidb-db-schema -d bigquery:kcidb-shivam.kcidb_01`, and the former command would break.

We need to figure out if we could still make the former command to work with the updated API, and if not, update the documentation and all the affected interfaces to reflect that.

shivam-Purohit:
Should this be enough?

spbnick:
Yes, the total code change looks good. Thank you, @shivam-Purohit!

Please squash the commits into one, though.

shivam-Purohit:
> Please squash the commits into one, though.

done.

PR #409, opened 2023-03-25
switch from mocks to pytest fixtures

NikkyXO:
The tests functions in kcidb/test_db.py switched from using mocks to pytests fixtures  and the database initialized in the client object constructor in kcidb/db/__init__.py to store its own database attribute @spbnick


spbnick (approved):
Looks great! Thank you, @NikkyXO. Merging.

PR #411, opened 2023-03-25
modify client to notice schema changes

NikkyXO:


NikkyXO:
i had to initialize the database in the constructor, because one of the checks kept failing, even though i already had a separate commit for that @spbnick

kcidb/db/__init__.py
```diff
@@ -134,6 +135,24 @@ def empty(self):
...
+    def reset(self, database=None):
+        """
+        Recreates a database driver instance on request
+        using existing database string
+        Args:
+            database:   A string specifying the database to access, formatted
+                        as "<DRIVER>:<PARAMS>" or just "<DRIVER>". Where
+                        "<DRIVER>" is the driver name, and "<PARAMS>" is the
+                        optional driver-specific database parameter string
+                        or None.
```
spbnick, line 147: The method shouldn't accept a new database string, as its whole purpose is to recreate the client for the existing database. If the user wants to connect to another database, they should just create a new client. Please remove the parameter.

kcidb/db/__init__.py
```diff
@@ -134,6 +135,24 @@ def empty(self):
...
+
+        Returns:
+            A new client object
```
spbnick, line 150: No, this method should not create a new client. We can already do that ourselves. What we need to do is to keep the same `Client` object, but make it re-create the driver instance it's using (`self.driver`).

Essentially you can move the `self.driver = ...` assignment and the following assertion from `__init__()` here, and then call this method from `__init__()` instead.

kcidb/db/__init__.py
```diff
@@ -74,6 +74,7 @@ def __init__(self, database):
                                   driver
         """
         assert isinstance(database, str)
+        self.database = database
```
spbnick, line 77: Normally, what you'd do in this situation, is to cherry-pick the commit you need into this PR's branch, and then add your changes in a separate commit on top, mentioning in the PR description, that it requires that other PR to get merged first. Then you drop the cherry-picked commit after that PR is merged by rebasing your branch on latest `main` and pushing to the PR again.

But you can leave this as is for this PR and focus on the main change for now.

kcidb/db/__init__.py
```diff
@@ -134,6 +135,15 @@ def empty(self):
...
         self.driver.empty()
 
+    def reset(self):
+        """
+        Recreates a database driver instance on request
+        using existing database string
+        Returns:
+            The Recreated driver instance
+        """
+        return Client(self.database)
```
spbnick, line 145: This just creates a new `Client()` instance. This method should instead modify the instance it's called for. It should create a new driver instance and store it in `self.driver`. It should not return anything. As I said in my previous comment:

> Essentially you can move the self.driver = ... assignment and the following assertion from __init__() here, and then call this method from __init__() instead.

kcidb/db/__init__.py
spbnick: Now that [your PR](https://github.com/kernelci/kcidb/pull/409) adding this is merged, you can rebase your PR on top of latest main, and get rid of this change in it.

spbnick (changes requested):

PR #419, opened 2023-03-29
Revert "tests: db: Switch from mocks to fixtures"

spbnick:
Revert the switch to fixtures in kcidb/test_db.py, because the `clean_database` fixture doesn't notice that the database was initialized, as the Client instance stays the same, and tests with BigQuery database fail. We need to add client resetting before trying to cleanup, and that's yet to be implemented.

E.g.: https://github.com/kernelci/kcidb/actions/runs/4543080237/jobs/8007502175

This reverts commit 3caa42f6f6054e3f6daaebc76c52221966237ef5.

FYI: @NikkyXO

### DIFF
--- a/kcidb/test_db.py
+++ b/kcidb/test_db.py
@@ -1,6 +1,7 @@
 """kcdib.db module tests"""
 
 import re
+import textwrap
 import datetime
 import json
 from itertools import permutations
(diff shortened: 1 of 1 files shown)


spbnick:
@NikkyXO, here's an example of why it's important to separate independent changes into commits: we were able to revert the breaking test change, while keeping your feature (storing the database string) intact.

PR #422, opened 2023-03-30
Added test to verify package modules are accessible as symbols

### LINKED ISSUE #404: Test that package modules are accessible as symbols
We'd like to be able to access the package modules as symbols after importing the package's module (`__init__.py`).

I.e. given the `kcidb` package contains the `db` module, this should work:

```python
import kcidb
print(kcidb.db)
```
Most of the code is already built that way. However, we don't have any tests verifying that.

Add a test to `test_kcidb.py` which starts with the `kcidb` package, lists modules under it (e.g. using `pkgutil`, like in [`kcidb.monitor.subscriptions`](https://github.com/kernelci/kcidb/blob/main/kcidb/monitor/subscriptions/__init__.py)), imports the package and verifies that each listed module is available as a symbol with type `<class 'module'>` inside it, then continues to all sub-packages recursively, until everything is verified.

octonawish-akcodes:
This PR added a test to verify package modules are accessible as symbols in the kcidb directory.
This PR closes #404 

@spbnick need reviews

### DIFF
--- a/test_kcidb.py
+++ b/test_kcidb.py
@@ -1,5 +1,7 @@
 """kcidb namespace tests"""
 
+import pkgutil
+import importlib
 import re
 import json
 import textwrap
@@ -30,6 +32,25 @@ def test_json_output_options():
                     stdout_re="\\{\n    \".*")
     assert_executes("", "kcidb.schema_main", "--indent=2",
                     stdout_re="\\{\n  \".*")
+
+
+def test_package_symbols():
+    """
+    Test that package modules are accessible as symbols after importing
+    the package's module (__init__.py).
+    """
+
+    def check_package(package):
+        for _, module_name, is_pkg in pkgutil.iter_modules(package.__path__):
+            module_fullname = f"{package.__name__}.{module_name}"
+            module = importlib.import_module(module_fullname)
+            symbol = getattr(package, module_name, None)
+            assert symbol is module, \
+                f"{package.__name__} missing for submodule {module_name}. "\
+                f"Expected symbol is {module}, but actual symbol is {symbol}."
+            if is_pkg:
+                check_package(module)
+    check_package(kcidb)
 
 
 def test_submit_main():


spbnick:
Please make sure CI passes before requesting reviews. Unless you need help with something. Thank you :heart:

octonawish-akcodes:
@spbnick Have a look now?

test_kcidb.py
```diff
@@ -32,6 +33,40 @@ def test_json_output_options():
...
+    package_path = kcidb.__path__[0]
+    package_name = kcidb.__name__
+    package_modules = [name for _, name, _
+                       in pkgutil.iter_modules([package_path])]
+
+    for module_name in package_modules:
+        module_fullname = f"{package_name}.{module_name}"
+        module = __import__(module_fullname, fromlist=["*"])
+        assert module_name in dir(kcidb)
+        assert isinstance(getattr(kcidb, module_name), type(module))
```
spbnick, line 56: I wonder if something like `assert getattr(kcidb, module_name, None) is module` would work instead.

test_kcidb.py
```diff
@@ -32,6 +33,40 @@ def test_json_output_options():
...
+
+        if hasattr(module, "__path__"):
```
spbnick, line 58: What is this check doing?

test_kcidb.py
```diff
@@ -32,6 +33,40 @@ def test_json_output_options():
...
+    Starting with the kcidb package, this test lists all modules under
+    it using pkgutil, imports the package and verifies that each listed
+    module is available as a symbol with type <class 'module'> inside it.
+    It then continues to all sub-packages recursively until everything
+    is verified.
+    """
+    package_path = kcidb.__path__[0]
+    package_name = kcidb.__name__
+    package_modules = [name for _, name, _
+                       in pkgutil.iter_modules([package_path])]
```
spbnick, line 50: Note that `pkgutil.iter_modules()` returns a boolean signifying if the module is a package or not, as the third value you're ignoring here. That should help you make the decisions in the loop. That would be easier if you use the `pkgutil.iter_modules()` in the loop directly, instead of storing the module names in a list. Something like this:

```python
for _, name, is_package in pkgutil.iter_modules([package_path]):
    # ...
```

spbnick:
Also, it's strange that this doesn't fail for any of the packages you managed to check here :thinking: 
See if you can make it actually catch a missing module symbol by removing one from a package.

test_kcidb.py
```diff
@@ -32,6 +33,40 @@ def test_json_output_options():
...
+
+    for module_name in package_modules:
+        module_fullname = f"{package_name}.{module_name}"
+        module = __import__(module_fullname, fromlist=["*"])
```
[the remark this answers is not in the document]
octonawish-akcodes: Okay

test_kcidb.py
octonawish-akcodes: I will look into it.

test_kcidb.py
octonawish-akcodes: It checks whether the module object is a package and if so then it will check the sub modules the same way top level module is being checked.

test_kcidb.py
octonawish-akcodes: Okay

test_kcidb.py
```diff
@@ -32,6 +34,30 @@ def test_json_output_options():
...
+def test_package_symbols():
+    """
+    Test that package modules are accessible as symbols after importing
+    the package's module (__init__.py).
+
+    Starting with the kcidb package, this test lists all modules under
+    it using pkgutil, imports the package and verifies that each listed
+    module is available as a symbol with type <class 'module'> inside it.
+    It then continues to all sub-packages recursively until everything
+    is verified.
```
spbnick, line 46: I think the second paragraph here is redundant and describes the algorithm, which is better discovered by reading the actual (very short) code. Could you delete it?

spbnick (changes requested):
Thank you, @octonawish-akcodes! This is pretty good already. I left you a couple requests inline.

Otherwise, have you tested this? E.g. by removing a symbol for a submodule, or adding a (dummy) submodule without adding the symbol?

test_kcidb.py
octonawish-akcodes: Okay

test_kcidb.py
```diff
@@ -32,6 +34,30 @@ def test_json_output_options():
...
+    """
+
+    def check_package(package):
+        for _, module_name, is_pkg in pkgutil.iter_modules(package.__path__):
+            module_fullname = f"{package.__name__}.{module_name}"
+            module = importlib.import_module(module_fullname)
+            assert hasattr(package, module_name)
+            assert isinstance(getattr(package, module_name), type(module))
```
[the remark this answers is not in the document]
spbnick: Could you figure out why, and what the solution should be? This test would be useless otherwise.

PR #425, opened 2023-04-04
Update submitter guide to use v4.1 schema

NikkyXO:


NikkyXO:
@spbnick please review

PR #426, opened 2023-04-04
submitter_guide: add sections on issues and incidents

NikkyXO:


NikkyXO:
please review added sections @spbnick, am i on right track

doc/submitter_guide.md
```diff
@@ -312,6 +312,41 @@ Date/Time Format][datetime_format].
...
+
+* `ERROR` - the test is faulty, the status of the tested code is unknown.
+* `FAIL` - the test has failed, the tested code is faulty.
+* `PASS` - the test has passed, the tested code is correct.
+* `DONE` - the test has finished successfully, the status of the tested code
+  is unknown.
+
+##### `culprit`
+The layers of the execution stack responsible for the issue
+It can be one of the following properties
```
spbnick, line 333: "It can contain one or more of the following properties" would be more correct.

doc/submitter_guide.md
```diff
@@ -312,6 +312,41 @@ Date/Time Format][datetime_format].
...
+
+* `Code` - the built or tested code, value is boolean
+* `Tool` - the static analyzer, the build toolchain, the test, etc
+* `Harness` - the system controlling the execution of the build.
```
spbnick, line 337: Each of these properties is all-lower-case, and should be listed as such, as anything else won't be accepted by the schema. All these properties are boolean, so perhaps just mention that in the paragraph above.

doc/submitter_guide.md
```diff
@@ -312,6 +312,41 @@ Date/Time Format][datetime_format].
...
+Example: `false`
+
+##### `test_status`
+The Status to assign to incident tests, it can be one of the following
+
+* `ERROR` - the test is faulty, the status of the tested code is unknown.
+* `FAIL` - the test has failed, the tested code is faulty.
+* `PASS` - the test has passed, the tested code is correct.
+* `DONE` - the test has finished successfully, the status of the tested code
+  is unknown.
```
spbnick, line 329: This is an incomplete list of possible status strings. It might be better to just refer to the description of the test `status` field above, as they're the same type.

doc/submitter_guide.md
```diff
@@ -312,6 +312,41 @@ Date/Time Format][datetime_format].
...
+
+##### `culprit`
+The layers of the execution stack responsible for the issue
+It can be one of the following properties
+
+* `Code` - the built or tested code, value is boolean
+* `Tool` - the static analyzer, the build toolchain, the test, etc
+* `Harness` - the system controlling the execution of the build.
+
+#### Incidents
```
spbnick, line 339: This could use a general description of what these objects are about and for.

spbnick (changes requested):
This is a good start, you get the idea of giving a limited description, which is easier to grasp. Apart from the inline comments, it would also be good to give an example of value for each field.

NikkyXO:
> This is a good start, you get the idea of giving a limited description, which is easier to grasp. Apart from the inline comments, it would also be good to give an example of value for each field.

I made recommended changes @spbnick

doc/submitter_guide.md
```diff
@@ -312,6 +312,94 @@ Date/Time Format][datetime_format].
 
 Example: `"2020-08-14T23:41:54+00:00"`
 
+#### Issues
+The following properties are used to describe issues found in reports
+
+##### `build_valid`
+The status to assign to incident builds
+
+Example: `"build_valid": false`
```
spbnick, line 321: It's enough to just show the value in the example, like the existing examples for all other properties. No need to also show the property name.

I.e. this:
```markdown
Example: `false`
```

And not this:
```markdown
Example: `"build_valid": false`
```

Here and everywhere else.

doc/submitter_guide.md
```diff
@@ -312,6 +312,94 @@ Date/Time Format][datetime_format].
...
+
+##### `present`
+True if the issue occurred in the linked objects, False if it was absent
+
+Example: `"present": True`
+
+##### `comment`
+A human-readable comment regarding the incident
+
+Example: `"comment": "<An human readable comment in string>"`
```
spbnick, line 367: Could you make up a comment that would be suitable for an incident?

spbnick (changes requested):
Thank you, @NikkyXO. I left a few inline comments concerning your new changes. There are also some of the old ones to resolve.

doc/submitter_guide.md
NikkyXO: @spbnick i reffered to the test:status section, the available status strings there are: Error, Fail, Pass, Done, Skip. That was what i also included

PR #429, opened 2023-05-08
Support I/O schema v4.2

spbnick:
Add support for I/O schema v4.2, which adds another test status: "MISS".

### DIFF
--- a/doc/submitter_guide.md
+++ b/doc/submitter_guide.md
@@ -273,8 +273,8 @@ Example: `"2020-08-14T23:08:10.008000+00:00"`
 ##### `status`
 The test status string, one of the following.
 
-* `ERROR` - the test is faulty, the status of the tested code is unknown.
 * `FAIL` - the test has failed, the tested code is faulty.
+* `ERROR` - the test is faulty, the status of the tested code is unknown.
 * `PASS` - the test has passed, the tested code is correct.
 * `DONE` - the test has finished successfully, the status of the tested code
   is unknown.
--- a/doc/submitter_guide.md
+++ b/doc/submitter_guide.md
@@ -271,20 +271,59 @@ Example: `"2020-08-14T23:08:10.008000+00:00"`
 #### Tests
 
 ##### `status`
-The test status string, one of the following.
+The test status string, one of the following:
 
-* `FAIL` - the test has failed, the tested code is faulty.
-* `ERROR` - the test is faulty, the status of the tested code is unknown.
-* `PASS` - the test has passed, the tested code is correct.
-* `DONE` - the test has finished successfully, the status of the tested code
-  is unknown.
-* `SKIP` - the test wasn't executed, the status of the tested code is unknown.
+* "FAIL" - the test completed and reported the tested code as faulty.
+* "ERROR" - the test didn't complete due to a failure in its code, and the
+  status of the tested code is unknown.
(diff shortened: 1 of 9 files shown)


doc/submitter_guide.md
```diff
@@ -271,20 +271,24 @@ Example: `"2020-08-14T23:08:10.008000+00:00"`
...
-* `PASS` - the test has passed, the tested code is correct.
-* `DONE` - the test has finished successfully, the status of the tested code
-  is unknown.
-* `SKIP` - the test wasn't executed, the status of the tested code is unknown.
+The test status string, one of the following:
+
+* "FAIL" - the test completed and reported the tested code as faulty.
+* "ERROR" - the test didn't complete due to a failure in its code, and the
+  status of the tested code is unknown.
+* "MISS" - the test didn't run due to a failure in the test harness, and the
```
tales-aparecida, line 279: I feel like we could extend this explanation a little bit to include the ideas that justified its inclusion in the list, for example, we should make it clear how `MISS` is not `null` and when `MISS` instead of `ERROR`

kcidb/templates/misc.j2
```diff
@@ -13,11 +13,12 @@
 
 LEGEND
 
-    💥 ERROR    - Aborted. Test, tested code, or both might be faulty.
     ❌ FAIL     - Failed. Tested code is likely faulty.
+    💥 ERROR    - Aborted. Test, tested code, or both might be faulty.
+    ⚫ MISS     - Missing. Planned, but failed to execute.
```
octonawish-akcodes, line 18: IMO we should use the 🚫  emoji in MISS, this can be more relevant as it suggests that something is missing.

kcidb/templates/test.j2
```diff
@@ -1,8 +1,9 @@
 {# Test macros #}
 
 {% macro status_emoji(status) %}
-    {{- "💥" if status == "ERROR" else
-        "❌" if status == "FAIL" else
+    {{- "❌" if status == "FAIL" else
+        "💥" if status == "ERROR" else
+        "⚫" if status == "MISS" else
```
octonawish-akcodes, line 6: The same goes here too, this 🚫 can be used.

kcidb/templates/misc.j2
octonawish-akcodes: We can also think of "🟡" (yellow circle) or "🟠" (orange circle).

kcidb/templates/test.j2
octonawish-akcodes: Same here too "🟡" (yellow circle) or "🟠" (orange circle).

kcidb/templates/misc.j2
octonawish-akcodes: Got it, Yeah then a black circle works fine.

tales-aparecida (approved):

doc/submitter_guide.md
spbnick: Expanded on the statuses, in the doc. Thanks!

spbnick:
I've changed the MISS emoji to 🟩. Contrasted with ✅ PASS it looks like something is missing. Since we expect more developers (who are less concerned with missing tests) to look at results in general, than CI system or harness maintainers, I think this is a good default. Let's return to this when we have more data and feedback from users.

tales-aparecida (approved):
Great work on the description, it's very clear!

PR #435, opened 2023-07-04
added cacging script gen1 function in main file

octonawish-akcodes:
Need reviews @spbnick and how to test it.


main.py
```diff
@@ -347,3 +347,21 @@ def send_message(message):
         finally:
             # Disconnect from the SMTP server
             smtp.quit()
+
+
+def pubsub(event, context):
```
spbnick: Please rename this to `kcidb_cache_urls` as we've discussed.

main.py
```diff
@@ -347,3 +347,21 @@ def send_message(message):
...
+    # Extract the Pub/Sub message data
+    pubsub_message = base64.b64decode(event['data']).decode()
+
+    # Split the message into individual URLs
+    urls = pubsub_message.splitlines()
```
spbnick, line 357: Just do this in the loop statement directly. The code would be short enough.

main.py
```diff
@@ -347,3 +347,21 @@ def send_message(message):
...
+
+    # Create an instance of the Cache class
+    cache = Client("cache_file_storage", 5 * 1024 * 1024)
```
spbnick, line 360: Please follow the pattern of `get_db_client()`/`get_oo_client()`/etc. to only create the cache client once per instance and only when necessary.

main.py
```diff
@@ -347,3 +347,21 @@ def send_message(message):
...
+
+    # Cache each URL
+    for url in urls:
+        cache.store(url)
+
+    # Print out the data from Pub/Sub, to prove that it worked
+    print("Hello, " + pubsub_message + "!")
```
spbnick: Please remove this printout. If you need that, convert it to a `LOGGER.debug()` call (similar to other places here), logging the message at the **beginning** of the function, before it has a chance to crash.

main.py
```diff
@@ -7,7 +7,7 @@
 import logging
 import smtplib
 import kcidb
-
+from kcidb.cache import Client
```
spbnick, line 10: Let's follow the pattern of just reaching to the client from `kcidb` package, as everywhere else her. This saves us a lot of imports as this file uses basically the complete functionality of KCIDB, because this is where it all comes together. That is, do not import `kcidb.cache` directly, and refer to the client via `kcidb.cache.Client`.

kcidb/cache/__init__.py
```diff
@@ -0,0 +1,149 @@
+import hashlib
```
spbnick, line 1: Please add a module docstring.

kcidb/cache/__init__.py
```diff
@@ -0,0 +1,149 @@
...
+        self.bucket_name = bucket_name
+        self.max_store_size = max_store_size
+        self.client = storage.Client()
+
+    def _extract_content_disposition(self, response):
+        content_disposition = response.headers.get("Content-Disposition")
+
+        if content_disposition is None:
+            parsed_url = urllib.parse.urlparse(response.url)
+            original_filename = parsed_url.path.rsplit('/', 1)[-1]
```
spbnick, line 24: What happens if there are no `/` characters? Is that possible? What if there's only one, do we get a usable filename?

kcidb/cache/__init__.py
```diff
@@ -0,0 +1,149 @@
...
+
+            encoded_filename = Header(original_filename, "utf-8").encode()
+
+            content_disposition = f'inline; filename="{encoded_filename}"'
```
spbnick, line 28: Let's switch to `attachment` disposition to be safe, here. We don't want browsers to try opening binary files directly. We can switch to using `inline` selectively for certain MIME types if that becomes a problem, later.

kcidb/cache/__init__.py
```diff
@@ -0,0 +1,149 @@
...
+        downloaded if it's already in the cache or if it doesn't match the
+        requirements (max_store_size).
+        Args:
+            url:    The URL to try to cache.
+        """
+        object_name = self._format_object_name(url)
+        blob = self.client.bucket(self.bucket_name).blob(object_name)
+
+        if blob.exists():
+            print(f"URL '{url}' is already cached.")
```
spbnick, line 44: Please remove printouts here and everywhere else in this module. These are not suitable for a library. Instead, if needed, we should provide a way for the caller to find out what exactly happened (e.g. through a return value) and make a decision what to do with that, whether to `print` something, to log a message, or do something else.

spbnick (changes requested):
This is a good start! Thank you @octonawish-akcodes.

main.py
octonawish-akcodes: okay

main.py
octonawish-akcodes: Ah!!! got it

kcidb/cache/__init__.py
octonawish-akcodes: Okkay

kcidb/cache/__init__.py
octonawish-akcodes: done

kcidb/cache/__init__.py
octonawish-akcodes: okay

main.py
octonawish-akcodes: Got it

main.py
octonawish-akcodes: Okay

spbnick (changes requested):
Looks alright to me, thank you, Abhishek, but let's work on the commit message a bit.

First of all, please write your commit messages in imperative mood (as if giving a command). I.e. instead of writing "Added", please write "Add".

Then let's work on the meaning. Right now you're saying you've added a "script". "Script" usually means a quickly put-together (somewhat poorly-made) program, an executable, which you can run by itself. You're adding no such thing. What you're adding here is a **module** (`kcidb.cache`), a couple functions in another module (`main.py`), and their dependencies.

Additionally, you're saying that you're adding a "gen1 function". However, there's no such thing as a "gen1 function" in general, without the context of Google Cloud Functions, which you don't have in your message. So it would make little sense to whoever else reads that. Further on, "gen1" is just an implementation detail which has little practical effect, and is especially pointless to mention, since *all* our Cloud Functions are "gen1".

What I would write instead, is something along the lines of "Add URL cache client and a caching Cloud Function". This says that we're adding a "URL cache client", without going into details of how it's implemented (i.e. the module, as that's normally how it's done), and mentions the Cloud Function, as that's an important high-level interface, but skips specifying which generation it is, as that's not really important, and because we gotta keep our commit subjects short.

In a larger community, it would be important to elaborate the details and the purpose in the commit message body, but we all know what this is about, here.

octonawish-akcodes:
Ah!!! okay. I will remember this in future PR's

PR #437, opened 2023-07-18
cloud: add storage support

octonawish-akcodes:
Added support for cloud storage deployment through cloud script. Need reviews @spbnick

### DIFF
--- a/cloud
+++ b/cloud
@@ -1243,6 +1243,22 @@ function pubsub_withdraw() {
     pubsub_topic_withdraw "$project" "$pick_notifications_trigger_topic"
 }
 
+# Deploy a Google Cloud Storage Bucket
+# Args: project storage_bucket
+function storage_deploy() {
+    declare -r project="$1"; shift
+    declare -r storage_bucket="$1"; shift
+    gsutil mb -p "$project" -c STANDARD -l "us-central1" "gs://$storage_bucket"
+    gsutil iam ch allUsers:objectViewer "gs://$storage_bucket/"
+}
+
+# Remove a Google Cloud Storage Bucket and its contents
+# Args: storage_bucket
+function storage_withdraw() {
+    declare -r storage_bucket="$1"; shift
+    gsutil -m rm -r "gs://$storage_bucket"
+}
+
 # Deploy a Cloud Function
 # Args: project name source [param_arg...]
 function cloud_function_deploy() {
@@ -1616,6 +1632,7 @@ declare -A -r SECTIONS=(
     ["pubsub"]="Pub/Sub topics and subscriptions"
     ["secrets"]="Secrets"
     ["firestore"]="Firestore database"
+    ["storage"]="Google cloud storage"
     ["cloud_functions"]="Cloud Functions"
     ["scheduler"]="Scheduler jobs"
     ["submitters"]="Submitter permissions"
@@ -1725,6 +1742,7 @@ function execute_command() {
     eval "declare -a argv=($argv)"
 
     declare -r load_queue_trigger_topic="${prefix}load_queue_trigger"
+    declare -r storage_bucket="${project}${prefix}cache"
     declare -r pick_notifications_trigger_topic="${prefix}pick_notifications_trigger"
     declare -r new_topic="${prefix}new"
     declare -r new_load_subscription="${prefix}new_load"
@@ -1882,7 +1900,7 @@ function execute_command() {
     app_deploy "$project"
     services_enable "$project" appengine secretmanager cloudfunctions \
                                cloudbuild cloudscheduler firestore \
-                               sqladmin
+                               sqladmin storage
 
     # Enable generating PostgreSQL superuser password (if not specified),
     # if the instance doesn't exist yet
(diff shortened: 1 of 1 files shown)


cloud
```diff
@@ -1243,6 +1243,20 @@ function pubsub_withdraw() {
     pubsub_topic_withdraw "$project" "$pick_notifications_trigger_topic"
 }
 
+function storage_deploy(){
```
spbnick, line 1246: Please observe the coding conventions regarding placement of the opening brace `{`.

Please document this and the other function, following conventions set by other functions here.

cloud
```diff
@@ -1243,6 +1243,20 @@ function pubsub_withdraw() {
...
+    
+    gsutil mb -p "$project" -c STANDARD -l "us-central1" "gs://$storage_bucket"
+
+    gsutil iam ch allUsers:objectViewer "gs://$storage_bucket/"
+
```
spbnick, line 1251: Please remove unnecessary empty lines around the inside of the function body (this one, and the one at the beginning). In the other function as well.

cloud
```diff
@@ -1911,6 +1927,9 @@ function execute_command() {
             --smtp-topic="$smtp_topic" \
             --smtp-subscription="$smtp_subscription"
         sections_run "$sections" firestore_deploy "$project"
+        sections_run "$sections" storage_deploy \
+            --project="$project" \
+            --bucket="$storage_bucket"
```
spbnick, line 1932: You're passing these as options, but nothing reads them. This only works because of the weird scoping rules of the shell, where a function called from another function has access to the caller's variables. It exists, but we should never use it, because it makes the code very hard to read, and should pass our arguments explicitly. The same way it's done in other functions. Additionally, we shouldn't be using options for a function with so little arguments. Use positional arguments instead.

spbnick (changes requested):
Glad to see you made it work!

I left a few comments inline.

Once you dealt with these, please make sure you're enabling the Google Cloud Storage service for the project in the `execute_command()` function (where all the other services are enabled).

cloud
octonawish-akcodes: Okay

cloud
octonawish-akcodes: Okay

cloud
octonawish-akcodes: got it, thank you. I realized it later better after your comment.

octonawish-akcodes:
> Also, please make sure to squash the commits into a good shape before requesting reviews. Thanks!

Done

cloud
```diff
@@ -1911,6 +1926,7 @@ function execute_command() {
             --smtp-topic="$smtp_topic" \
             --smtp-subscription="$smtp_subscription"
         sections_run "$sections" firestore_deploy "$project"
+        sections_run "$sections" storage_deploy "$project" "$storage_bucket"
```
spbnick, line 1929: You're passing the arguments, but still nothing actually receives them in the functions. This still only works because of the quirk of Bash namespacing. Please receive the arguments explicitly in the functions you've added. Look at other functions accepting positional arguments for how to do it.

spbnick (changes requested):
Looks great apart from argument passing (comment inline).

cloud
```diff
@@ -1725,6 +1742,7 @@ function execute_command() {
     eval "declare -a argv=($argv)"
 
     declare -r load_queue_trigger_topic="${prefix}load_queue_trigger"
+    declare -r storage_bucket="${project}${prefix}cache"
```
spbnick, line 1745: Please add `_` between `${project}` and `${prefix}`. This will make the bucket name easier to read.

spbnick (changes requested):

cloud
```diff
@@ -1243,6 +1243,22 @@ function pubsub_withdraw() {
     pubsub_topic_withdraw "$project" "$pick_notifications_trigger_topic"
 }
 
+# Deploy a Google Cloud Storage Bucket
+# Args: project storage_bucket
+function storage_deploy() {
+    declare -r project="$1"; shift
+    declare -r storage_bucket="$1"; shift
```
spbnick, line 1250: Once we passed the storage bucket to this (and the other function), it's clear that it's a **storage** bucket, so we don't need to have `storage_` in the variable name. Please name it just `bucket` in these functions (but not in the calling function, where there is more context we need to separate it from).

cloud
```diff
@@ -1243,6 +1243,22 @@ function pubsub_withdraw() {
...
+    gsutil mb -p "$project" -c STANDARD -l "us-central1" "gs://$storage_bucket"
+    gsutil iam ch allUsers:objectViewer "gs://$storage_bucket/"
```
spbnick, line 1252: Please stop these commands generating output, like "bucket created" and "bucket removed".

spbnick (changes requested):

spbnick:
To fix the `gsutil` reporting an error after `cloud` exits, save the original `TMPDIR` variable in another read-only global variable called `ORIG_TMPDIR`, and then supply it as the `TMPDIR` to every invocation of `gsutil`.

cloud
spbnick: Use the `-q` option to `gsutil`. Like this: `gsutil -q mb blah blah`.

spbnick:
Please also use the `-b on` option to `gsutil mb`, so we get the uniform (simpler) access control.

octonawish-akcodes:
Shifting this PR to Upstream directly as a separate branch PR #438

octonawish-akcodes:
Closing as work shifted to #438

PR #439, opened 2023-07-22
cloud: Add storage_bucket to cloud functions env

octonawish-akcodes:
Added `storage_bucket` to cloud functions environment so that cache client can use that bucket name and store the artifacts in there.

Need reviews @spbnick

### DIFF
--- a/cloud
+++ b/cloud
@@ -1298,6 +1298,7 @@ function cloud_function_withdraw() {
 # Output deployment environment for Cloud Functions
 # Args: --format=yaml|sh
 #       --log-level=NAME
+#       --cache-bucket-name=NAME
 #       --optimize=PYTHONOPTIMIZE
 #       --heavy-asserts=true|false
 #       --new-topic=NAME --new-load-subscription=NAME
@@ -1323,6 +1324,7 @@ function cloud_functions_env() {
                           smtp_to_addrs smtp_password_secret \
                           smtp_topic smtp_subscription \
                           pgpass_secret \
+                          cache_bucket_name \
                           database \
                           clean_test_databases \
                           empty_test_databases \
@@ -1352,6 +1354,9 @@ function cloud_functions_env() {
     )
     if [ -n "$extra_cc" ]; then
         env[KCIDB_EXTRA_CC]="$extra_cc"
+    fi
+    if [ -n "$cache_bucket_name" ]; then
+        env[KCIDB_CACHE_BUCKET_NAME]="$cache_bucket_name"
     fi
     if [ -n "$smtp_to_addrs" ]; then
         env[KCIDB_SMTP_TO_ADDRS]="$smtp_to_addrs"
@@ -1764,7 +1769,7 @@ function execute_command() {
     eval "declare -a argv=($argv)"
 
     declare -r load_queue_trigger_topic="${prefix}load_queue_trigger"
-    declare -r storage_bucket="${project}_${prefix}cache"
+    declare -r cache_bucket_name="${project}_${prefix}cache"
     declare -r pick_notifications_trigger_topic="${prefix}pick_notifications_trigger"
     declare -r updated_urls_topic="${prefix}updated_urls"
     declare -r new_topic="${prefix}new"
@@ -1891,6 +1896,7 @@ function execute_command() {
 
     declare -r -a env_args=(
         --log-level="$log_level"
+        --cache-bucket-name="$cache_bucket_name"
         --optimize="$optimize"
         --heavy-asserts="$heavy_asserts"
         --new-topic="$new_topic"
@@ -1953,7 +1959,7 @@ function execute_command() {
             --smtp-topic="$smtp_topic" \
             --smtp-subscription="$smtp_subscription"
         sections_run "$sections" firestore_deploy "$project"
-        sections_run "$sections" storage_deploy "$project" "$storage_bucket"
+        sections_run "$sections" storage_deploy "$project" "$cache_bucket_name"
         sections_run "$sections" cloud_functions_deploy \
             --project="$project" \
             --prefix="$prefix" \
@@ -1985,7 +1991,7 @@ function execute_command() {
             "$project" "$new_topic" "${submitters[@]}"
         sections_run "$sections" scheduler_withdraw "$project" "$prefix"
         sections_run "$sections" cloud_functions_withdraw "$project" "$prefix"
-        sections_run "$sections" storage_withdraw "$storage_bucket"
+        sections_run "$sections" storage_withdraw "$cache_bucket_name"
         sections_run "$sections" firestore_withdraw \
             "$project" "$spool_collection_path"
         sections_run "$sections" pubsub_withdraw \
--- a/main.py
+++ b/main.py
@@ -71,6 +71,8 @@ UPDATED_PUBLISH = bool(os.environ.get("KCIDB_UPDATED_PUBLISH", ""))
 # The publisher object for the queue with patterns matching objects updated by
 # loading submissions.
 _UPDATED_QUEUE_PUBLISHER = None
+# KCIDB cache storage bucket name
+CACHE_BUCKET_NAME = os.environ.get("KCIDB_CACHE_BUCKET_NAME")
 
 
 def get_smtp_publisher():
@@ -356,7 +358,7 @@ def get_cache_client():
     global _CACHE_CLIENT
     if _CACHE_CLIENT is None:
         _CACHE_CLIENT = kcidb.cache.Client(
-            "cache_file_storage", 5 * 1024 * 1024
+            CACHE_BUCKET_NAME, 5 * 1024 * 1024
         )
     return _CACHE_CLIENT
 


octonawish-akcodes:
Supported test conducted:

Pub/Sub message published to `kcidb_updated_urls` topic
image

`kcidb_cache_urls` logs
image

Deployment:
image

cloud
```diff
@@ -1353,6 +1353,9 @@ function cloud_functions_env() {
     if [ -n "$extra_cc" ]; then
         env[KCIDB_EXTRA_CC]="$extra_cc"
     fi
+    if [ -n "$storage_bucket" ]; then
+        env[CACHE_BUCKET_NAME]="$storage_bucket"
```
spbnick, line 1357: The environment variable name should start with `KCIDB_`. I.e. this one would become `KCIDB_CACHE_BUCKET_NAME`.

main.py
```diff
@@ -355,8 +355,9 @@ def get_cache_client():
     # It's alright, pylint: disable=global-statement
     global _CACHE_CLIENT
     if _CACHE_CLIENT is None:
+        bucket_name = os.environ.get('CACHE_BUCKET_NAME')
```
spbnick, line 358: Could you follow the pattern of storing the bucket name in an (appropriately-named) global variable, same way as the rest of the clients (database, Pub/Sub, etc.) get their parameters?

cloud
spbnick: Please pass the argument via a parameter **explicitly**, not via shared namespace.
Also, a better name for both the `cloud_functions_env()` parameter and the variable inside `execute_command()` would be something like `cache_bucket_name`, since `storage_bucket` is a bit too generic.

spbnick (changes requested):

main.py
```diff
@@ -356,7 +358,7 @@ def get_cache_client():
     global _CACHE_CLIENT
     if _CACHE_CLIENT is None:
```
tales-aparecida, line 359: I guess we'd probably want to check CACHE_BUCKET_NAME taking get_smtp_publisher as an example

```suggestion
    if _CACHE_CLIENT is None and CACHE_BUCKET_NAME is not None:
```

main.py
```diff
@@ -71,6 +71,8 @@
 # The publisher object for the queue with patterns matching objects updated by
 # loading submissions.
 _UPDATED_QUEUE_PUBLISHER = None
+# Storage bucket name
```
tales-aparecida, line 74:
```suggestion
# KCIDB cache storage bucket name
```

Based on the description we've used on _CACHE_CLIENT

main.py
tales-aparecida: Otherwise, I don't know if y'all have discussed about this... do we want to keep using "cache_file_storage" as a default value for CACHE_BUCKET_NAME?

cloud
```diff
@@ -1323,6 +1323,7 @@ function cloud_functions_env() {
                           smtp_to_addrs smtp_password_secret \
                           smtp_topic smtp_subscription \
                           pgpass_secret \
+                          cache_bucket_name \
```
tales-aparecida, line 1326: This should show up in "Args:", don't it?

cloud
```diff
@@ -2299,6 +2305,7 @@ function execute() {
     declare test="false"
     declare log_level="INFO"
     declare optimize=""
+    declare cache_bucket_name=""
```
tales-aparecida, line 2308: nit; Move this between `log_level` and `optimize`, just so we can keep it uniform with the rest of the code

cloud
```diff
@@ -1764,7 +1769,7 @@ function execute_command() {
     eval "declare -a argv=($argv)"
 
     declare -r load_queue_trigger_topic="${prefix}load_queue_trigger"
-    declare -r storage_bucket="${project}_${prefix}cache"
+    declare -r cache_bucket_name="${project}_${prefix}cache"
```
tales-aparecida, line 1772: I might've misunderstand what Nik meant by "not via shared namespace", but I guess it includes removing this definition (given that `--prefix="${namespace:-}${namespace:+_}kcidb_" \`)

You can see that `cache_bucket_name` is already defined via getopt_vars a few lines above...
Keeping this line here you should see that the bucket was using the previous naming scheme.

main.py
octonawish-akcodes: Hmm we havent discussed about that, As much as I remember from my meeting with Nik.

main.py
octonawish-akcodes: okay

cloud
octonawish-akcodes: okay

cloud
```diff
@@ -1752,6 +1756,7 @@ function execute_command() {
                           smtp_mocked \
                           test \
                           log_level \
+                          cache_bucket_name \
```
spbnick, line 1759: We don't need to pass this parameter to `execute_command()`, because it is supposed to generate it itself, from the project name and the prefix.

cloud
spbnick: Thanks, @tales-aparecida. We were not supposed to have the `cache_bucket_name` as a parameter to this function. It's supposed to be generated right here, as written above.

What I [meant](https://github.com/kernelci/kcidb/pull/439#discussion_r1271876093) by "not via shared namespace" was that @octonawish-akcodes was not passing the `storage_bucket` to the `cloud_functions_env()` function explicitly (it didn't have it as a parameter), and his code only worked because in shell a function sees the variables declared in the calling function :see_no_evil:

main.py
spbnick: Also, since we're requiring the `CACHE_BUCKET_NAME` to be set, we don't need to check for it here.

spbnick (changes requested):
Thank you, @octonawish-akcodes, @tales-aparecida! I left some comments inline.

main.py
tales-aparecida: >  requiring the CACHE_BUCKET_NAME to be set

Should we be using `CACHE_BUCKET_NAME = os.environ["KCIDB_CACHE_BUCKET_NAME"]` then?

octonawish-akcodes: Yes

cloud
```diff
@@ -1735,6 +1739,7 @@ function escape_whitespace() {
 #       --smtp-mocked=true|false
 #       --test=true|false
 #       --log-level=NAME
+#       --cache-bucket-name=NAME
```
tales-aparecida, line 1742:
```suggestion
```

cloud
tales-aparecida: Actually, based on https://github.com/kernelci/kcidb/pull/439/files#r1273098653 just

```suggestion
```

cloud
```diff
@@ -2267,7 +2272,7 @@ function execute() {
         fi
         if [[ $command == @(deploy|env|shell) ]]; then
             getopt_longopts+=",extra-cc:,smtp-to-addrs:"
-            getopt_longopts+=",log-level:,optimize:,heavy-asserts,mute-updates"
+            getopt_longopts+=",log-level:,optimize:,cache-bucket-name:,heavy-asserts,mute-updates"
```
spbnick, line 2275: We don't need this here, since `cloud` is not taking the option.

octonawish-akcodes: okay

cloud
octonawish-akcodes: done

cloud
octonawish-akcodes: Yeah it just take it from the environment of that instance after I exported it.

cloud
octonawish-akcodes: okay

main.py
octonawish-akcodes: okay

cloud
octonawish-akcodes: okay

cloud
octonawish-akcodes: done

cloud
```diff
@@ -2386,6 +2394,7 @@ function execute() {
                     --smtp-mocked="$smtp_mocked" \
                     --test="$test" \
                     --log-level="$log_level" \
+                    --cache-bucket-name="$cache_bucket_name" \
```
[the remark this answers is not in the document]
octonawish-akcodes: okay

cloud
```diff
@@ -2315,6 +2322,7 @@ function execute() {
             --psql-password-file)
                 password_set_file psql_superuser "$2"; shift 2;;
             --log-level) log_level="$2"; shift 2;;
+            --cache-bucket-name) cache_bucket_name="$2"; shift 2;;
```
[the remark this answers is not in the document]
octonawish-akcodes: okay

main.py
```diff
@@ -71,6 +71,8 @@
...
+CACHE_BUCKET_NAME = os.environ.get("KCIDB_CACHE_BUCKET_NAME", None)
```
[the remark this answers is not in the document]
octonawish-akcodes: removed the None condition

cloud
spbnick: We still need the argument in the function documentation, as @tales-aparecida hinted:
> This should show up in "Args:", don't it?

spbnick (changes requested):
LGTM, except the documentation comment inline. Thank you, @octonawish-akcodes!

spbnick (approved):
LGTM! Thank you, Abhishek :smiley:

PR #442, opened 2023-07-26
Add check for bucket existence before creation

octonawish-akcodes:
Added a check for bucket before creation.

Need reviews @spbnick

### DIFF
--- a/cloud
+++ b/cloud
@@ -1257,8 +1257,12 @@ function pubsub_withdraw() {
 function storage_deploy() {
     declare -r project="$1"; shift
     declare -r bucket="$1"; shift
-    TMPDIR="$TMPDIR_ORIG" gsutil -q mb -p "$project" -c STANDARD \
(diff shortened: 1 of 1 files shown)


spbnick:
We don't need to do this in the Python client, but when deploying the bucket in the `cloud` script, instead.

spbnick (changes requested):
LGTM, except please add `TMPDIR="$TMPDIR_ORIG" ` to both of the commands you added, as I've requested on our call today. Otherwise they can produce the error we saw before. Thank you!

octonawish-akcodes:
Done!

spbnick:
LGTM, but please fix the commit subject. We're checking for "bucket" existence, not storage. And we don't do that in the cache client.

octonawish-akcodes:
> LGTM, but please fix the commit subject. We're checking for "bucket" existence, not storage. And we don't do that in the cache client.

Looks good now?

spbnick (approved):

PR #443, opened 2023-07-26
Add empty method to cache client

octonawish-akcodes:
Added empty method to cache client to delete the contents of the bucket, to be used in test scenarios.

Need reviews @spbnick


kcidb/cache/__init__.py
```diff
@@ -157,3 +157,11 @@ def fetch(self, url):
         if blob.exists():
             return blob.download_as_bytes()
         return None
+
+    def empty(self):
+        """Empty the cache (remove all contents)."""
+        bucket = self.client.bucket(self.bucket_name)
+        blobs = bucket.list_blobs()
+
+        for blob in blobs:
```
spbnick, line 166: Looks good. Probably not efficient for large buckets, but should do alright for our usecase. Could you merge this and the previous line, though? I.e. do not assign the result of `list_blobs()` to a variable and use it directly in the loop? It's kinda trivial, and we could save some code lines this way.

octonawish-akcodes: Done, Have a look now?

spbnick (dismissed):
Looks good, thanks!

spbnick (approved):

PR #446, opened 2023-07-27
Add caching url test

octonawish-akcodes:
Need reviews @spbnick

### DIFF
--- a/test_main.py
+++ b/test_main.py
@@ -48,16 +48,16 @@ def test_url_caching(empty_deployment):
         topic
     )
     publisher.publish([
-        "https://github.com/kernelci/kcidb/blob/main/test_kcidb.py",
-        "https://github.com/kernelci/kcidb/blob/main/get-url-stats"
+        "https://raw.githubusercontent.com/kernelci/kcidb/main/test_kcidb.py",
+        "https://raw.githubusercontent.com/kernelci/kcidb/main/conftest.py"
     ])
 
     cache = kcidb.cache.Client(os.environ["KCIDB_CACHE_BUCKET_NAME"])
 
     for i in range(12):
         # Check if the URLs are in the cache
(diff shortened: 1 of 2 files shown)


test_main.py
```diff
@@ -37,3 +39,31 @@ def test_import():
...
         os.environ.update(orig_env)
+
+def test_url_caching(empty_deployment):
+    publisher = kcidb.mq.URLListPublisher(
+        os.environ["PROJECT"],
+        os.environ["KCIDB_UPDATED_URLS_TOPIC"]
+    )
+    publisher.publish([
+        "https://github.com/kernelci/kcidb/blob/main/test_kcidb.py",
+        "https://github.com/kernelci/kcidb/blob/main/get-url-stats"
```
spbnick, line 50: Use the URLs of the "raw" files. We would get a smaller and faster response, and they would cost GitHub much less.
E.g. the first one should be [https://raw.githubusercontent.com/kernelci/kcidb/main/test_kcidb.py](https://raw.githubusercontent.com/kernelci/kcidb/main/test_kcidb.py)

test_main.py
```diff
@@ -37,3 +39,31 @@ def test_import():
     finally:
         os.environ.clear()
         os.environ.update(orig_env)
+
+def test_url_caching(empty_deployment):
+    publisher = kcidb.mq.URLListPublisher(
+        os.environ["PROJECT"],
```
spbnick, line 45: Uh, this one should be `GCP_PROJECT`, shouldn't it?

spbnick (changes requested):
Looks pretty good! Go ahead and make it work now!

Please separate the addition of the `URLListPublisher` into its own commit, though.

test_main.py
octonawish-akcodes: Yeah it was just a demo layout to show u. I did changed it.

test_main.py
octonawish-akcodes: got it

PR #447, opened 2023-07-27
cloud: Add cache_redirector function

octonawish-akcodes:
Added cache redirection server function.

Need reviews @spbnick

### DIFF
--- a/main.py
+++ b/main.py
@@ -6,6 +6,8 @@ import base64
 import datetime
 import logging
 import smtplib
+from urllib.parse import unquote
+import functions_framework
 import kcidb
 
 # Name of the Google Cloud project we're deployed in
@@ -384,3 +386,48 @@ def kcidb_cache_urls(event, context):
     # Cache each URL
     for url in pubsub_message.splitlines():
         cache.store(url)
+
+
+@functions_framework.http
+def kcidb_cache_redirect(request):
+    """
+    Handle the cache redirection for incoming HTTP GET requests.
+
+    This function takes an HTTP request and processes it for
+    cache redirection. If the request is a GET request,
+    it extracts the URL from the request, checks if the URL exists
+    in the cache, and performs a redirect if necessary.
+
+    Args:
+        request (object): The HTTP request object.
+
+    Returns:
+        tuple: A tuple containing the response body, status code,
+        and headers epresenting the redirect response.
+    """
+    if request.method == 'GET':
+        url_to_fetch = unquote(request.query_string.decode("ascii"))
+        LOGGER.debug("URL %s", url_to_fetch)
+
+        if not url_to_fetch:
+            # If the URL is empty, return a 400 (Bad Request) error
+            response_body = "Put a valid URL to query from \
+                the caching system."
+            return (response_body, 400, {})
+
+        # Check if the URL is in the cache
+        cache_client = get_cache_client()
+        cache = cache_client.map(url_to_fetch)
+        if cache:
+            LOGGER.debug("Redirecting to the cache at %s", cache)
+            # Redirect to the cached URL if it exists
+            return ("", 302, {"Location": cache})
+
+        # If the URL is not in the cache or not provided,
+        # redirect to the original URL
+        LOGGER.debug("Redirecting to the origin at %s", url_to_fetch)
+        return ("", 302, {"Location": url_to_fetch})
+
+    # If the request method is not GET, return 405 (Method Not Allowed) error
+    response_body = "Method not allowed."
+    return (response_body, 405, {'Allow': 'GET'})
--- a/cloud
+++ b/cloud
@@ -1474,6 +1474,17 @@ function cloud_functions_deploy() {
                           --max-instances=1 \
                           --timeout 540
 
+    cloud_function_deploy "$project" "${prefix}cache_redirect" \
+                          "$source" \
+                          --entry-point kcidb_cache_redirect \
+                          --env-vars-file "$env_yaml_file" \
+                          --runtime python37 \
+                          --trigger-http \
+                          --allow-unauthenticated \
+                          --memory 256MB \
+                          --max-instances=1 \
+                          --timeout 540
+
     cloud_function_deploy "$project" "${prefix}load_queue" \
                           "$source" \
                           --entry-point kcidb_load_queue \
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,7 @@ google-cloud-firestore
 google-cloud-storage
 google-cloud-secret-manager
 psycopg2
+functions-framework
 importlib_metadata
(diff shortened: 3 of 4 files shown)


main.py
```diff
@@ -384,3 +386,41 @@ def kcidb_cache_urls(event, context):
     # Cache each URL
     for url in pubsub_message.splitlines():
         cache.store(url)
+
+
+def kcidb_cache_redirector(request):
```
spbnick, line 391: This is a regular function, so let's use the convention of using a verb for function name. I.e. use "redirect" instead of "redirector".

main.py
```diff
@@ -384,3 +386,41 @@ def kcidb_cache_urls(event, context):
...
+    redirect to the publicly accessible cached URL.
+    If the URL is not in the cache or not provided, the function responds
+    with a 302 redirect to the original URL provided in the query string.
+
+    Args:
+        request (flask.Request): The HTTP request object.
+
+    Returns:
+        flask.Response: The HTTP response object with a redirect status code.
+    """
```
spbnick, line 407: Great docstring :ok_hand: :sparkles: 

However, please rewrite the first line in imperative mood.

main.py
```diff
@@ -384,3 +386,41 @@ def kcidb_cache_urls(event, context):
...
+
+            if cache_blob:
+                # Redirect to the cached URL if it exists
+                return redirect(cache_blob.public_url, code=302)
+
+        # If the URL is not in the cache or not provided,
+        # redirect to the original URL
+        return redirect(url, code=302)
+
+    return None
```
spbnick, line 426: What does it mean to return `None` from this function? What would it do?

main.py
```diff
@@ -384,3 +386,41 @@ def kcidb_cache_urls(event, context):
...
+            cache_blob = storage_client.bucket(
+                CACHE_BUCKET_NAME).get_blob(url)
+
+            if cache_blob:
+                # Redirect to the cached URL if it exists
+                return redirect(cache_blob.public_url, code=302)
+
+        # If the URL is not in the cache or not provided,
+        # redirect to the original URL
+        return redirect(url, code=302)
```
spbnick, line 424: Looks good as a prototype. Next please switch to the cache client we developed and its URL mapping logic.

main.py
```diff
@@ -384,3 +386,41 @@ def kcidb_cache_urls(event, context):
...
+    if request.method == 'GET':
+        url = next(iter(request.args), None)
+        if url:
+            # Check if the URL is in the cache
+            cache_blob = storage_client.bucket(
+                CACHE_BUCKET_NAME).get_blob(url)
+
+            if cache_blob:
+                # Redirect to the cached URL if it exists
+                return redirect(cache_blob.public_url, code=302)
```
spbnick, line 420: Hmm, do we *have* to use Flask here? What does Google Cloud documentation say about this?

spbnick (changes requested):

main.py
octonawish-akcodes: Okay

main.py
octonawish-akcodes: Okay

main.py
octonawish-akcodes: I just add it as a common practice if it didn't get the `GET` then return None

spbnick: E.g. what kind of response the browser will get?

main.py
```diff
@@ -384,3 +385,57 @@ def kcidb_cache_urls(event, context):
     # Cache each URL
     for url in pubsub_message.splitlines():
         cache.store(url)
+
+
+def kcidb_cache_redirect(request):
```
spbnick, line 390: Please add the decorator Google Cloud Functions doc recommends here.

main.py
```diff
@@ -384,3 +385,57 @@ def kcidb_cache_urls(event, context):
...
+            and headers representing the redirect response.
+        """
+        headers = {
+            'Location': url,
+        }
+        return ('', 302, headers)
+
+    if request.method == 'GET':
+        full_url = unquote(request.url)
+        url_to_fetch = urlparse(full_url).query
```
spbnick, line 425: The above two lines should be just this:
```python
url_to_fetch = unquote(request.query_string.decode("ascii"))
```

main.py
```diff
@@ -384,3 +385,57 @@ def kcidb_cache_urls(event, context):
...
+            # Check if the URL is in the cache
+            cache_client = kcidb.cache.Client(
+                CACHE_BUCKET_NAME,
+                5 * 1024 * 1024
+            )
+            cache = cache_client.map(url_to_fetch)
+            if cache:
+                LOGGER.debug("Redirecting to the cache at %s", cache)
+                # Redirect to the cached URL if it exists
+                return redirect_response(cache)
```
spbnick, line 436: Let's not create a dedicated function for redirections just for these two calls. Just do it directly:
```python
return ("", 302, {"Location": cache})
```

spbnick (changes requested):
Please update the commit message to match your change.

spbnick (approved):
After looking at it for a while, let's not change anything else, for now. So, it's good to go in.

However, have you tested that it works?

main.py
```diff
@@ -384,3 +386,40 @@ def kcidb_cache_urls(event, context):
...
+        and headers epresenting the redirect response.
+    """
+    if request.method == 'GET':
+        url_to_fetch = unquote(request.query_string.decode("ascii"))
+        if url_to_fetch:
+            # Check if the URL is in the cache
+            cache_client = kcidb.cache.Client(
+                CACHE_BUCKET_NAME,
+                5 * 1024 * 1024
+            )
```
spbnick, line 415: Oh, actually, please use the function we added for getting the cache client.

main.py
```diff
@@ -384,3 +386,48 @@ def kcidb_cache_urls(event, context):
...
+            return ("", 302, {"Location": cache})
+
+        # If the URL is not in the cache or not provided,
+        # redirect to the original URL
+        LOGGER.debug("Redirecting to the origin at %s", url_to_fetch)
+        return ("", 302, {"Location": url_to_fetch})
+
+    # If the request method is not GET, return 405 (Method Not Allowed) error
+    response_body = "Method not allowed."
+    return (response_body, 405, {})
```
tales-aparecida, line 433: Please, find a way to include the header "allow", it might look something like this:
```suggestion
    return (response_body, 405, {'Allow': 'GET'})
```
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Allow

spbnick:
LGTM, thank you, Abhishek.

However, please squash all commits into one ("cloud: Add kcidb_cache_redirect function"), as we don't need them without it.

octonawish-akcodes:
@spbnick Done

spbnick (approved):

PR #450, opened 2023-08-02
reformat public url return value

octonawish-akcodes:
reformat public url return value in `_format_public_url` of `cache` client. This will fix the white spaces that were leading in the redirection URL generated by `_format_public_url`


spbnick (approved):
LGTM! Thanks, Abhishek!

PR #453, opened 2023-08-09
cloud: Restructure and documented cache bucket

octonawish-akcodes:
Restructure and documented cache bucket, also added the function deployment region variable.


spbnick (approved):
LGTM! Thanks, Abhishek!

PR #456, opened 2023-08-16
cloud: Wait for PostgreSQL proxy to become ready

spbnick:
Wait for the cloud_sql_proxy to become ready before running the command that's supposed to use it. This prevents command failures in times of higher load, when the proxy takes a bit to start, authenticate, and connect.

### DIFF
--- a/cloud
+++ b/cloud
@@ -321,6 +321,36 @@ function psql_proxy_session() {
     chmod 0600 "$PSQL_PROXY_PGPASS"
     password_get_pgpass psql_superuser postgres >| "$PSQL_PROXY_PGPASS"
 
+    # Wait for the proxy to become ready
+    declare max_checks=10
+    declare delay=3
+    declare checks
+    declare output=""
+    declare ready="false"
+    for ((checks=0; checks < max_checks; checks++)); do
+        if output+=$(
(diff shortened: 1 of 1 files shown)


cloud
```diff
@@ -321,6 +321,36 @@ function psql_proxy_session() {
...
+    # Check if the wait was successful
+    if ! "$ready"; then
+        echo "PostgreSQL proxy ${proxy@Q} is not ready " \
+             "after $((checks * delay)) seconds." >&2
+        if [ -n "$output" ]; then
+            echo "pg_isready output:" >&2
+            echo "$output" >&2
+        fi
+    fi
+    "$ready"
```
tales-aparecida, line 352: Is this line intentional?

spbnick: It's a bit tricky, yes. It's trying to trigger an `errexit` here, so that the trap triggers at this spot. It might be unnecessary, but I'm just trying to follow the convention we have in `cloud` so far - most things abort because something returned non-zero.

tales-aparecida: Oh, I see! So I was reading it right, just forgot to take `set -eEuCo pipefail` into consideration. Now I see that you call `false` in other places as well

PR #462, opened 2023-08-24
Update in increase max instances and reduce timeout for kcidb_cache_redirect

octonawish-akcodes:
Increased max instances and reduced timeout for the `kcidb_cache_redirect` cloud function.

Need reviews @spbnick

### DIFF
--- a/cloud
+++ b/cloud
@@ -1491,8 +1491,8 @@ function cloud_functions_deploy() {
                           --trigger-http \
                           --allow-unauthenticated \
                           --memory 256MB \
-                          --max-instances=1 \
-                          --timeout 540
+                          --max-instances=16 \
+                          --timeout 30
 
     cloud_function_deploy "$project" "${prefix}load_queue" \
                           "$source" \


spbnick (changes requested):
The change itself looks good, thank you. The commit message could be a bit better. First, please use imperative case, i.e. write "Increase" and "reduce" - it's easier to reason about changes phrased this way. Then it's good to make the subject a little shorter, and put any details into the commit body instead. So, something like this:
```
cloud: Adjust kcidb_cache_redirect to better handle production

Increase maximum instances for kcidb_cache_redirect Cloud Function and reduce
its timeout to perform better in production, which is going to be enabled soon.
```

octonawish-akcodes:
> The change itself looks good, thank you. The commit message could be a bit better. First, please use imperative case, i.e. write "Increase" and "reduce" - it's easier to reason about changes phrased this way. Then it's good to make the subject a little shorter, and put any details into the commit body instead. So, something like this:
> 
> ```
> cloud: Adjust kcidb_cache_redirect to better handle production
> 
> Increase maximum instances for kcidb_cache_redirect Cloud Function and reduce
> its timeout to perform better in production, which is going to be enabled soon.
> ```
Check now?

spbnick (approved):

spbnick:
Looks great, thanks! Let's wait for CI.

PR #463, opened 2023-08-24
client: Expose future based submissions

broonie:
Currently all the KCIDB submission APIs are synchronous, meaning we have
network latency in the call which isn't ideal for large numbers of small
submissions. The underlying BigQuery PubSub APIs offer an API based on
futures which should hopefully allow us to do better here, allowing the
client to submit data and then later check that it has actually been sent
so that the PubSub library can batch I/O or do it asynchronously.

This is already wired up at the publisher level so just add a top level
client method which exposes the future based API. We may want to also
expose something of the flow control parameters via the PublishFlowControl
passed when creating the client, though the current defaults (1000 messages
and 10Mb in flight) look reasonable.

### DIFF
--- a/kcidb/__init__.py
+++ b/kcidb/__init__.py
@@ -83,6 +83,30 @@ class Client:
             raise NotImplementedError
         return self.mq_publisher.publish(data)
 
+    def submit_future(self, data):
+        """
+        Submit reports without blocking for the interaction with the
+        database.
+
+        Args:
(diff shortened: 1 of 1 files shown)


spbnick:
> P.S. If you want your PR tested in production in the future, create it from a branch in KCIDB itself. Then it will get access to the required tokens.

By "in production" I meant "in deployment", of course :see_no_evil:

PR #464, opened 2023-08-24
submit: Use correct function for future submissions

broonie:
The previous function had not been tested due to difficulties with test
setups, fix the name of the function for future submission.

### DIFF
--- a/kcidb/__init__.py
+++ b/kcidb/__init__.py
@@ -83,7 +83,7 @@ class Client:
             raise NotImplementedError
         return self.mq_publisher.publish(data)
 
-    def submit_future(self, data):
+    def future_submit(self, data):
         """
         Submit reports without blocking for the interaction with the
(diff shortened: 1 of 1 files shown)


spbnick:
Ha-ha, I missed that too 😁
But could you maybe rename `submit_future` to `future_submit` too, to keep things consistent?

broonie:
I can do that, but note that the iterator function is called submit_iter() so there'd be some inconsistency at the kcidb top level. Either way is fine.

spbnick:
> I can do that, but note that the iterator function is called submit_iter() so there'd be some inconsistency at the kcidb top level.

Yeah, submit_iter is kinda special, as it's not returning a future, so I think it's OK. Thanks!

LGTM, but let's wait for CI.

spbnick (approved):

PR #467, opened 2023-08-25
cache: Add debug logging to cache client's store method

octonawish-akcodes:
Need reviews @spbnick

### DIFF
--- a/kcidb/cache/__init__.py
+++ b/kcidb/cache/__init__.py
@@ -57,6 +57,8 @@ class Client:
         blob = self.client.bucket(self.bucket_name).blob(object_name)
 
         if blob.exists():
+            LOGGER.debug("URL '%s' already exists, "
+                         "not caching.", url)
             return
 
         try:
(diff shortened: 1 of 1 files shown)


spbnick:
Please request reviews only after CI has passed :+1:

kcidb/cache/__init__.py
```diff
@@ -52,11 +52,13 @@ def store(self, url):
 
         # Cache every 4096th URL only for the trial period
         if not object_name.endswith("000"):
+            LOGGER.debug("Skipping URL '%s', hash not matched.", url)
```
spbnick, line 55: Please remove this one, as this would generate *waaaay* too many log messages (as I mentioned on our call).

kcidb/cache/__init__.py
```diff
@@ -52,11 +52,13 @@ def store(self, url):
...
             return
 
         blob = self.client.bucket(self.bucket_name).blob(object_name)
 
         if blob.exists():
+            LOGGER.debug("URL '%s' already exists in cache.", url)
```
spbnick, line 61: Please add `, not caching` on the end, so it's clear what's the outcome. Here, and in the following two messages.

spbnick (changes requested):
Looks good overall, but asking for a few tweaks. Thank you, Abhishek!

Also, since you already have the `cache: ` tag in your commit subject, you don't need the second `cache` in it. So it could be just `cache: Add debug logging to client's store method`. Otherwise a perfect commit subject :+1: !

kcidb/cache/__init__.py
octonawish-akcodes: Okay

spbnick (approved):
Looks great! Thank you, Abhishek!

spbnick:
Deployment testing failed, because GitHub had some problem. It should likely be OK, and we'll run more tests before deploying, so let's merge this.

PR #469, opened 2023-08-25
cache: Remove one 0 from hash name check to cache more URLs

octonawish-akcodes:


kcidb/cache/__init__.py
```diff
@@ -51,7 +51,7 @@ def store(self, url):
         object_name = self._format_object_name(url)
 
         # Cache every 4096th URL only for the trial period
```
spbnick, line 53: Looks good, but please update the comment as well.

spbnick: Yep!

spbnick (changes requested):

kcidb/cache/__init__.py
octonawish-akcodes: done

octonawish-akcodes:
done

spbnick (approved):
LGTM! Thank you, Abhishek. Let's see how CI behaves. Those changed hash requirements could have affected tests.

PR #480, opened 2023-11-30
Update firestore filtering

### LINKED ISSUE #431: Update firestore filtering
Our tests have started producing a deprecation warning from the Firestore API:
```
kcidb/test_monitor.py::test_email_generated[empty_deployment]
  /opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/site-packages/google/cloud/firestore_v1/base_collection.py:290: UserWarning: Detected filter using positional arguments. Prefer using the 'filter' keyword argument instead.
    return query.where(field_path, op_string, value)
```

Update the interface we're using so the warning disappears.

danghai:
Fix #431

### DIFF
--- a/kcidb/monitor/spool/__init__.py
+++ b/kcidb/monitor/spool/__init__.py
@@ -12,6 +12,7 @@ import datetime
 import email
 import email.policy
 from google.cloud import firestore
+from google.cloud.firestore_v1.base_query import FieldFilter
 from kcidb.misc import ArgumentParser, iso_timestamp, log_and_print_excepthook
 from kcidb.monitor.misc import is_valid_firestore_id
 from kcidb.monitor.output import Notification
@@ -238,8 +239,9 @@ class Client:
             isinstance(until, datetime.datetime) and until.tzinfo
         if until is None:
             until = datetime.datetime.now(datetime.timezone.utc)
+        field_filter = FieldFilter("created_at", "<=", until)
         for snapshot in \
-            self._get_coll().where("created_at", "<=", until). \
+            self._get_coll().where(filter=field_filter). \
                 select([]).stream():
             snapshot.reference.delete()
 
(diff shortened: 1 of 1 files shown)


danghai:
@spbnick , I am not sure why check_deployment is still skipped?

danghai:
Do you have the command that I can run the test locally? I mean this one:
```
  File "/home/runner/work/kcidb/kcidb/kcidb/test_monitor.py", line 841, in test_email_generated
    assert len(remaining_obj_types) == 0, \
AssertionError: Missing ***'build', 'test', 'checkout', 'revision'*** emails for <class 'kcidb_io.schema.v03_00.Version'>
assert 4 == 0
 +  where 4 = len(***'build', 'checkout', 'revision', 'test'***)
```

spbnick:
> Do you have the command that I can run the test locally? I mean this one:

Most of the tests you can run with just `pytest`. However, some of the tests require access to a deployment, which requires you creating a Google Cloud project (you can have six months of free trial if you provide a credit card, IIRC).

The general workflow looks something like this:
```bash
./cloud deploy "<gcp_project>" "<namespace>" 1 -v --smtp-mocked --test --heavy-asserts # Create a deployment in your project
./cloud shell "<gcp_project>" "<namespace>" 1 --smtp-mocked --test --heavy-asserts # Start a shell with environment variables pointing to the deployment
export KCIDB_DEPLOYMENT="This deployment is empty" # Only do this with *actual* empty deployments, needed for the monitor test
pytest # Run the tests, repeat as many times as necessary, with whatever options you need
exit # Exit the shell
./cloud withdraw "<gcp_project>" "<namespace>" 1 -v --smtp-mocked --test # Withdraw (destroy) the deployment (leaving some shared objects behind)
```

Where `<gcp_project>` is the name of the Google Cloud Project you have created. And `<namespace>` is a "namespace" - a prefix for the names of (almost) all created objects, that lets you distinguish deployments within a single project. The latter can be an empty string.

There's of course much more to it, but you can start with these, and modify, after reading the corresponding `--help` output and the docs, where there are any.

spbnick:
> Oh it passes now after re-run :D. @spbnick could you review my PR? It resolves the warning.

Ah great! However, in general the test shouldn't be flaky and this warrants investigation. Don't rerun it if you see it again, let me see what's going on first, please.

The change looks good, but the commit message could be better. As such it reads like "Change firestore filtering", and doesn't say much. Could you make it clearer? Even better if it has a reference to the Google documentation covering this change in the API.

Thank you :bowing_man:

spbnick (approved):
LGTM! Thank you, Hai!

Issue #242, opened 2021-11-23
Add tests for kcidb-ingest


spbnick:
Add tests for the kcidb-ingest tool. Use `test_notify_main()` in `test_kcidb.py` as a reference.

Mbaoma:
Hi, i'd like to tinker around with this for a few days and see if I can do anything reasonable. cool? @spbnick

spbnick:
Absolutely! The issue as is is more of a reminder, and is really bare-bones, sorry, didn't have the time to fill it in yet, for this Outreachy round

You can start with checking out the `kcidb` package, then installing it for local development, e.g. with `pip3 install --user -e '.[dev]'` (see more instruction in docs), and then playing with the installed `kcidb-ingest` tool.

E.g. you can take the attached [test.json.gz](https://github.com/kernelci/kcidb/files/10901810/test.json.gz), unpack it, and feed it to the tool: `kcidb-ingest < test.json`. It should output a bunch of email messages one after another.

The test should try to feed it various options and inputs and verify that it produces appropriate outputs. You can try to figure out which those would be, but if not, I can post a description of what could be done tomorrow.

Have fun!

Mbaoma:
Hi @spbnick I try to setup kcidb on my mac, but keep getting a python wheel error.
are there specific steps to set this up on a mac?

Mbaoma:
<img width="718" alt="image" src="https://user-images.githubusercontent.com/49791498/224381403-19235bac-a1c9-424e-b076-3a218c57ad74.png">

spbnick:
Ah, it could be a big pain in the rear to get this working on a Mac, although that's possible. You'd need to install a bunch of packages from Brew, and a toolchain or something like that, and I can't remember exactly which ones you need and how to install them. I would recommend you install a Linux distro (e.g. Ubuntu) in a VM, and work from there. If you'd prefer not to do that, you'll need to wait for me to find a couple hours to help you with this, next week.

spbnick:
Good luck!

spbnick:
If you can't find people to help you, try asking ChatGPT :robot:

Mbaoma:
sure thing!

spbnick:
Removing @Mbaoma from the assignees, as Outreachy's initial contribution period is now over.

PR #481, opened 2023-12-03
test_kcidb: Add tests for kcidb-ingest

danghai:
Fix: #242
Output:
image

### DIFF
--- a/test_kcidb.py
+++ b/test_kcidb.py
@@ -429,3 +429,144 @@ def test_notify_main():
                               "Subject: Test build: .*\x00"
                               "Subject: Test test: .*\x00"
                               "Subject: Test test: .*\x00")
+
+
+def test_ingest_main():
+    """Check kcidb-ingest works"""
+    assert_executes('', "kcidb.ingest_main")
+    assert_executes('{', "kcidb.ingest_main",
+                    status=1, stderr_re=".*JSONParseError.*")
+    assert_executes('{}', "kcidb.ingest_main",
+                    status=1, stderr_re=".*ValidationError.*")
+    empty = json.dumps(dict(version=dict(major=SCHEMA.major,
+                                         minor=SCHEMA.minor)))
+    assert_executes(empty, "kcidb.ingest_main")
+    assert_executes(empty + empty, "kcidb.ingest_main")
+    git_commit_hash1 = "4ff6a2469104218a044ff595a0c1eb469ca7ea01"
+    git_commit_hash2 = "fe3fc1bc47d6333d7d06bc530c6e0c1044bab536"
+
+    one_of_everything = json.dumps(dict(
+        version=dict(major=SCHEMA.major, minor=SCHEMA.minor),
+        checkouts=[
+            dict(id="test:checkout:1",
+                 git_commit_hash=git_commit_hash1,
+                 patchset_hash="",
+                 origin="test")
+        ],
+        builds=[
+            dict(id="test:build:1",
+                 origin="test",
+                 checkout_id="test:checkout:1",
+                 valid=False)
+        ],
+        tests=[
+            dict(id="test:test:1",
+                 origin="test",
+                 build_id="test:build:1",
+                 waived=False,
+                 status="PASS")
+        ],
+        issues=[
+            dict(id="test:issue:1",
+                 version=1,
+                 origin="test",
+                 report_url="https://test.com/bug/1",
+                 report_subject="Bug in kernel",
+                 culprit=dict(code=True)),
+        ],
+        incidents=[
+            dict(id="test:incident:1",
+                 issue_id="test:issue:1",
+                 issue_version=1,
+                 origin="test",
+                 test_id="test:test:1",
+                 present=True),
+        ],
+    ))
+    assert_executes(one_of_everything, "kcidb.ingest_main",
+                    stdout_re="Subject: Test revision: .*\x00"
+                              "Subject: Test checkout: .*\x00"
+                              "Subject: Test build: .*\x00"
+                              "Subject: Test test: .*\x00"
+                              "Subject: Test bug: .*\x00"
+                              "Subject: Test issue: .*\x00"
+                              "Subject: Test incident: .*\x00")
+
+    two_of_everything = json.dumps(dict(
+        version=dict(major=SCHEMA.major, minor=SCHEMA.minor),
+        checkouts=[
+            dict(id="test:checkout:1",
+                 git_commit_hash=git_commit_hash1,
+                 patchset_hash="",
+                 origin="test"),
+            dict(id="test:checkout:2",
+                 git_commit_hash=git_commit_hash2,
+                 patchset_hash="",
+                 origin="test"),
+        ],
+        builds=[
+            dict(id="test:build:1",
+                 origin="test",
+                 checkout_id="test:checkout:1",
+                 valid=False),
+            dict(id="test:build:2",
+                 origin="test",
+                 checkout_id="test:checkout:2",
+                 valid=False),
+        ],
+        tests=[
+            dict(id="test:test:1",
+                 origin="test",
+                 build_id="test:build:1",
+                 waived=False,
+                 status="PASS"),
+            dict(id="test:test:2",
+                 origin="test",
+                 build_id="test:build:2",
+                 waived=False,
+                 status="FAIL"),
+        ],
+        issues=[
+            dict(id="test:issue:1",
+                 version=1,
+                 origin="test",
+                 report_url="https://test.com/bug/1",
+                 report_subject="Bug in kernel",
+                 culprit=dict(code=True)),
+            dict(id="test:issue:2",
+                 version=1,
+                 origin="test",
+                 report_url="https://non-test.com/bug/1",
+                 report_subject="Another bug in kernel",
+                 culprit=dict(tool=True))
+        ],
+        incidents=[
+            dict(id="test:incident:1",
+                 issue_id="test:issue:1",
+                 issue_version=1,
+                 origin="test",
+                 test_id="test:test:1",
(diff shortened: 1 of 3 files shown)


danghai:
@spbnick The tool `kcidb-ingest` or `kcidb-notify` get hang when I do not pass any arguments?

test_kcidb.py
```diff
@@ -429,3 +429,93 @@ def test_notify_main():
...
+                 valid=False)
+        ],
+        tests=[
+            dict(id="test:test:1",
+                 origin="test",
+                 build_id="test:build:1",
+                 waived=False,
+                 status="PASS")
+        ]
+    ))
```
spbnick, line 469: This is not really "everything" despite what the variable name says. We also have "issues" and "incidents" now, and it would be good to add those too. Look around the tests for sample data.

spbnick (changes requested):
> The tool kcidb-ingest or kcidb-notify get hang when I do not pass any arguments?

Not really, they simply wait for (JSON) input on standard input.

The change itself looks good, but it would be nice to add more sample data, as I mention in the inline comment. Thank you!

test_kcidb.py
danghai: @spbnick where I can find the tests for sample data?

spbnick: A command like this would have got you started: `grep -r --include 'test_*.py' incident`

Here's a good one: https://github.com/kernelci/kcidb/blob/c2de2921297d315117acdc1a3fd6305c27245e72/kcidb/test_monitor.py#L196-L231

But see the schema docs to figure out what means what.

test_kcidb.py
```diff
@@ -429,3 +429,139 @@ def test_notify_main():
...
+
+    assert_executes(two_of_everything,
+                    "kcidb.ingest_main",
+                    stdout_re="Subject: Test revision: .*\x00"
+                              "Subject: Test revision: .*\x00"
+                              "Subject: Test checkout: .*\x00"
+                              "Subject: Test checkout: .*\x00"
+                              "Subject: Test build: .*\x00"
+                              "Subject: Test build: .*\x00"
+                              "Subject: Test issue: .*\x00"
```
danghai, line 566: @spbnick I get failure assertion when I have duplicate `Subject: Test issue: .*\x00`, and `Subject: Test incident: .*\x00`. It should be 2 for each one right?

danghai:
@spbnick I see the failure deployment. It seems it is not relate to my change?

test_kcidb.py
spbnick: What's the actual output you get from kcidb-ingest in this case?

spbnick:
> I see the failure deployment. It seems it is not relate to my change?

I see connection failures in the latest results, and those seems to be an infrastructure issue. Try just restarting the job in this case.

test_kcidb.py
danghai: image

spbnick: Sorry, screenshots are generally useless for lots of text, and these don't really include any relevant information. Can you upload the complete output of a test run to e.g. a [gist](https://gist.github.com/)?

danghai: @spbnick , yeah I upload gist here:
https://gist.github.com/danghai/20fc34fc815f8ed5e3f33d5177468192

spbnick:
Thanks for working on this Hai. I have taken a closer look at what's going on there and found a couple issues in the notification templates, please a couple problems with your test. I added a commit fixing the issues in the templates before yours. And another one fixing the test after yours. Please take a look, and squash the test fix into your commit, but please keep the template fix separate.

Please also review my changes, and see if you can find anything wrong with them.

One final thing: please add a blank line between the commit message subject (the first line) and the commit message body (everything else). This is the format most git tools expect and it messes up their output if the blank line is not there. GitHub tries to deal with it and hides the problem, but most other tools don't.

test_kcidb.py
spbnick: Thanks! In general, text pastes from the terminal are preferable to screenshots.

danghai:
Hi @spbnick , sorry for late response because of vacation. I try to rebase, and squash the last 2 commit. I get this result for log. I think it is wrong?
```
danghai@danghai-virtual-machine:~/kcidb$ git rebase -i c2de2921297d315117acdc1a3fd6305c27245e72
[detached HEAD 5eacfe5] templates: Handle missing values in issues and incidents
 Author: Nikolai Kondrashov <Nikolai.Kondrashov@redhat.com>
 Date: Fri Dec 8 17:27:11 2023 +0200
 3 files changed, 150 insertions(+), 4 deletions(-)
Successfully rebased and updated refs/heads/haidang/test_ingest.
danghai@danghai-virtual-machine:~/kcidb$ git log
commit 5eacfe5bce11af2b38d850c43c16e29a066648b2 (HEAD -> haidang/test_ingest)
Author: Nikolai Kondrashov <Nikolai.Kondrashov@redhat.com>
Date:   Fri Dec 8 17:27:11 2023 +0200

    templates: Handle missing values in issues and incidents
    
    Handle some missing values in issue and incident templates properly.
    
    The 'default' filter unfortunately handles 'undefined' only, so use a
    combination of the filter and the tests to handle both interim and final
    None values when accessing fields deep in a structure.
    
    test_kcidb: Add tests for kcidb-ingest
    
    Fix: #242
    Fix IDs of sample objects in test_ingest_main(), and make sure all
    notifications that should be generated are accounted for.

commit c2de2921297d315117acdc1a3fd6305c27245e72 (main, example)
Author: danghai <haihoangdang91@gmail.com>
Date:   Sun Dec 3 15:35:51 2023 -0500

    Update firestore to use FieldFilter.
    Fix: #431
    UserWarning in build log: Detected filter using positional arguments.
    Prefer using the 'filter' keyword argument instead
    The filter argument takes BaseFilter class which is an abstract class
    and it is implemented in FieldFilter. This class can be imported from
    google.cloud.firestore_v1.base_query
    Reference: https://github.com/googleapis/python-firestore/issues/705
               https://github.com/GoogleCloudPlatform/python-docs-samples/pull/10407

```
I do command:
git reset --hard <remote branch>
git rebase -i <commit at third from top>
I change 2 squash for 2 last commits. 
I combine 2 commit message for 2 last commits

spbnick:
Yeah, you squashed wrong commits, and you don't have the commit I pushed to your branch. You gotta do `git fetch origin` to update you're repo's info about the remote branch I updated, then do something like `git reset --hard origin/haidang/test_ingest` to make your branch to be exactly as the remote one, and go from there.

danghai:
I do the same as you recommend. However I mess up my PR. Could I create a new PR for it?
image
I need to squash commit `92f173e` and `b2dcdd8`, or `4014193` and `b2dcdd8`?

spbnick:
@danghai, you need to figure out how these things work. These skills are important for open-source contributors. Spend as much time as you need to slowly go through the steps, experiment, and learn.

Some facts to get you started (which might or might not be useful for you):

* A branch or a tag is a pointer to a directed acyclic graph (DAG) of commits.
* When you commit something you create a new commit node in that graph, with an edge pointing to the parent commit from it.
* All your branches in your local repo are pointers to various spots on that graph
* Same for the remote branches
* Remote and local graphs and branches don't have to be the same
* When you fetch from a remote repo, your repo its knowledge of remote branches, and that's all.
* You can use different methods to make your local branches to point to the same commits as the remote ones, but you don't have to.
* You can make branches point to different commits without actually committing anything.

Go and experiment with git, go through a couple tutorials, try to get a feel for it.

Your target in this PR is to keep commit [templates: Handle missing values in issues and incidents](https://github.com/kernelci/kcidb/pull/481/commits/40141939cd9a5e4fd3e053607f6ebe8f005644c9) separate, but squash [test_kcidb: Add tests for kcidb-ingest](https://github.com/kernelci/kcidb/pull/481/commits/b2dcdd8a05a5f4247415bac750681e97fd52124c) and [tests: Misc fixes to test_ingest_main](https://github.com/kernelci/kcidb/pull/481/commits/92f173eba092cc38ca41e6b176ef082af31d20e3) together, and pay attention to the resulting commit message, so it makes sense by itself and is not just two messages pasted together, and keeps to the standard format (subject on the first line, blank line, then the body).

My request is a standard request that open-source maintainers make for contributions, and I (or others) cannot keep giving you exact instructions of how to fulfill it every time.

danghai:
@spbnick Thank you. I squash them together. Is it good enough?

spbnick:
Thank you, Hai. The change itself looks good, but the commit message doesn't make much sense. You have to describe the change from the point of the parent commit. Let's take it piece by piece.

Your commit message subject says: `test_kcidb: Add tests for kcidb-ingest`. And that's indeed what it does. Except we don't generally use `test_kcidb: ` tag, it's just a little too specific. You can see that in the output of `git log --oneline test_kcidb.py`. Just `tests: ` would be enough here.

You have a blank line after the subject, great! Thanks :grin: 

Then the message body says: `Fix: #242`, which is fine by itself, and is definitely useful, but it's best to put these things at the end of the message, as reference. You don't start a paper with references, you put them at the end, right? Similarly, if you have a detailed description of the change in the commit message body, it's better to start reading that, and only then follow the references (issue numbers).

Finally, the message body says:
> Fix IDs of sample objects in test_ingest_main(), and make sure all notifications that should be generated are accounted for.

And that's not at all what the change actually does. There's no `test_ingest_main()` before this commit, so we cannot be fixing anything in it. Your subject says what the change does pretty succinctly, and you don't really need to say more, but if you want to say something about what the change is doing, describe what's **actually** happening in the diff of the commit the message belongs to, please.

danghai:
@spbnick Thank you. I update the change

spbnick (approved):
LGTM! Thank you, Hai!

PR #488, opened 2024-01-15
Break down cloud tool

### LINKED ISSUE #451: Break down cloud tool
The [`cloud` tool's source code](https://github.com/kernelci/kcidb/blob/main/cloud) is a single file 2312 lines long, which is shameful, and makes it hard to read, navigate, and maintain.

Break it down into smaller modules (`.sh` include files). Put modules into the `kcidb` directory. Group functions and variables by prefix, and put them into correspondingly-named files. E.g. all functions and variables with prefix `psql_`/`PSQL_` should go into a file named `kcidb/psql.sh`, and so on.

The original `cloud` file must add the directory it's located in in front of the `PATH` environment variable (following its syntax, i.e. using the `:` path separator), and then include the modules using the `.` (`source`) command. E.g. like this:
```bash
export PATH="$(dirname "$(realpath "$0")")/kcidb:$PATH"

. psql.sh
```

Each module file should start with a short description of its contents, and should have anti-recursive inclusion protection around its code and definitions. E.g. the `kcidb/psql.sh` file should have:
```bash
# PostgreSQL database deployment definitions and management

if [ -z "${_PSQL_SH+set}" ]; then
declare _PSQL_SH=

# Code and definitions go here

fi # _PSQL_SH
```

Various definitions reused by multiple files, that are too few for a separate module, should go into `kcidb/misc.sh`, but don't have to have a `misc_` prefix added to their names.

spbnick:
Fixes #451


tales-aparecida (approved):
Already merged, but in any case LGTM 😆

spbnick:
Thanks for taking a look, @tales-aparecida!

PR #511, opened 2024-03-20
postgresql: fix get_version() for older PostgreSQL versions

hardboprobot:
The bare 'RETURN' statement in get_version() is caught as a syntax error by PostgreSQL 12.x. Return the integer literal using a simple SQL statement instead.

**NOTE**: Not tested with other PostgreSQL versions.


spbnick (approved):
LGTM, thank you, Ricardo!

PR #516, opened 2024-04-03
doc/submitter_guide: prefix for `checkout_id`

JenySadadia:
As per the convention `checkout_id` value is expected to have `submitter:` prefix.
Update the example in the submitter guide accordingly.


spbnick (approved):
Oh, I see now, I missed that one! Thank you for the fix, Jeny!

PR #520, opened 2024-04-25
tests.yaml: add KernelCI-specific tests

JenySadadia:
Add KernelCI tests to test catalogue as per KCIDB norms.

### DIFF
--- a/tests.yaml
+++ b/tests.yaml
@@ -70,6 +70,25 @@ fwts:
 jcstress:
   title: The Java Concurrency Stress tests
   home: https://github.com/openjdk/jcstress
+kernelci_baseline:
+  title: Test plan for checking kernel errors using dmesg
+  home: https://github.com/kernelci/kernelci-core/blob/main/config/runtime/tests/baseline.jinja2
+kernelci_kver:
+  title: Kernel Version test
+  home: https://github.com/kernelci/kernelci-pipeline/blob/main/config/runtime/kver.jinja2
+  description: |
+    Checks kernel version using "git describe" and verify it with kernel
+    Makefile version.
+kernelci_sleep:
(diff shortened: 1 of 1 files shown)


spbnick:
This looks great, thank you, Jenny!

However, are there any better home pages for those tests, something human-oriented? Or at least could you replace the URLs pointing to the Jinja2 fiels with URLs pointing to the actual test code?

JenySadadia:
> However, are there any better home pages for those tests, something human-oriented?

Do you mean some kind of documentation? I am sorry, we don't have it yet.

> Or at least could you replace the URLs pointing to the Jinja2 files with URLs pointing to the actual test code?

The API uses these jinja2 templates and generates test code (python scripts that we don't store) on the fly and uses them to run tests.

spbnick:
> The API uses these jinja2 templates and generates test code (python scripts that we don't store) on the fly and uses them to run tests.

That's OK, even if complicated, but I assume the executed tests themselves have separate code, which actually tells more about what they do, and so would be more helpful for people e.g. investigating a failure, or just trying to find out about them?

JenySadadia:
> That's OK, even if complicated, but I assume the executed tests themselves have separate code, which actually tells more about what they do, and so would be more helpful for people e.g. investigating a failure, or just trying to find out about them?

As I mentioned we don't store the test python script. Jinja2 templates are all we have atm.

> and so would be more helpful for people e.g. investigating a failure, or just trying to find out about them?

Yes, I understand. Once the documentation for tests has been added, I replace the URLs with the doc links. But currently, we don't have anything apart from these templates.

spbnick (approved):
Got it. Thanks, Jenny!

PR #522, opened 2024-04-29
tests.yaml: rename `kernelci_tast` test

JenySadadia:
Rename `kernelci_tast` to `tast` as it is a known test suite from ChromeOS and not specific to KernelCI. Update test suite link as well.


spbnick (approved):
LGTM, thank you, Jeny!

PR #529, opened 2024-06-06
Support v4.4 I/O schema

spbnick:


spbnick:
@hardboprobot, here's the preliminary support for our draft v4.4 I/O schema. Deployment tests are still running, but it should be sufficient to create a database, and start submitting "transitions". We can make a development/test deployment in staging for us to play with. We can start with generating the transitions using SQL directly, and then see whether we can/need to implement the ORM support for it.

Please review!

spbnick:
This attempts to separate the "dummy" support for the schema from the actual database support. However, I haven't tested yet if it would work. Will do that later.

hardboprobot (approved):
There are so many things here I'm not familiar with that I can't really say anything other than if you're ok with the changes and they're tested and work, then it's fine by me. I like the plan, let me know when I can jump in to play with transitions and detections.

spbnick:
> There are so many things here I'm not familiar with that I can't really say anything other than if you're ok with the changes and they're tested and work, then it's fine by me. I like the plan, let me know when I can jump in to play with transitions and detections.

No problem! I just wanted you to take a look to get a feel for it, and tell me if you spotted anything weird.

Would you like to have a call and go over how to do a deployment, get a data sample, load it up, and so on? Or would you prefer I do all that for you and just give you the database credentials?

In any case once you've got a feel for the schema and tried implementing the analysis, we can get on with implementing it in KCIDB. As a hardcoded function, for example, for the start.

PR #530, opened 2024-06-06
tests.yaml: add `kernelci_watchdog_reset` test

JenySadadia:
Add KernelCI Watchdog reset test suite.


spbnick (approved):
LGTM! Although we need to point to actual tests eventually, not KernelCI Jinja2 templates that are hard to understand.

JenySadadia:
> LGTM! Although we need to point to actual tests eventually, not KernelCI Jinja2 templates that are hard to understand.

Yes, I'll update it when we have docs ready to point at.

PR #531, opened 2024-06-06
Add flowchart about the status ownership and mention "running" around the absent status

tales-aparecida:
CKI had this chart for a while in https://cki-project.org/docs/test-maintainers/status-meaning/ and it's been useful for explaining ownership!

### DIFF
--- a/doc/submitter_guide.md
+++ b/doc/submitter_guide.md
@@ -315,6 +315,65 @@ Finally, the last line in the table above (without a status) corresponds to
 the absent "status" field, and means that the test is only scheduled to
 execute, the system hasn't got to it yet, and so we have no data on anything.
 
+Another way to visualize the status values and the three layered testing stack,
+is to put them into an execution outcome chart:
+
+```mermaid
+block-beta
+    columns 5
+
+    B1("SKIP\ntest not applicable")
+    space
+    C1("DONE\nno test result required")
+    space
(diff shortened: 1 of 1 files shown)


tales-aparecida:
image

doc/submitter_guide.md
```diff
@@ -312,8 +312,8 @@ means that the harness worked alright, and either it, or the test itself,
 decided not to run this specific test, because it was inapplicable.
 
 Finally, the last line in the table above (without a status) corresponds to
-the absent "status" field, and means that the test is only scheduled to
-execute, the system hasn't got to it yet, and so we have no data on anything.
+the absent "status" field, and means that the test is either scheduled to
+execute, the system hasn't got to it yet, and so we have no data on anything, or it is still running.
```
spbnick, line 316: Perhaps:
```suggestion
the absent "status" field, and means that the test is either only scheduled, or is still executing,
and so we have no status data yet.
```

spbnick (changes requested):
Looks good to me mostly, thank you very much, Tales!

I left a couple comments inline, and do you think you could add a lead-in to that diagram, in a separate paragraph? Something like "Another way to organize status values is to put them into an execution outcome chart:"

spbnick:
Thanks a lot, Tales! Merging :+1:

Issue #532, opened 2024-06-19
Create an initial stable-rc report email in KCIDB


padovan:
We need to create a report for our ongoing work for of reporting stable-rc results with Greg KH.

The report will be sent to ourselves first. @crazoes is the main recipient.

Info given by @spbnick:

Here's a [sample](https://groups.io/g/kernelci-results-staging/message/68715?p=%2C%2C%2C20%2C0%2C0%[…]eated%2C%2C%2522Testing+done+for+%2522%2C20%2C2%2C60%2C106682364) of notifications we send to Mark. It's sent an hour after last update to the revision.

Here's the code of his subscription: https://github.com/kernelci/kcidb/blob/main/kcidb/monitor/subscriptions/mark_brown.py

Here's the template entry point for notification body for revisions: https://github.com/kernelci/kcidb/blob/main/kcidb/templates/revision_description.txt.j2 (edited)

JenySadadia:
Hello @crazoes 

Here is the sample report generated for `stable-rc` from recent submissions to KCIDB:
```
Subject: stable-rc report for linux-stable-rc.git:linux-6.6.y@v5.16-rc2-19-g383a44aec91c


Below is the summary of results Kernel CI database has recorded
for this revision so far. See complete and up-to-date report at:

    https://kcidb.kernelci.org/d/revision/revision?orgId=1&var-git_commit_hash=580e509ea1348fc97897cf4052be03c248be6ab6&var-patchset_hash=

OVERVIEW

        Builds: ❌ FAIL
         Tests: ✅ PASS

REVISION

    Status
        ✅ PASS
    Commit
        name: v5.16-rc2-19-g383a44aec91c
        hash: 580e509ea1348fc97897cf4052be03c248be6ab6
    Checked out from
        https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable-rc.git linux-6.6.y
    By
        broonie, maestro

BUILDS

    Status
             ❌ 1  ✅ 1
    Architectures
        arm  ❌ 1  ✅ 1
    Failures
        ❌ 1  arm multi_v7_defconfig
    By
        broonie, maestro

TESTS

    Status
        ✅ 1
    By
        maestro

See complete and up-to-date report at:

    https://kcidb.kernelci.org/d/revision/revision?orgId=1&var-git_commit_hash=580e509ea1348fc97897cf4052be03c248be6ab6&var-patchset_hash=

LEGEND

[...]

PR #537, opened 2024-07-01
Linux `stable-rc` subscription

JenySadadia:
Closes https://github.com/kernelci/kcidb/issues/532

Add a subscription script for generating notifications for Linux `stable-rc` tree. Also, add a test method to `kcidb/test_monitor.py` for the same.

### DIFF
--- a/kcidb/monitor/subscriptions/linux_stable_rc.py
+++ b/kcidb/monitor/subscriptions/linux_stable_rc.py
@@ -1,0 +1,32 @@
+"""Linux stable-rc subscription"""
+from datetime import (timezone, datetime, timedelta)
+
+from kcidb.monitor.output import NotificationMessage as Message
+
+
+def match_revision(revision):
+    """Match revisions of stable-rc tree"""
+    # Repo we're interested in
+    repo_url = "https://git.kernel.org/pub/scm/linux/kernel/git/stable/" \
+        "linux-stable-rc.git"
+
+    # If the revision is not from stable-rc repo,
+    # or there are no finished builds
+
+    if repo_url not in revision.repo_branch_checkouts:
+        return ()
+
+    if revision.builds_valid is None:
+        return ()
+
+    # Send notification 3 hours after a revision is created/updated
+    return (Message(
+        subject='KernelCI report for stable-rc: '
+                '{% include "stable_rc_revision_summary.txt.j2" %}',
+        to=["Jeny Sadadia <jeny.sadadia@collabora.com>",
+            "Gustavo Padovan <gustavo.padovan@collabora.com>",
+            "Shreeya Patel <shreeya.patel@collabora.com>"],
+        body='{% include "stable_rc_revision_description.txt.j2" %}',
+        cc=["KernelCI Results Staging <kernelci-results-staging@groups.io>"],
+        due=datetime.now(timezone.utc) + timedelta(hours=3)
+    ),)
--- a/kcidb/oo/__init__.py
+++ b/kcidb/oo/__init__.py
@@ -5,6 +5,7 @@ Kernel CI report object-oriented (OO) data representation.
 import sys
 from abc import ABC, abstractmethod
 from functools import reduce
+import re
 from cached_property import cached_property
 import kcidb.db
 from kcidb.misc import LIGHT_ASSERTS
@@ -580,6 +581,37 @@ class Build(Object, TestContainer,
             key=lambda v: VALID_PRIORITY[v], default=None
         )
         return self.__getattr__("valid") if valid is None else valid
+
+    @cached_property
+    def log_error(self):
+        """Get one-liner build error from log_excerpt.
+        Return None if self.log_excerpt is not set or is empty"""
+
+        log_line_list = self.log_excerpt.split("\n")
+
+        patterns = [
+            r'\.c:.*error',
+            r'\.h:.*error',
+            r'error.*modpost',
+            'No rule to make target',
+            'tail will be killed now',
+            'No such file',
+        ]
+
+        log_line_list.reverse()
+        for pattern in patterns:
+            for log_line in log_line_list:
+                # Allow leading and trailing optional characters
+                if re.search(r'.*' + pattern + r'.*', log_line):
+                    return log_line
+
+        # If pattern is not found return last non-empty log line
+        for log_line in log_line_list:
+            log_line = log_line.strip()
+            if log_line:
+                return log_line
+
+        return None
 
 
 class Test(Object, IncidentIssueBugContainer):
--- a/kcidb/oo/__init__.py
+++ b/kcidb/oo/__init__.py
@@ -390,6 +390,22 @@ class Node(IncidentIssueBugContainer):
             waived_status_nodes[node.waived][node.status].append(node)
         return waived_status_nodes
 
+    @cached_property
+    def waived_status_tests(self):
+        """
+        A dictionary of all waived values and dictionaries of all status
+        values and lists of tests (test runs) with corresponding waived and
+        status values, for this and all child nodes, all in order of
+        decreasing priority.
+        """
+        waived_status_tests = {
+            waived: {status: [] for status in TEST_STATUS_PRIORITY}
+            for waived in TEST_WAIVED_PRIORITY
+        }
+        for test in self.tests:
+            waived_status_tests[test.waived][test.status].append(test)
+        return waived_status_tests
+
     def __getitem__(self, name):
         assert isinstance(name, (str, type(None)))
         return Node(self, name)
--- a/kcidb/templates/stable_rc_build.j2
+++ b/kcidb/templates/stable_rc_build.j2
@@ -1,0 +1,40 @@
+{# Build template macros #}
+
+{% macro build_stats(container) %}
+    {% if container.builds %}
+        {% set invalid_builds =
+                container.builds | selectattr("valid", "false") | list %}
+        {% set valid_builds =
+                container.builds | selectattr("valid", "true") | list %}
+        {% set invalid_build_count = invalid_builds | length %}
+        {% set valid_build_count = valid_builds | length %}
+        {{- valid_build_count | string + " passed, " +
+            invalid_build_count | string + " failed" }}
+    {% endif %}
+{% endmacro %}
+
+{% macro container_summary(container, max_list_len) %}
+    {% if container.builds %}
+        {{- "\nBUILDS" }}
+        {% set invalid_builds =
+                container.builds | selectattr("valid", "false") | list %}
+        {% set invalid_build_count = invalid_builds | length %}
+        {% if invalid_builds %}
+            {{- "\n    Failures" }}
+            {% for origin, builds in invalid_builds|groupby("origin") %}
+                {% for build in builds %}
+                    {{- "       -"+  build.architecture + " (" + build.config_name + ")" }}
+                    {{- "\n       Build detail: https://kcidb.kernelci.org/d/build/build?orgId=1&var-id=" + build.id}}
+                    {% if build.log_excerpt %}
+                        {{- "       Build error: " + build.log_error }}
+                    {% endif %}
+                {% endfor %}
+                {{- "       CI system: " + origin + "\n\n"-}}
+            {% endfor %}
+        {% else %}
+            {{- "\n    No build failures found" }}
+        {% endif %}
+    {% else %}
+        {{- "\n    No builds found" }}
+    {% endif %}
+{% endmacro %}
--- a/kcidb/templates/stable_rc_test.j2
+++ b/kcidb/templates/stable_rc_test.j2
@@ -1,0 +1,37 @@
+{# Test macros #}
+
+{% macro tests_stats(container) %}
+    {% if container.tests %}
+        {% set waived_status_nodes = container.tests_root.waived_status_nodes %}
+        {% set failed_nodes = waived_status_nodes[false]["FAIL"] %}
+        {% set passed_nodes = waived_status_nodes[false]["PASS"] %}
+        {{- passed_nodes | length | string + " passed, " +
+            failed_nodes | length | string + " failed" }}
+    {% endif %}
+{% endmacro %}
+
+{% macro container_summary(container, max_list_len) %}
+    {{- "\nBOOT TESTS" }}
+    {% if container.tests %}
+        {% set failed_boot_tests = container.tests_root["boot"].waived_status_tests[false]["FAIL"] %}
+        {% if failed_boot_tests %}
+            {{- "\n    Failures" }}
+            {% for origin, boot_tests in failed_boot_tests|groupby("origin") %}
+                {% for architecture, tests in boot_tests|groupby("build.architecture") %}
+                    {% set device_list = tests | selectattr('misc.platform', 'defined') | list %}
+                    {% if device_list %}
+                        {{- "      " + architecture }}:({{ tests|map(attribute="build.config_name") | unique | join("") }})
+                    {% endif %}
+                    {% for device in device_list %}
+                        {{- "      -" + device.misc.platform }}
+                    {% endfor %}
+                {% endfor %}
+                {{- "      CI system: " + origin + "\n" }}
+            {% endfor %}
+        {% else %}
+            {{- "\n    No boot failures found" }}
+        {% endif %}
+    {% else %}
+        {{- "\n    No tests found" }}
+    {% endif %}
+{% endmacro %}
--- a/kcidb/test_monitor.py
+++ b/kcidb/test_monitor.py
@@ -843,3 +843,113 @@ def test_email_generated(empty_deployment):
         # Check we get no more notification messages
         assert len(email_subscriber.pull(1, 5)) == 0, \
             f"Extra emails for {io_version!r}"
+
+
+def test_linux_stable_rc():
+    """Check stable-rc subscription works"""
+    notifications = match(
+        {
+            "version": {
+                "major": 4,
+                "minor": 3
+            },
+            "checkouts": [
+                {
+                    "id": "maestro:66969be5aa07c494f8d22863",
+                    "origin": "maestro",
+                    "tree_name": "stable-rc",
+                    "git_repository_url": "https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable-rc.git",
+                    "git_commit_hash": "51945679d212aae61a418eff41370c13da94f94d",
+                    "git_repository_branch": "linux-5.4.y",
+                    "git_commit_name": "v5.4.279-79-g51945679d212",
+                    "patchset_files": [],
+                    "patchset_hash": "",
+                    "start_time": "2024-07-16T16:12:21.683+00:00",
+                    "contacts": [],
+                    "valid": True
+                },
+                {
+                    "id": "tuxsuite:51945679d212aae61a418eff41370c13da94f94d",
+                    "origin": "tuxsuite",
+                    "git_repository_url": "https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable-rc.git",
+                    "git_commit_hash": "51945679d212aae61a418eff41370c13da94f94d",
+                    "git_commit_name": "v5.4.279-79-g51945679d212",
+                    "patchset_files": [],
+                    "patchset_hash": "",
+                    "start_time": "2024-07-16T16:15:56+00:00",
+                    "contacts": [],
+                    "valid": True
+                }
+            ],
+            "builds": [
+                {
+                    "id": "maestro:66969c63aa07c494f8d22ad8",
+                    "checkout_id": "maestro:66969be5aa07c494f8d22863",
+                    "origin": "maestro",
+                    "comment": "v5.4.279-79-g51945679d212",
+                    "start_time": "2024-07-16T16:14:27.761+00:00",
+                    "architecture": "arm64",
+                    "config_name": "cros://chromeos-5.4/arm64/chromiumos-qualcomm.flavour.config",
+                    "valid": False,
+                    "log_excerpt": "./arch/arm64/include/asm/memory.h: " \
+                        "In function ‘__tag_set’:\n" \
+                        "./arch/arm64/include/asm/memory.h:238:22: warning: cast from "\
+                        "pointer to integer of different size [-Wpointer-to-int-cast]\n" \
+                        "238 | u64 __addr = (u64)addr & ~__tag_shifted(0xff);\n" \
+                        "| ^\n" \
+                        "/tmp/ccFJEFDb.s: Assembler messages:\n" \
+                        "/tmp/ccFJEFDb.s:129: Error: invalid barrier type -- `dmb ishld'\n" \
+                        "/tmp/ccFJEFDb.s:234: Error: invalid barrier type -- `dmb ishld'\n" \
+                        "/tmp/ccFJEFDb.s:510: Error: invalid barrier type -- `dmb ishld'\n" \
+                        "/tmp/ccFJEFDb.s:537: Error: invalid barrier type -- `dmb ishld'\n" \
+                        "/tmp/ccFJEFDb.s:1132: Error: invalid barrier type -- `dmb ishld'\n" \
+                        "/tmp/ccFJEFDb.s:1216: Error: invalid barrier type -- `dmb ishld'\n" \
+                        "make[1]: *** [arch/arm64/kernel/vdso32/Makefile:166: " \
+                        "arch/arm64/kernel/vdso32/vgettimeofday.o] Error 1\n" \
+                        "make: *** [arch/arm64/Makefile:170: vdso_prepare] Error 2"
+                },
+                {
+                    "id": "tuxsuite:2jKrMEMuY667cDnJgLrHdpnjxvS",
+                    "checkout_id": "tuxsuite:51945679d212aae61a418eff41370c13da94f94d",
+                    "origin": "tuxsuite",
+                    "start_time": "2024-07-16T16:15:59+00:00",
+                    "architecture": "s390",
+                    "config_name": "defconfig",
+                    "valid": False
+                },
+                {
+                    "id": "maestro:66969c5faa07c494f8d22ac1",
+                    "checkout_id": "maestro:66969be5aa07c494f8d22863",
+                    "origin": "maestro",
+                    "comment": "v5.4.279-79-g51945679d212",
+                    "start_time": "2024-07-16T16:14:23.836+00:00",
+                    "architecture": "x86_64",
+                    "config_name": "x86_64_defconfig",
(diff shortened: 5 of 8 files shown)


JenySadadia:
Hello @spbnick 

This is ready to integrate and get real-time email notifications.
Can you help me to get it deployed?

kcidb/oo/__init__.py
```diff
@@ -187,6 +188,39 @@ def architecture_valid_builds(self):
             reverse=True
         ))
 
+    def get_build_failure_log(self, log_excerpt):
```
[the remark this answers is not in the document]
spbnick: After you've done that, you can add a cached property (e.g. `build_log_excerpt_error`) to the `BuildContainer`, which would get the unique strings from `log_excerpt_errors` of each build, and return unique ones in a list. This way you can get build errors for any build container (revision, checkout, or issue).

Something like this (and assuming `log_excerpt_errors` is a set):
```python
class BuildContainer(ABC):
    """Abstract build container"""

    # . . .

    @cached_property
    def build_log_excerpt_errors(self):
        """A set of errors extracted from build log excerpts (if any)"""
        return reduce(lambda x, y: x | y,
                      build.log_excerpt_errors for build in self.builds)
```

kcidb/monitor/subscriptions/linux_stable_rc.py
```diff
@@ -0,0 +1,35 @@
...
+from datetime import (timezone, datetime, timedelta)
+
+from kcidb.monitor.output import NotificationMessage as Message
+
+
+def match_revision(revision):
+    """Match revisions of stable-rc tree"""
+    # Repo we're interested in
+    repo_url = "https://git.kernel.org/pub/scm/linux/kernel/git/stable/\
+linux-stable-rc.git"
```
spbnick, line 11: Please wrap this like this instead (with clearer indentation):
```python
    repo_url = "https://git.kernel.org/pub/scm/linux/kernel/git/stable/" \
        "linux-stable-rc.git"
```

kcidb/monitor/subscriptions/linux_stable_rc.py
```diff
@@ -0,0 +1,35 @@
...
+        return ()
+
+    if revision.builds_valid is None:
+        return ()
+
+    # We need to send the notification one hour after the last build came in.
+    # However, we currently have no way of distinguishing build updates
+    # from updates to build's tests, and we can't send a notification
+    # about a revision from a build update yet. So we're sending the
+    # notification one hour after any revision update.
```
spbnick, line 26: Does this comment really apply to what you wanted to do here, as is? If not, please describe what you actually want and/or what you're doing.

kcidb/monitor/subscriptions/linux_stable_rc.py
```diff
@@ -0,0 +1,35 @@
...
+    return (Message(
+        subject='KernelCI report for stable-rc: \
+{% include "stable_rc_revision_summary.txt.j2" %}',
```
spbnick, line 29: Please fix this wrapping similarly.

kcidb/oo/__init__.py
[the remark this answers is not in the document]
spbnick: And then you would be able to do something like this in the templates:

```jinja2
{% for error in revision.build_log_excerpt_errors %}
{{error}}
{% endfor %}
```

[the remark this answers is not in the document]
spbnick: Or, if you wish, you can instead add a property which returns a dictionary of errors to lists of builds where it appears.

kcidb/oo/__init__.py
```diff
@@ -187,6 +188,39 @@ def architecture_valid_builds(self):
...
+            for log in log_list:
+                if re.search(pattern, log):
+                    return log
+
+        # If pattern is not found return last non-empty log line
+        log_list.reverse()
+        for log in log_list:
+            log = log.strip()
+            if log:
+                return log
```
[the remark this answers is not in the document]
spbnick: If you don't feel comfortable messing around with those regexes, splitting lines and doing loops is alright too. The memory won't be an issue until we get to processing the full logs, and your approach is quite clear. 👍🏻

kcidb/oo/__init__.py
```diff
@@ -187,6 +188,39 @@ def architecture_valid_builds(self):
...
+            r'.*\.h:.*error.*',
+            r'.*error.*modpost',
+            'No rule to make target',
+            'tail will be killed now',
+            r'.*error.*',
+            r'.*No such file.*',
+            r'kern.*No irq handler for vector',
+            'package_dtbs=done',
+            'RESULT=fail'
+        ]
```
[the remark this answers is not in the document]
JenySadadia: I think it also catches other failures with `RESUT=fail`.
I have used the same set of error patterns @helen-fornazier is using for the dashboard.
Maybe she could confirm what errors she included here.

kcidb/test_monitor.py
```diff
@@ -843,3 +843,98 @@ def test_email_generated(empty_deployment):
...
+                    "comment": "v5.4.279-79-g51945679d212",
+                    "start_time": "2024-07-16T16:14:27.761+00:00",
+                    "architecture": "arm64",
+                    "config_name": "cros://chromeos-5.4/arm64/chromiumos-qualcomm.flavour.config",
+                    "valid": False,
+                    "log_excerpt": "./arch/arm64/include/asm/memory.h: In function ‘__tag_set’:\n./arch/arm64/include/asm/memory.h:238:22: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]\n238 | u64 __addr = (u64)addr & ~__tag_shifted(0xff);\n| ^\n/tmp/ccFJEFDb.s: Assembler messages:\n/tmp/ccFJEFDb.s:129: Error: invalid barrier type -- `dmb ishld'\n/tmp/ccFJEFDb.s:234: Error: invalid barrier type -- `dmb ishld'\n/tmp/ccFJEFDb.s:510: Error: invalid barrier type -- `dmb ishld'\n/tmp/ccFJEFDb.s:537: Error: invalid barrier type -- `dmb ishld'\n/tmp/ccFJEFDb.s:1132: Error: invalid barrier type -- `dmb ishld'\n/tmp/ccFJEFDb.s:1216: Error: invalid barrier type -- `dmb ishld'\nmake[1]: *** [arch/arm64/kernel/vdso32/Makefile:166: arch/arm64/kernel/vdso32/vgettimeofday.o] Error 1\nmake: *** [arch/arm64/Makefile:170: vdso_prepare] Error 2"
```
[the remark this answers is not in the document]
JenySadadia: Updated it.

kcidb/monitor/subscriptions/linux_stable_rc.py
JenySadadia: Fixed it.

kcidb/monitor/subscriptions/linux_stable_rc.py
JenySadadia: Fixed it.

kcidb/oo/__init__.py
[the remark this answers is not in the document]
nuclearcat: Btw in some cases logs might be enormous, but i guess unlikely they will reach kcidb with such size.
Also this patterns need to be well-though, maybe even not hardcoded, but separate (config?) file, as they might need to be updated often.

kcidb/oo/__init__.py
```diff
@@ -233,6 +233,14 @@ def tests_root(self):
...
         return Node(self, "")
 
+    def get_boot_tests(self, tests):
+        """Get boot tests by matching "boot" prefix in test path"""
+        boot_tests = []
+        for test in tests:
+            if test.path.split(".")[0].startswith("boot"):
+                boot_tests.append(test)
+        return boot_tests
+
```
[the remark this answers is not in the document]
JenySadadia: I don't get this.
I tried `container.tests_root["boot"].waived_status_nodes` (I assume you meant `container` as `revision` is undefined in the template) and received some kind of dictionary:
```
{False: {&#x27;FAIL&#x27;: [], &#x27;ERROR&#x27;: [], &#x27;M=
ISS&#x27;: [], &#x27;PASS&#x27;: [], &#x27;DONE&#x27;: [], &#x27;SKIP&#x27;: =
[], None: []}, True: {&#x27;FAIL&#x27;: [], &#x27;ERROR&#x27;: [], &#x27;MISS=
&#x27;: [], &#x27;PASS&#x27;: [], &#x27;DONE&#x27;: [], &#x27;SKIP&#x27;: [],=
 None: []}, None: {&#x27;FAIL&#x27;: [], &#x27;ERROR&#x27;: [], &#x27;MISS&#x=
27;: [], &#x27;PASS&#x27;: [], &#x27;DONE&#x27;: [], &#x27;SKIP&#x27;: [], No=
ne: []}}
```

kcidb/oo/__init__.py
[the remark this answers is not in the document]
JenySadadia: > If you don't feel comfortable messing around with those regexes, splitting lines and doing loops is alright too. The memory won't be an issue until we get to processing the full logs, and your approach is quite clear. 👍🏻

Yes, I would rather keep it simpler.

kcidb/oo/__init__.py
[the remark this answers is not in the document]
spbnick: That's the same `waived_status_nodes` dictionary you were using before, in your templates, only restricted to `boot` tests and its subtests.

BTW, could you perhaps share the data you were using to test this?

kcidb/oo/__init__.py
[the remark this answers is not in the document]
spbnick: > Btw in some cases logs might be enormous, but i guess unlikely they will reach kcidb with such size.
Also this patterns need to be well-though, maybe even not hardcoded, but separate (config?) file, as they might need to be updated often.

We would have to limit the size of files we archive, with the particular limit yet to be decided. Otherwise, yes, I think we should process only the last N MB of each log file for performance reasons.

kcidb/oo/__init__.py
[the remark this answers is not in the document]
JenySadadia: Thanks for the pointers.
I moved `get_build_failure_log` to `Build` class from `BuildContainer` and it internally uses `log_excerpt` to get the failure log. That's a lot easier now. Thanks for the help.

I didn't go for the below solution as I need to print build failure inline with separate builds. The above approach seemed easier.
```
{% for error in revision.build_log_excerpt_errors %}
{{error}}
{% endfor %}
```

kcidb/oo/__init__.py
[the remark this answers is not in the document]
JenySadadia: Okay, I'll check again.
I am using the same test data I have added to `test_monitor.py`.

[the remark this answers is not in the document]
JenySadadia: OK, I tried this:
```
{% set boot_tests = container.tests_root["boot"].waived_status_nodes %}
{% set failed_boot_tests = boot_tests[false]["FAIL"] %}
{{- failed_boot_tests }}
```
And that prints an empty list.

kcidb/oo/__init__.py
```diff
@@ -581,6 +582,42 @@ def valid(self):
         )
         return self.__getattr__("valid") if valid is None else valid
 
+    @cached_property
+    def get_build_failure_log(self):
+        """Get one-liner build error from log_excerpt.
+        Return None if self.log_excerpt is not set"""
```
spbnick, line 596: This is not the only case where `None` could be returned, it will also be returned if `log_excerpt` is empty or consist entirely of whitespace, which could be good to mention.

spbnick (changes requested):

kcidb/monitor/subscriptions/linux_stable_rc.py
spbnick: Um, somehow I thought I already responded to this comment, but can't find it. Sorry, if you get it twice.

The comment implies that you somehow *wanted* to have notifications *only* after build updates. Was that really the case? If not, please remove that from the comment.

kcidb/oo/__init__.py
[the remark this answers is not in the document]
spbnick: OK, the `waived_status_nodes` would be a dictionary of waived->status->*nodes*. If you want a waived->status->*tests* for a particular node, you can apply this:
```diff
diff --git a/kcidb/oo/__init__.py b/kcidb/oo/__init__.py
index de54af1..73676ab 100644
--- a/kcidb/oo/__init__.py
+++ b/kcidb/oo/__init__.py
@@ -331,6 +331,22 @@ class Node(IncidentIssueBugContainer):
             )
         ]
 
+    @cached_property
+    def waived_status_tests(self):
+        """
+        A dictionary of all waived values and dictionaries of all status
+        values and lists of tests (test runs) with corresponding waived and
+        status values, for this and all child nodes, all in order of
+        decreasing priority.
+        """
+        waived_status_tests = {
+            waived: {status: [] for status in TEST_STATUS_PRIORITY}
+            for waived in TEST_WAIVED_PRIORITY
+        }
+        for test in self.tests:
+            waived_status_tests[test.waived][test.status].append(test)
+        return waived_status_tests
+
     @cached_property
     def incidents(self):
         """A list of incidents for this and all child nodes"""
```

And then refer to `waived_status_tests` instead of `waived_status_nodes`. E.g. this: `container.tests_root["boot"].waived_status_tests[false]["FAIL"]` will give you all non-waived failed `boot` or `boot.*` test *runs* (actual rows in the `tests` table).

If you go this route, please apply the above patch in a separate commit. Thank you!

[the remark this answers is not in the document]
JenySadadia: Thanks for the above patch.
I've added it and updated related template as well.
Should I set you as the author for the commit?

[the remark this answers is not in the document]
spbnick: Awesome!

> Should I set you as the author for the commit?

Don't bother :grin:

kcidb/monitor/subscriptions/linux_stable_rc.py
JenySadadia: I think we should send the report once all builds and tests are received for the particular checkout.
In the case of maestro, we set checkout node's `valid` flag to `True`  once all the build and test executions have been completed. 
But that might not be true for other CI systems.
I think we should wait longer than 1 hr after a build is updated.
Do we currently have any frequency set for sending `stable-rc` report that you generate manually? @crazoes

spbnick: Yeah, we cannot expect anybody else to do the same thing with checkouts. Also such use is really a hack, as `valid` means that you've managed to apply the referenced patch. Nothing else. And you don't have patches, so it should always be valid.

Whatever logic you find out works for you, please put the corresponding comment over it, don't leave the original. It pertains to the specific plan and workaround we had with Mark.

kcidb/oo/__init__.py
JenySadadia: Fixed it.

kcidb/oo/__init__.py
```diff
@@ -581,6 +582,42 @@ def valid(self):
         )
         return self.__getattr__("valid") if valid is None else valid
 
+    @cached_property
+    def get_build_failure_log(self):
```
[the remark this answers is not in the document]
JenySadadia: Renamed it to `log_error`.

spbnick:
Thanks for the fixes, Jeny! Please tell me when you're ready for another review.

JenySadadia:
> Thanks for the fixes, Jeny! Please tell me when you're ready for another review.

Hello,
Yes, it's ready for another round of review. I'll address the remaining unresolved comments after talking to Helen and Shreeya.

kcidb/oo/__init__.py
```diff
@@ -187,6 +188,39 @@ def architecture_valid_builds(self):
...
+        patterns = [
+            'Kernel panic',
+            't access tty',
+            r'.*\.c:.*error.*',
+            r'.*\.h:.*error.*',
+            r'.*error.*modpost',
+            'No rule to make target',
+            'tail will be killed now',
+            r'.*error.*',
+            r'.*No such file.*',
```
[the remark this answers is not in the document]
JenySadadia: Fixed it.

kcidb/monitor/subscriptions/linux_stable_rc.py
crazoes: @JenySadadia we send the report once Greg sends out an email notification of the release. So for stable-rc specifically we need to parse the stable-rc tree and whenever there is a new changes pushed, we need to generate a report and send it to me initially. I will then forward the email after verification.

I did work on a script to parse the stable-rc tree long time back :- 
https://gitlab.collabora.com/shreeya/kernelci/-/blob/main/email_checker.py

You can integrate this script to generate a report accordingly.

JenySadadia: Polling for an email in the inbox looks a bit tricky here.
Can we integrate this kind of script in KCIDB? @spbnick

spbnick: I assume we want to send our results *not before* Greg's announcement, and preferably in response to it, right?

KCIDB subscriptions are done for database changes only. There's no object for a " particular message posted on a particular maillist" (or perhaps a "release announcement") in KCIDB database, and therefore no event on its appearance/update like the rest of the data. And I don't think we should add it for Greg only, even though his work is very important.

I think our plan for now could be "generate a notification when done", and let a human decide when to send it, as at the moment we require a human for review anyway. Then, when we're in a better place, and there's not so much work to do everywhere, we could think of making it better.

kcidb/templates/stable_rc_build.j2
```diff
@@ -0,0 +1,40 @@
...
+            {{- "\n    Failures" }}
+            {% for origin, builds in invalid_builds|groupby("origin") %}
+                {% for build in builds %}
+                    {{- "       -"+  build.architecture + " (" + build.config_name + ")" }}
+                    {{- "\n       Build detail: https://kcidb.kernelci.org/d/build/build?orgId=1&var-id=" + build.id}}
+                    {% if build.log_excerpt %}
+                        {{- "       Build error: " + build.log_error }}
+                    {% endif %}
+                {% endfor %}
+                {{- "       CI system: " + origin + "\n\n"-}}
```
spbnick, line 32: I think spelling out which CI system reported which build is not really necessary, and makes the list harder to read. I think it's enough to list the contributing systems at the end of all builds, on one line.

Interspersing errors with CI system names makes it harder to read, and makes the list longer.

Ideally, it would be good to group builds by a particular error, and show how many builds had it. We can introduce that once we are able to encode errors into "issues", and be able to link to a particular issue for a particular revision.

kcidb/templates/stable_rc_test.j2
```diff
@@ -0,0 +1,37 @@
...
+        {% if failed_boot_tests %}
+            {{- "\n    Failures" }}
+            {% for origin, boot_tests in failed_boot_tests|groupby("origin") %}
+                {% for architecture, tests in boot_tests|groupby("build.architecture") %}
+                    {% set device_list = tests | selectattr('misc.platform', 'defined') | list %}
+                    {% if device_list %}
+                        {{- "      " + architecture }}:({{ tests|map(attribute="build.config_name") | unique | join("") }})
+                    {% endif %}
+                    {% for device in device_list %}
+                        {{- "      -" + device.misc.platform }}
```
spbnick, line 26: Yeah, we gotta standardize this, but we can merge it as is for the time being.

kcidb/monitor/subscriptions/linux_stable_rc.py
spbnick: We still need to fix this comment before merging.

kcidb/monitor/subscriptions/linux_stable_rc.py
```diff
@@ -0,0 +1,35 @@
...
+    # However, we currently have no way of distinguishing build updates
+    # from updates to build's tests, and we can't send a notification
+    # about a revision from a build update yet. So we're sending the
+    # notification one hour after any revision update.
+    return (Message(
+        subject='KernelCI report for stable-rc: '
+                '{% include "stable_rc_revision_summary.txt.j2" %}',
+        to=["Jeny Sadadia <jeny.sadadia@collabora.com>",
+            "Gustavo Padovan <gustavo.padovan@collabora.com>",
+            "Shreeya Patel <shreeya.patel@collabora.com>"],
```
spbnick, line 32: Could you add the `kernelci-results-staging@groups.io` into CC, so others could see the notifications?

kcidb/templates/stable_rc_build.j2
spbnick: I won't hold the PR over this, though, if you prefer it this way. In the end it's up to recipients, how they want their notifications.

spbnick (changes requested):
Thank you, Jeny! This looks much better now. I left a few comments, but there's nothing major left, only that comment in the subscription code, adding a CC, and removing non build-error regexes, as I requested earlier.

kcidb/monitor/subscriptions/linux_stable_rc.py
JenySadadia: Sure.

kcidb/templates/stable_rc_build.j2
JenySadadia: Isn't it useful to report which CI system reported which failure?
I think it's good to have that info for further communication for a specific failure.
The same goes for test failures.
> Ideally, it would be good to group builds by a particular error, and show how many builds had it. We can introduce that once we are able to encode errors into "issues", and be able to link to a particular issue for a particular revision.

I have grouped failures based on architecture atm. We can change it later as required.

kcidb/monitor/subscriptions/linux_stable_rc.py
JenySadadia: I guess Shreeya would like to review the report first and then forward this to the mailing list.
So, I don't think we should add `CC` here.

kcidb/oo/__init__.py
```diff
@@ -581,6 +598,43 @@ def valid(self):
...
         return self.__getattr__("valid") if valid is None else valid
 
+    @cached_property
+    def log_error(self):
+        """Get one-liner build error from log_excerpt.
+        Return None if self.log_excerpt is not set or is empty"""
+
+        log_line_list = self.log_excerpt.split("\n")
+
+        patterns = [
```
pawiecz, line 608: Looks build-related:
- r'\.c:.*error',
- r'\.h:.*error',
- 'No rule to make target',
- 'No such file',

Doesn't look build-related:
- 'Kernel panic',
- 't access tty',
- r'kern.*No irq handler for vector',

Needs double-checking:
- r'error.*modpost',
- 'tail will be killed now',
- 'error', (pretty generic pattern btw)
- 'package_dtbs=done',
- 'RESULT=fail'

nuclearcat: I am not sure if "'package_dtbs=done'," is error
Also i think generic 'error" might lead to lot of false positives, for example build warning for error.c

kcidb/monitor/subscriptions/linux_stable_rc.py
JenySadadia: Maestro usually allows a checkout to be completed with all builds and tests within 3 hrs. But of course, that time would differ per CI system basis.
So, I guess we can set the `due` time accordingly.
Bdw list of revisions is managed internally by KCIDB. So when is `match_revision` event triggered? On every checkout update or build update?

kcidb/oo/__init__.py
nuclearcat: "package_dtbs=done" should be definitely removed from error list

kcidb/templates/stable_rc_build.j2
spbnick: > Isn't it useful to report which CI system reported which failure?

I don't think that's really important. More important is what the configuration, the command-line arguments, and the toolchain were. The CI system shouldn't matter, ultimately. And you can see it when you open the specific build URL. Certainly not something you need to know immediately.

But sure, please go ahead with what you've already built.

kcidb/monitor/subscriptions/linux_stable_rc.py
spbnick: That mailing list is only for us, including for test notifications, so that's not really a problem.

kcidb/oo/__init__.py
nfraprado: I confirm @pawiecz 's assessment.

Additionally, `r'error.*modpost'` is also a build error.

`RESULT=fail` looks like LAVA reporting a result, so most likely not a build failure (but rather run failure).

`'tail will be killed now'` I have seen that in build logs in the past, so it is build-related, not sure if it is an error though.

JenySadadia: I appreciate all the help.
Updated the list accordingly.

kcidb/monitor/subscriptions/linux_stable_rc.py
JenySadadia: Okay, added it.

kcidb/monitor/subscriptions/linux_stable_rc.py
JenySadadia: Gentle ping ^ @spbnick

spbnick: Sorry, missed this question. `match_revision` is triggered on any new/updated checkout/build/test arriving for the particular revision.

JenySadadia: Okay. So, it means it will create a new report on every build or test submission.
In that case, we will have hundreds of reports (because of test submissions) for the same checkout.
And I don't think we want that in this case.
We can restrict it to listening to only checkout events and `match_checkout` won't work I guess as it won't contain any build or test data.
What do you suggest?

spbnick: The function is triggered on every update, yes. However it creates/updates a notification with the ID of the object it triggered on, the revision ID, plus the subscription name, and a notification with a particular ID is only sent *once* by the system. So on every update it will create/update a notification with the same ID, scheduled to be sent in an hour (or however long you want). This way the sending time is pushed forward by updates until the "last" update.

spbnick: I.e. the function *creates* or *updates* a notification, it doesn't send it.

JenySadadia: Thank you for the clarification.
I have set `3 hrs` delay for sending a notification based on some observations. We can tweak it once we have some real-time notifications.
Updated the comment accordingly.

spbnick (approved):
LGTM! Thank you, Jeny :heart:

PR #539, opened 2024-07-05
tests.yaml: remove `kernelci_baseline`

JenySadadia:
As KCI baseline tests are actually boot tests, we'll be referring these tests as `boot` from now on.


spbnick (approved):
LGTM, thank you, Jeny!

PR #540, opened 2024-07-05
tests.yaml: add `rt-tests`

JenySadadia:
Add RT linux test suite.


spbnick:
I don't like that there's `tests` in the name of a test suite, but since it seems to be generally recognized, we should keep it that way.

I'll also check with CKI, they're likely running those as well, which means we'll have to report them under this name.

LGTM, thank you!

Issue #545, opened 2024-07-11
Make the `tree_name` field an index


helen-fornazier:
Improve performance of the queries by making `tree_name` field an index.

We discussed about using the `git_repository_url instead`, but for Grafana it is better to show the `tree_name` instead.

spbnick:
This was done a while ago.

PR #548, opened 2024-07-29
postgresql: Add schema v4.7 with a bunch more indexes

spbnick:
Add PostgreSQL schema v4.7 adding a bunch more indexes, mostly for issues and incidents.

Concerns https://github.com/kernelci/kcidb/issues/545


spbnick:
Merging without review, since Helen is on PTO.

PR #550, opened 2024-08-05
Fix `stable-rc` report

JenySadadia:
- Fix boot tests overview
- Handle missing build architecture and config for build failure summary
- Only take into account results from `maestro` and `broonie` CI systems as of now
-  Add `Also checked out from` tag in `REVISION` section to list checkout URL with the same commit from trees other than `stable-rc`

### DIFF
--- a/kcidb/templates/stable_rc_revision_description.txt.j2
+++ b/kcidb/templates/stable_rc_revision_description.txt.j2
@@ -14,7 +14,7 @@ OVERVIEW
         Builds: {{ build_macros.build_stats(revision) }}
 {% endif %}
 {% if revision.tests %}
-         Tests: {{ test_macros.tests_stats(revision) }}
+    Boot tests: {{ test_macros.tests_stats(revision) }}
 {% endif %}
     CI systems: {{ revision.checkouts | map(attribute="origin") |
            unique | sort | join(", ") }}
--- a/kcidb/templates/stable_rc_revision_description.txt.j2
+++ b/kcidb/templates/stable_rc_revision_description.txt.j2
@@ -1,4 +1,5 @@
 {# Revision description template #}
+{% import "stable_rc.j2" as stable_rc_macros %}
 {% import "stable_rc_test.j2" as test_macros %}
 {% import "stable_rc_build.j2" as build_macros %}
 {% import "misc.j2" as misc_macros %}
@@ -17,7 +18,7 @@ OVERVIEW
     Boot tests: {{ test_macros.tests_stats(revision) }}
 {% endif %}
     CI systems: {{ revision.checkouts | map(attribute="origin") |
-           unique | sort | join(", ") }}
+           unique | sort | select('in', stable_rc_macros.selected_origins) | join(", ") }}
 
 REVISION
 
--- a/kcidb/templates/stable_rc_revision_description.txt.j2
+++ b/kcidb/templates/stable_rc_revision_description.txt.j2
@@ -29,13 +29,26 @@ REVISION
         {% if revision.git_commit_hash %}
             {{- "        hash: " + revision.git_commit_hash }}
         {% endif %}
+    {# The stable-rc repo's URL #}
+    {% set stable_rc_repo = 'https://git.kernel.org/pub/scm/linux/' +
+                            'kernel/git/stable/linux-stable-rc.git' %}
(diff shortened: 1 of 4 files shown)


spbnick:
LGTM! Thank you for a quick fix, Jeny, merging!

Next time, if you'd like to run deployment tests on your PR, you can push your commits to a branch in the upstream (this) repo.

spbnick:
Ah, not merging, will merge once you get the "draft" off!

JenySadadia:
I still have a couple of reviews to address. That's why I kept it in draft.

kcidb/monitor/subscriptions/linux_stable_rc.py
```diff
@@ -19,6 +19,22 @@ def match_revision(revision):
     if revision.builds_valid is None:
         return ()
 
+    # Only take into account results from maestro and broonie as of now
+    checkouts_copy = revision.checkouts.copy()
+    for checkout in checkouts_copy:
+        if checkout.origin not in ("maestro", "broonie"):
+            revision.checkouts.remove(checkout)
```
JenySadadia, line 26: Hi @spbnick 
To report results only from `maestro` and `broonie`, we need to filter out the data accordingly.
Filter for `revision.checkouts` isn't working where builds and tests are working fine.
Could you please check what's wrong with this?
I also observed `self.checkouts` in `Revision` class displays different memory address for checkout objects than `revision.checkouts` here.

spbnick: OK, first of  all, please  don't remove anything from the objects. They're cached, and they will be reused for other subscriptions. Treat them as read-only. I don't think I ever implemented proper copying support, so don't rely on that. Finally, even if you copy and modify them somehow, the templates won't see those modifications, because they're supplied with the original objects provided to your subscription's match function, not your copies.

You'll have to filter the objects on the fly as you're iterating through them in your subscription function, as well as in your templates. Thankfully both Python and Jinja2 (to some extent) are good at that.

Otherwise, we would need to implement advanced filtering, where you could get the same object, but with only specified origin's data. I've been planning that for a while, but so far haven't really needed it. Perhaps we'll need to do that soon. For now, please try ad-hoc filtering and see how hard/easy it is.

> I also observed self.checkouts in Revision class displays different memory address for checkout objects than revision.checkouts here.

I would have to see exact code reproducing this to tell you what's going on.

JenySadadia: Okay. Thank you. I have updated the code accordingly.

kcidb/templates/stable_rc_revision_description.txt.j2
```diff
@@ -29,13 +29,23 @@ REVISION
         {% if revision.git_commit_hash %}
             {{- "        hash: " + revision.git_commit_hash }}
         {% endif %}
+        {% set stable_rc_checkouts, other_checkouts = revision.filter_stable_rc_checkouts() %}
```
spbnick, line 32: You can do the filtering inside template quite easily. E.g.:

```jinja2
    {# The stable-rc repo's URL #}
    {% set stable_rc_repo = 'https://git.kernel.org/pub/scm/linux/' +
                            'kernel/git/stable/linux-stable-rc.git' %}
    {# List of other repo's URLs #}
    {% set other_repos = revision.repo_branch_checkouts.keys() |
                         reject('==', stable_rc_repo) %}
    Checked out from
        {{- "        " +
            (([stable_rc_repo] + (revision.repo_branch_checkouts[stable_rc_repo] | list)) |
            reject("none") | join(" ")) }}
    {% if other_repos %}
    Also checked out from
        {% for repo, branch_checkouts in revision.repo_branch_checkouts.items() %}
            {% if repo in other_repos %}
                {{- "        " +
                    (([repo] + (branch_checkouts | list)) |
                    reject("none") | join(" ")) }}
            {% endif %}
        {% endfor %}
    {% endif %}
```

spbnick (changes requested):
Looks good mostly, thank you, Jeny!

I left a few comments inline.

Also, let's get the fix in first and then focus on improvements, next time :+1:

JenySadadia:
Thanks Nick. I have addressed all the comments.

spbnick (approved):
LGTM! Thank you, Jeny :heart: Merging!

PR #555, opened 2024-08-08
kcidb/oo: remove a build error pattern

JenySadadia:
The below build log isn't an actual error message: `Build script is completed, tail will be killed now`. Hence, remove the related pattern from the list of build errors.

### DIFF
--- a/kcidb/oo/__init__.py
+++ b/kcidb/oo/__init__.py
@@ -610,7 +610,6 @@ class Build(Object, TestContainer,
             r'\.h:.*error',
             r'error.*modpost',
             'No rule to make target',
-            'tail will be killed now',
             'No such file',
         ]
 


JenySadadia:
@crazoes I am wondering what should be the actual build error log for the report for failures such as `https://kcidb.kernelci.org/d/build/build?orgId=1&var-id=maestro:66b399764ec9bd2e74797ca8`.

JenySadadia:
Looks like the build was completed successfully and somehow reported with `fail` result to KCIDB:
https://kernelci-api.westus3.cloudapp.azure.com/viewer?node_id=66b399764ec9bd2e74797ca8

nuclearcat:
Probably it did timeout due overload of build cluster:
```
 "error_code": "node_timeout",
    "error_msg": "Node timed-out",
```
So we need to delete error_code and msg when build complete.

crazoes:
@JenySadadia these are false positives and maybe the kernel build isn't failing but the kselftest build is failing here. I have asked @musamaanjum to work on fixing these issues. You can check more details here  :- https://github.com/orgs/kernelci/projects/22/views/1?pane=issue&itemId=72345288

JenySadadia:
@crazoes The issue is from the KCIDB bridge side.
When the node times out, we send it with `FAIL` status as KCIDB doesn't have `incomplete/miss` for build results.
@spbnick  How do I send incomplete builds or should we avoid to send such builds to KCIDB?

helen-fornazier:
fyi, there is an issue open to transform the valid status in builds to be a enum: https://github.com/kernelci/kcidb-io/issues/82

PR #556, opened 2024-08-09
stable-rc: fix empty builds failures section

JenySadadia:
Fix empty builds section like the below:
```
BUILDS

    Failures
```


spbnick:
LGTM!

spbnick:
@JenySadadia, if you'd like, we can dump a bunch of recent data from the database for you to try the subscription on, so your feedback loop is shorter.

JenySadadia:
> @JenySadadia, if you'd like, we can dump a bunch of recent data from the database for you to try the subscription on, so your feedback loop is shorter.

Yes, that'd be helpful.

PR #569, opened 2024-09-06
Support I/O schema v4.5 for checkouts

JenySadadia:


kcidb/db/postgresql/v04_09.py
```diff
@@ -0,0 +1,165 @@
...
+        CREATE TYPE UNIT_PREFIX AS ENUM ('metric', 'binary');
+    EXCEPTION
+        WHEN duplicate_object THEN null;
+    END $$
+""")
+
+
+# It's OK, pylint: disable=too-many-ancestors
+class Schema(PreviousSchema):
+    """PostgreSQL database schema v4.8"""
```
spbnick, line 23: Please update the version number in the docstring.

kcidb/db/postgresql/v04_09.py
```diff
@@ -0,0 +1,165 @@
+"""Kernel CI report database - PostgreSQL schema v4.8"""
```
spbnick, line 1: Please update the version number in the docstring.

kcidb/db/postgresql/v04_09.py
```diff
@@ -0,0 +1,165 @@
...
+from .v04_08 import Schema as PreviousSchema
+
+
+CREATE_UNIT_PREFIX_TYPE_STATEMENT = textwrap.dedent("""\
+    DO $$ BEGIN
+        CREATE TYPE UNIT_PREFIX AS ENUM ('metric', 'binary');
+    EXCEPTION
+        WHEN duplicate_object THEN null;
+    END $$
+""")
```
spbnick, line 18: We don't need this here.

kcidb/db/postgresql/v04_09.py
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    # A map of index names and schemas
+    INDEXES = merge_dicts(PreviousSchema.INDEXES, dict(
+        checkouts_git_commit_tags=Index(
+            "checkouts", ["git_commit_tags"]
+        ),
+        checkouts_git_commit_message=Index(
+            "checkouts", ["git_commit_message"]
+        ),
```
spbnick, line 59: A regular BTREE index would be useless for this field, because nobody is going to search for the exact complete message, or even its beginning. We would need to support full-text search on this field, and for that we would need to read this chapter: https://www.postgresql.org/docs/current/textsearch.html

So let's not add the index right now. We mostly need to display the commits (or rather their subjects), right now, and that doesn't require an index.

kcidb/db/postgresql/v04_09.py
```diff
@@ -0,0 +1,165 @@
...
+        """
+        Inerit the database data from the previous schema version (if any).
+
+        Args:
+            conn:   Connection to the database to inherit. The database must
+                    comply with the previous version of the schema.
+        """
+        assert isinstance(conn, cls.Connection)
+        with conn, conn.cursor() as cursor:
+            cursor.execute(CREATE_UNIT_PREFIX_TYPE_STATEMENT)
```
spbnick, line 140: This is already done in the previous schema, please remove.

kcidb/db/sqlite/v04_04.py
```diff
@@ -0,0 +1,115 @@
+"""Kernel CI report database - SQLite schema v4.3"""
```
spbnick, line 1: Please update the version in the docstring.

kcidb/db/sqlite/v04_04.py
```diff
@@ -0,0 +1,115 @@
...
+                git_repository_branch_tip=BoolColumn(),
+                tree_name=TextColumn(),
+                message_id=TextColumn(),
+                start_time=TimestampColumn(),
+                log_url=TextColumn(),
+                log_excerpt=TextColumn(),
+                comment=TextColumn(),
+                valid=BoolColumn(),
+                misc=JSONColumn(),
+            )),
```
spbnick, line 85: You don't need to duplicate the `Table()` schema here, because it should be correct in the `v04_00.py` one. Just replace the statement. Like `v04_01.py` does.

kcidb/db/sqlite/v04_04.py
```diff
@@ -0,0 +1,115 @@
...
+        PreviousSchema.TABLES_ARGS,
+        checkouts=merge_dicts(
+            PreviousSchema.TABLES_ARGS["checkouts"],
+            columns=dict(
+                {
+                    "git_commit_tags": JSONColumn(),
+                    "git_commit_message": TextColumn(),
+                    "git_repository_branch_tip": BoolColumn(),
+                }
+            )
```
spbnick, line 34: You're replacing *all* columns with the new ones here. That's why CI fails. See how `v04_03.py` does this.

Also you can run pytest locally to reproduce this.

kcidb/orm/data.py
```diff
@@ -344,10 +344,14 @@ def format_dot(self):
...
                 id=_CHECKOUT['id'],
                 git_commit_hash=_CHECKOUT['git_commit_hash'],
+                git_commit_tags=_CHECKOUT['git_commit_tags'],
+                git_commit_message=_CHECKOUT['git_commit_message'],
                 patchset_hash=_CHECKOUT['patchset_hash'],
                 origin=_CHECKOUT['origin'],
                 git_repository_url=_CHECKOUT['git_repository_url'],
                 git_repository_branch=_CHECKOUT['git_repository_branch'],
+                git_repository_branch_tip=_CHECKOUT[
+                    'git_repository_branch_tip'],
```
spbnick, line 354: Could you put the closing bracket on the next line, on the same column the key starts at? This way it will be easier to notice.

spbnick (changes requested):
Thank you, Jeny! I left a few comments, including the one explaining the CI failures you're seeing.

Aside from that, if you push your changes to a branch in the upstream repo (instead of your fork) and open a PR from **there**, you would get your changes tested against a deployment. That would include PostgreSQL and BigQuery tests.

kcidb/db/postgresql/v04_09.py
JenySadadia: Yes, it was for `number.prefix`. Dropped it.

kcidb/db/postgresql/v04_09.py
JenySadadia: Dropped the index.

kcidb/db/sqlite/v04_04.py
JenySadadia: Thanks for the pointer.

kcidb/db/sqlite/v04_04.py
JenySadadia: Thanks for the pointer.

spbnick:
Please move your logic for fluster tests from Python to Jinja template code, either the way I did it, or whatever way you prefer. Do not keep my `DRAFT:` commit, but make sure to keep the max body size amendment, as a separate commit. Thank you!

spbnick:
Also, I think we need to make the notification generation a bit more flexible, so you could prepare data to be rendered inside the subscription, and then pass it to Jinja.

spbnick:
Wrong PR :see_no_evil:

JenySadadia:
> > I'll keep that in mind next time. Thanks.
> 
> Let's do that now, as this is a good way of testing PostgreSQL and BigQuery support. Otherwise we would have to merge it untested.

OK, I'll create a new PR then.

JenySadadia:
Replaced by https://github.com/kernelci/kcidb/pull/571

PR #571, opened 2024-09-11
Support I/O schema v4.5 for checkouts

JenySadadia:
Support below checkout fields added to v4.5 schema:
- `git_commit_tags`
- `git_commit_message`
- `git_repository_branch_tip`


kcidb/db/bigquery/v04_04.py
```diff
@@ -0,0 +1,96 @@
+"""Kernel CI report database - BigQuery schema v4.3"""
```
spbnick, line 1: Please update the version in the docstring.

kcidb/db/postgresql/v04_09.py
```diff
@@ -0,0 +1,134 @@
...
+    # A map of table names and schemas
+    TABLES = {
+        name: Table(**args) for name, args in TABLES_ARGS.items()
+    }
+
+    # A map of index names and schemas
+    INDEXES = merge_dicts(PreviousSchema.INDEXES, dict(
+        checkouts_git_commit_tags=Index(
+            "checkouts", ["git_commit_tags"]
+        ),
```
spbnick, line 47: I don't think the default BTREE index type would be useful here. I would say `GIN` would be more useful, similarly to `environment_compatible` in `v04_08.py`

kcidb/db/postgresql/v04_09.py
```diff
@@ -0,0 +1,134 @@
...
+                git_repository_branch_tip=BoolColumn(),
+                tree_name=TextColumn(),
+                message_id=TextColumn(),
+                start_time=TimestampColumn(),
+                log_url=TextColumn(),
+                log_excerpt=TextColumn(),
+                comment=TextColumn(),
+                valid=BoolColumn(),
+                misc=JSONColumn(),
+            )),
```
spbnick, line 95: Could you avoid duplicating the query schema here? Instead inherit it the same way `v04_01.py` does. And we would also need to fix `v04_08.py` to do the same.

kcidb/db/sqlite/v04_04.py
```diff
@@ -0,0 +1,116 @@
...
+                git_repository_branch_tip=BoolColumn(),
+                tree_name=TextColumn(),
+                message_id=TextColumn(),
+                start_time=TimestampColumn(),
+                log_url=TextColumn(),
+                log_excerpt=TextColumn(),
+                comment=TextColumn(),
+                valid=BoolColumn(),
+                misc=JSONColumn(),
+            )),
```
spbnick, line 86: Same here about the schema.

kcidb/db/bigquery/v04_00.py
```diff
@@ -504,10 +504,13 @@ class Schema(AbstractSchema):
         checkout="SELECT\n"
                  "    id,\n"
                  "    git_commit_hash,\n"
+                 "    NULL AS git_commit_tags,\n"
+                 "    NULL AS git_commit_message,\n"
```
spbnick, line 508: I think it would be valuable to expose these two fields in "revisions" as well, but the aggregation expressions [might not be so trivial across all the databases](https://chatgpt.com/share/017c66e8-a332-4157-bc25-6b12f98903cd) and this would need some extra testing. So let's leave that for when we need it.

spbnick (changes requested):

kcidb/db/postgresql/v04_09.py
JenySadadia: Right. I'll fix it.

kcidb/db/postgresql/v04_09.py
JenySadadia: Updated it.

kcidb/db/sqlite/v04_04.py
JenySadadia: Updated it.

spbnick (approved):
LGTM! Thanks for fixing the older versions too :heart: 

Let's wait for CI to complete.

helen-fornazier:
awesome, thank you for this.

o/ should I be able to see these columns in playground already?

PR #575, opened 2024-09-13
`kcidb-triage` tool integration

JenySadadia:


JenySadadia:
@helen-fornazier The tool has been deployed on staging. Please test and let me know if it works as expected.
https://console.cloud.google.com/run/detail/us-central1/iss-ed-test-1-kcidb-iss-ed/metrics?project=kernelci-staging

helen-fornazier:
> @helen-fornazier The tool has been deployed on staging. Please test and let me know if it works as expected. https://console.cloud.google.com/run/detail/us-central1/iss-ed-test-1-kcidb-iss-ed/metrics?project=kernelci-staging

I get this when I try to access the url there, are you able to access it?

image

JenySadadia:
> > @helen-fornazier The tool has been deployed on staging. Please test and let me know if it works as expected. https://console.cloud.google.com/run/detail/us-central1/iss-ed-test-1-kcidb-iss-ed/metrics?project=kernelci-staging
> 
> I get this when I try to access the url there, are you able to access it?
> 
> image

Please open cloud shell, fire the command `gcloud beta run services proxy --project kernelci-staging iss-ed-test-1-kcidb-iss-ed --region=us-central1`. That will provide you with the link you would be able to access.

helen-fornazier:
I think we need credentials to postgreesql

if I fill the issue part with

image


and when I click on `Submit`, I get this: 

image

JenySadadia:
> I think we need credentials to postgreesql
> 
> if I fill the issue part with
> 
> image
> 
> and when I click on `Submit`, I get this:
> 
> image

Let me check the error logs.

spbnick:
Submission should be done via the message queue (`...kcidb_new`), not directly to PostgreSQL. We'll need to figure out what's up there.

JenySadadia:
Here are the error logs:
```
  File "/usr/local/lib/python3.9/site-packages/kcidb/issue_editor/kcidb_triage_server.py", line 130, in submit_issue
    submission_id = client.submit(report)
  File "/usr/local/lib/python3.9/site-packages/kcidb/__init__.py", line 84, in submit
    return self.mq_publisher.publish(data)
  File "/usr/local/lib/python3.9/site-packages/kcidb/mq/__init__.py", line 110, in publish
    return self.future_publish(data).result()
  File "/usr/local/lib/python3.9/site-packages/google/cloud/pubsub_v1/publisher/futures.py", line 66, in result
    return super().result(timeout=timeout)
  File "/usr/local/lib/python3.9/concurrent/futures/_base.py", line 446, in result
    return self.__get_result()
  File "/usr/local/lib/python3.9/concurrent/futures/_base.py", line 391, in __get_result
    raise self._exception
  File "/usr/local/lib/python3.9/site-packages/google/cloud/pubsub_v1/publisher/_batch/thread.py", line 274, in _commit
    response = self._client._gapic_publish(
  File "/usr/local/lib/python3.9/site-packages/google/cloud/pubsub_v1/publisher/client.py", line 267, in _gapic_publish
    return super().publish(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/google/pubsub_v1/services/publisher/client.py", line 1067, in publish
    response = rpc(
  File "/usr/local/lib/python3.9/site-packages/google/api_core/gapic_v1/method.py", line 131, in __call__
    return wrapped_func(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/google/api_core/retry/retry_unary.py", line 293, in retry_wrapped_func
    return retry_target(
  File "/usr/local/lib/python3.9/site-packages/google/api_core/retry/retry_unary.py", line 153, in retry_target
    _retry_error_helper(
  File "/usr/local/lib/python3.9/site-packages/google/api_core/retry/retry_base.py", line 212, in _retry_error_helper
    raise final_exc from source_exc
  File "/usr/local/lib/python3.9/site-packages/google/api_core/retry/retry_unary.py", line 144, in retry_target
    result = target()
  File "/usr/local/lib/python3.9/site-packages/google/api_core/timeout.py", line 120, in func_with_timeout
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/google/api_core/grpc_helpers.py", line 78, in error_remapped_callable
    raise exceptions.from_grpc_error(exc) from exc
google.api_core.exceptions.PermissionDenied: 403 User not authorized to perform this action.
```
Submission is happening using message queue but it's not able to find KCIDB credentials for `client.submit`.  How does authentication work for other cloud run services?

spbnick:
Right, you're not supposed to interact **between** deployments and submit your issues/incidents from one deployment to another. Let alone to reach from one project into another. You need to use your deployment's submission queue. Moreover, you don't need to have a "configuration file". You can pass all the configuration you need via the service environment.

Check out `run_deploy()` in `kcidb/cloud/run.sh`. You can see examples of using environment variables. If you need anything (such as the project and topic name), pass it via environment variables. You already have project in that function, you only need to add the (deployment's) topic name. Regarding the origin, I would say it should be `maestro`, since only Maestro folks will be using it.

However, check with Gustavo, because an issue UI is a priority for ProFUSION now, AFAIK. Perhaps you don't even need to do this.

cloud
```diff
@@ -126,6 +126,7 @@ function execute_command() {
     declare -r cache_bucket_name="${project}_${prefix}cache"
     declare -r pick_notifications_trigger_topic="${prefix}pick_notifications_trigger"
     declare -r purge_db_trigger_topic="${prefix}purge_db_trigger"
+    declare -r topic="playground_kcidb_new"
```
spbnick, line 129: Now, you don't hardcode the topic name, you use the submission topic of this deployment. Then when we deploy to playground it will be the one you use here.

You simply need to use the `new_topic` defined below, instead.

cloud
```diff
@@ -380,7 +381,8 @@ function execute_command() {
             --cost-mon-image="$cost_mon_image" \
             --cost-upd-service-account="$cost_upd_service_account" \
             --iss-ed-service="$iss_ed_service" \
-            --iss-ed-image="$iss_ed_image"
+            --iss-ed-image="$iss_ed_image" \
+            --topic="$topic"
```
spbnick, line 385: Please use the name `--new-topic` as everywhere else instead.

kcidb/cloud/run.sh
```diff
@@ -263,6 +265,8 @@ YAML_END
                     limits:
                       cpu: "0.25"
                       memory: "256M"
+                  args:
+                    - $(yaml_quote "$topic")
```
spbnick, line 269: Don't pass it as arguments, instead pass it as an environment variable (e.g. called `KCIDB_NEW_TOPIC`). See the Grafana deployment above, under the `env` key, or read the reference, although it's not very nice: https://cloud.google.com/run/docs/reference/yaml/v1

The project name is likely defined by default as `PROJECT_ID`.

You can use the `kcidb.mq.IOPublisher` class to publish the data: https://github.com/kernelci/kcidb/blob/ded2da11dd92a94466991f3f35b49e6ab048550f/kcidb/mq/__init__.py#L369

Something like this:
```python
client = kcidb.mq.IOPublisher(project_id, topic_name)
client.publish(issue_and_or_incident_data)
```

cloud
JenySadadia: Thanks for the pointers.

kcidb/cloud/run.sh
JenySadadia: @spbnick Thanks for all the help, Nick.
The above method for publishing the data did work.
Bdw just to understand, could you please explain a bit more about `IOPublisher` class and how it manages communications between different deployment instances?

JenySadadia:
@helen-fornazier Deployment https://console.cloud.google.com/run/detail/us-central1/iss-ed-test-2-kcidb-iss-ed/metrics?cloudshell=true&project=kernelci-staging did work and I'm able to submit issues and incidents.
Please test the tool and let me know if anything's wrong.

Below are the responses I got while testing the tool:
```
submit_issue:
{
  "issue_id": "maestro:6ebc39b62a499a7d5941de6e98b231e5613b161d",
  "issue_version": 0,
  "submission_id": "12429416962151915"
}
```
```
submit_incident:
{
  "submission_id": "12429213719411014"
}
```
@spbnick How do I verify these submissions in DB/dashboard?

kcidb/cloud/run.sh
spbnick: `IOPublisher` is a thin wrapper around the [Google Cloud Pub/Sub publisher client](https://cloud.google.com/python/docs/reference/pubsub/latest/google.cloud.pubsub_v1.publisher.client.Client). All it does is validate and encode messages, and then publish them to a topic. See an overview of Pub/Sub, if you'd like an intro: https://cloud.google.com/pubsub/docs/overview

The `IOPublisher` doesn't really "manage communication between deployment instances", all it does is publish stuff to a topic. It depends on which credentials (normally supplied via environment variables) it's using and permissions set up on the project and the topic in the Google Cloud, whether it would be able to publish to a different instance. It doesn't really know anything instances/deployments.

However, the idea of KCIDB deployments is that each is self-contained. A deployment does everything KCIDB does, it's just a copy of KCIDB setup. And so the IOPublisher we'll be creating and using, will be submitting data *within* its deployment. That's ensured by providing it with appropriate parameters within the `cloud` tool and the code you're writing. Once you're done implementing and testing this, and we merge your PR, the staging, playground, and production deployments will be updated to include your setup. And each deployment will get its own instance of the editor.

spbnick:
> @spbnick How do I verify these submissions in DB/dashboard?

In theory, each complete deployment includes an instance of Grafana set up to access the PostgreSQL database. However, due to the limited and convoluted support for deploying Grafana dashboards we only deploy an empty Grafana instance, and then we manually load it with dashboards, as necessary. However, since we're still in the transition period between our Grafana versions and different dashboards, I don't have anything that already has issue support to offer you for loading into an instance.

As such, the only thing you have now is to connect with a PostgreSQL client to you instance's database, and check manually that you get everything in, and perhaps generate notifications using the command-line tools. Otherwise we'll have to limit ourselves to reviews and tests, deploy this, see how it goes, fix any issues, and repeat.

helen-fornazier:
@JenySadadia for some reason I'm still getting 500 internal error when I click to submit

image

helen-fornazier:
@JenySadadia the new deployment is working, thanks!

kcidb/cloud/run.sh
JenySadadia: Thanks for all the info Nick :) Much appreciated!

JenySadadia:
@spbnick As @helen-fornazier has confirmed the tool is working fine, could you please review the source code of the editor?

kcidb/issue_editor/templates/index.html
```diff
@@ -0,0 +1,126 @@
+<!DOCTYPE html>
```
spbnick, line 1: Hmm, this file isn't really a template, but is just a pure HTML file, so I don't think it should be under that directory. We can simplify things if we just put it into `kcidb/issue_editor`.

setup.py
```diff
@@ -102,5 +102,6 @@
     ),
     package_data={
         "": ["*.j2"],
+        "kcidb": ["issue_editor/*", "issue_editor/templates/*"],
```
spbnick, line 105: This is a bit involved, and considering we don't actually have templates, could you just add `"*.html"` to the list of package data patterns?

kcidb/issue_editor/kcidb_triage_server.py
```diff
@@ -0,0 +1,202 @@
...
+
+    report = {
+        "version": {
+            "major": 4,
+            "minor": 3
+        },
+        "checkouts": [],
+        "builds": [],
+        "tests": [],
+        "issues": [],
```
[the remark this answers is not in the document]
JenySadadia: Yes, of course.

setup.py
JenySadadia: I guess  Helen may have added the `templates` directory for future use cases.
@helen-fornazier Could you please confirm if that's not the case and I can delete it?

JenySadadia: @spbnick Looks like a Flask application directory structure should have `templates` directory for HTML files. And that's the default location for `render_template` function (but yes, we can change that behavior).
As per the convention, we should keep it.
See https://flask.palletsprojects.com/en/2.3.x/tutorial/layout/

spbnick: OK. Then could we rename `index.html` to `index.html.j2`? Because that's what it essentially is, even if you're not using any Jinja2 there (it's still being interpreted as such). This will also make sure that the file is included into the package.

JenySadadia: I have renamed the template.
But I need to make changes in `setup.py` to include jinja2 templates recursively.

setup.py
```diff
@@ -101,6 +101,6 @@
         ]
     ),
     package_data={
-        "": ["*.j2"],
+        "": ["*.j2", "**/*.j2"]
```
spbnick, line 104: Ah, sure! Only, we won't need the `"*.j2"` anymore as `"**/*.j2"` includes that (and I suspect `"*.j2"` never worked as intended :see_no_evil:).

spbnick (changes requested):
Looks great! I have one last nitpick. Thank you, Jeny!

spbnick (approved):
Thank you, Jeny :heart:

PR #577, opened 2024-09-17
Update fluster test report link

JenySadadia:
Include `tree` and `branch` for accurate test report link.

### DIFF
--- a/kcidb/templates/fluster_revision_description.txt.j2
+++ b/kcidb/templates/fluster_revision_description.txt.j2
@@ -13,21 +13,31 @@ REVISION
         {% if revision.git_commit_hash %}
             {{- "        hash: " + revision.git_commit_hash }}
         {% endif %}
-     Checked out from
+    Checked out from
         {% for repo, branch_checkouts
            in revision.repo_branch_checkouts.items() %}
             {{- "        " +
                 (([repo] + (branch_checkouts | list)) |
                  reject("none") | join(" ")) }}
         {% endfor %}
-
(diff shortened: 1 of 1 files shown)


kcidb/templates/fluster_revision_description.txt.j2
```diff
@@ -13,21 +13,40 @@ REVISION
...
+                    {% set _ = ns.branch_tree.__setitem__(
+                        branch, tree_name) %}
+                {% endif %}
+            {% endfor %}
+        {% endfor %}
+    {% endfor %}
+    {% if ns.branch_tree %}
+        {% set branch = ns.branch_tree | first %}
+        {% set tree = ns.branch_tree[branch] %}
+    {% endif %}
```
spbnick, line 39: I think you could simplify this quite a bit by iterating over `revision.checkouts`, and finding the first one with both branch and tree.

Something like this:
```jinja
    {% set ns = namespace(branch=none, tree=none) %}
    {% for checkout in revision.checkouts %}
        {% if checkout.tree is not none and checkout.branch is not none and
              ns.branch is none %}
            {% set ns.branch = checkout.branch %}
            {% set ns.tree = checkout.tree %}
        {% endif %}
    {% endfor %}
```

Then just use `ns.branch` and `ns.tree` in your URL.

JenySadadia: That's a lot easier. Thanks.

spbnick (approved):
LGTM!

PR #581, opened 2024-09-20
Push production budget limit up temporarily

spbnick:


spbnick:
Right, we gotta fix the new errors, but we also got to stop the monitor from killing our service for now. This is not going to get deployed, but I've manually edited the YAML for the service for now. We can fix CI later. Merging as is.

PR #583, opened 2024-09-24
cloud: Mute updates to relieve PostgreSQL temporarily

spbnick:
This will stop notifications working on production.

CC @JenySadadia, am I right that we have no notifications sent to maintainers ATM?

I'll need to fix the performance, which I've been working on before LPC, but now I gotta get rid of COVID first.


JenySadadia:
> This will stop notifications working on production.
> 
> CC @JenySadadia, am I right that we have no notifications sent to maintainers ATM?

We are sending reports to `stable-rt` maintainers' mailing list.
> 
> I'll need to fix the performance, which I've been working on before LPC, but now I gotta get rid of COVID first.

PR #584, opened 2024-09-26
stable-rc: display git commit tags in the report

JenySadadia:


spbnick (approved):
LGTM, although it would be nice to wrap that long `set` statement in the template.

spbnick (approved):
Nice! Thank you, Jeny :pray:

PR #586, opened 2024-10-03
Automate incidents creation

JenySadadia:
Automate the creation of incidents based on issues using `kcidb_match` tool.


kcidb/tools/kcidb_match.py
```diff
@@ -0,0 +1,471 @@
...
+        self.add_pattern(issue.id, issue.version, pattern_object)
+
+
+class PatternValidator:
+    """Pattern validator class"""
+    def __init__(self):
+        self.schema = copy.deepcopy(kcidb.io.SCHEMA.json)
+        self.remove_required_fields(self.schema)
+
+    def remove_required_fields(self, schema):
```
helen-fornazier, line 133: """We use the same KCIDB schema to validade, but without any required fields, so remove the required rule for our validation"""

kcidb/tools/kcidb_match.py
```diff
@@ -0,0 +1,471 @@
...
+        if "properties" in schema:
+            for _, subschema in schema["properties"].items():
+                self.remove_required_fields(subschema)
+        if "items" in schema:
+            self.remove_required_fields(schema["items"])
+        if "$defs" in schema:
+            for _, subschema in schema["$defs"].items():
+                self.remove_required_fields(subschema)
+
+    def validate_pattern_object(self, pattern_object):
```
helen-fornazier, line 145: """Validate if pattern_object from the user was given as expected, as per KCIDB schema"""

kcidb/tools/kcidb_match.py
```diff
@@ -0,0 +1,471 @@
...
+                    return False
+
+        if 'tests' in pattern_obj and 'log_regex' in pattern_obj['tests'][0]:
+            return self.check_log_match(kcidb_obj, pattern_obj, "tests")
+        if 'builds' in pattern_obj and 'log_regex' in pattern_obj['builds'][0]:
+            return self.check_log_match(kcidb_obj, pattern_obj, "builds")
+
+        return True
+
+    def check_log_match(self, kcidb_io_object, pattern_object, key):
```
helen-fornazier, line 221: """When the pattern_object contains a log_regex, we first try to match it against log_excerpt, if we can't find any occurrence there, we download the entire log file and try to match on it, return True if we were able to find the log_regex, False otherwise"""

kcidb/tools/kcidb_match.py
```diff
@@ -0,0 +1,471 @@
...
+            'issue_version': issue_version,
+            'present': True,
+            type_id_key: type_id_value,
+        }
+
+    def generate_incident_on_match(self, kcidb_io_object, issue_id,
+                                   issue_version, issue_pattern_object):
+        """Generate incident if issue pattern is found in a build/test
+        object"""
+        incidents = []
```
helen-fornazier, line 275: it seems this is returning a single instance, we can remove the array. This was probably a left over of some of my experimentations

kcidb/tools/kcidb_match.py
```diff
@@ -0,0 +1,471 @@
...
+                           "checkouts": [test.build.checkout._data]}
+        return self.generate_incidents_from_db(kcidb_io_object)
+
+    def generate_incidents_from_build(self, build):
+        """Generate incident from build object"""
+        kcidb_io_object = {"builds": [build._data],
+                           "checkouts": [build.checkout._data]}
+        return self.generate_incidents_from_db(kcidb_io_object)
+
+    def generate_incidents_from_db(self, kcidb_io_object):
```
helen-fornazier, line 297: """generate incidents by trying to match the kcidb_io_object against the patterns saved in DB_NAME"""

kcidb/tools/kcidb_match.py
```diff
@@ -0,0 +1,471 @@
...
+            pattern_object = json.loads(pattern_object_json)
+            incidents.extend(self.generate_incident_on_match(
+                kcidb_io_object, issue_id, issue_version, pattern_object))
+
+        return {
+            "version": KCIDB_IO_VERSION,
+            "incidents": incidents
+        }
+
+    def generate_incidents_from_object(self, kcidb_io_object):
```
helen-fornazier, line 311: """Similar to generate_incidents_from_db, but instead of trying to match against patterns saved in DB_NAME, it tries to match against patterns in the issues defined in kcidb_io_object (considering this object a collection os checkouts, builds, tests and issues)"""

helen-fornazier: Maybe this doesn't make much sense and we can drop this

kcidb/tools/kcidb_match.py
```diff
@@ -0,0 +1,471 @@
...
+                kcidb_io_object, issue['id'], issue['version'],
+                pattern_object))
+
+        return {
+            "version": KCIDB_IO_VERSION,
+            "incidents": incidents
+        }
+
+
+def fetch_kcidb_io_object(query_id, db_conn, query_type):
```
helen-fornazier, line 329: this fetches the parents of the object using kcidb-query, since kcidb already provides the objects with all its parent, we can drop this

kcidb/tools/kcidb_match.py
```diff
@@ -0,0 +1,471 @@
...
+    command = ["kcidb-query", f"-{query_type[0]}", query_id, "-d",
+               db_conn, "--parents"]
+    result = subprocess.run(command, capture_output=True, text=True)
+    if result.returncode != 0:
+        raise RuntimeError(f"Failed to fetch KCIDB IO object: {result.stderr}")
+    return json.loads(result.stdout)
+
+
+def validate_kcidb_io_object(kcidb_io_object):
```
helen-fornazier, line 338: this validates if the given kcidb_io_object contains the minumum for this tool to work, but since we'll be receiving things from kcidb, we can drop this too

kcidb/tools/kcidb_match.py
```diff
@@ -0,0 +1,471 @@
...
+        parser.error("--db_conn is required when using --check_test_id or "
+                     "--check_build_id")
+
+    if args.check_test_id or args.check_build_id:
+        args.ignore_db = True
+
+    return args
+
+
+def main():
```
helen-fornazier, line 430: we need to re-think about this command line tool, maybe it could be simpler as:

cat issue.json | kcidb-match # nothing to output
cat tests.json | kcidb-match # and this would output the incidents that was generated

we could also have

kcidb-match --list-patterns # that would list all the saved objects inside patterns.db
kcidb-match --clear-db # that would erase pattern.db

but those two are not a priority, we can do this by navigating patterns.db with other tools

JenySadadia:
@helen-fornazier Thanks for the comments.
It would have been easier if you had committed the changes directly to this PR.
But no worries, I'll do it.

kcidb/tools/kcidb_match.py
JenySadadia: This function is being used in `main`. If it's not useful anymore I can drop it and modify the `main` function.

kcidb/tools/kcidb_match.py
JenySadadia: The same comment as https://github.com/kernelci/kcidb/pull/586#discussion_r1787259604

kcidb/tools/kcidb_match.py
JenySadadia: The same comment as https://github.com/kernelci/kcidb/pull/586#discussion_r1787259604

kcidb/tools/kcidb_match.py
JenySadadia: A thought:
@helen-fornazier Do you think this tool is going to be used as a command line tool?
The current use of this tool is with the subscription module.
If we are going to expand the use cases of this tool, then only we should spend time refactoring the whole `main` function.

kcidb/tools/kcidb_match.py
```diff
@@ -0,0 +1,482 @@
...
+        If snippet_lines < 0: the last snippet_lines log lines
+    """
+    response = requests.get(url, timeout=60)
+    if len(response.content) == 0:
+        return None
+    try:
+        raw_bytes = gzip.decompress(response.content)
+        text = raw_bytes.decode('utf-8')
+    except gzip.BadGzipFile:
+        text = response.text
```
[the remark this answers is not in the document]
helen-fornazier: make sense, as a suggestion of the MAX_CONTENT_LENGTH, we could get a random log, check its size, multiplied by 10 and voilà (we can review this length latter if needed)

kcidb/tools/kcidb_match.py
helen-fornazier: For command line tool we would only use it as developers for debug, a normal user wouldn't use it I would say

kcidb/tools/kcidb_match.py
JenySadadia: I believe you meant I can drop this and other similar functions from `main`.

kcidb/tools/kcidb_match.py
JenySadadia: Thanks for clarifying.
As this tool is a high priority and we want to merge this asap, I won't spend time refactoring the main function atm. We can follow this up later if required.

helen-fornazier:
Proposal for command line tool usage:

```
cat issue.json | kcidb-issues --pattern-db-file=DB_NAME # this would save the patterns in `DB_NAME`

cat tests.json | kcidb-issues --pattern-db-file=DB_NAME # this would dump a json with the incidents (if it was created) matched agains `DB_NAME`
```

with this, we can use this tool to debug, and also to validate a issue and test example in the editor UI before submitting it to kcidb

@tales-aparecida @JenySadadia @spbnick what do you think?

so we can drop all the options I had previously added there

helen-fornazier:
> Proposal for command line tool usage:
> 
> ```
> cat issue.json | kcidb-issues --pattern-db-file=DB_NAME # this would save the patterns in `DB_NAME`
> 
> cat tests.json | kcidb-issues --pattern-db-file=DB_NAME # this would dump a json with the incidents (if it was created) matched agains `DB_NAME`
> ```
> 
> with this, we can use this tool to debug, and also to validate a issue and test example in the editor UI before submitting it to kcidb
> 
> @tales-aparecida @JenySadadia @spbnick what do you think?
> 
> so we can drop all the options I had previously added there

Following up this comment https://github.com/kernelci/kcidb/pull/586#issuecomment-2392895992

I think we can leave it to a second PR, no problem for me

helen-fornazier (dismissed):

kcidb/tools/kcidb_match.py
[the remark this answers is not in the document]
JenySadadia: Hello,
It looks like we can't decompress gzip data in chunks. We need to use the whole compressed data for it.
Or do you mean, something like the below?
```

    try:
        raw_bytes = gzip.decompress(response.content)
        text = raw_bytes.decode('utf-8')
    except gzip.BadGzipFile:
	chunk = next(response.iter_content(MAX_CONTENT_LENGTH, 	decode_unicode=True), None)
```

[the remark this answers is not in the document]
tales-aparecida: That could be enough, but this is also something that could be worked in a future PR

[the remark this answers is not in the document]
JenySadadia: Yes, let's keep it for later.

kcidb/tools/kcidb_match.py
```diff
@@ -0,0 +1,426 @@
...
+        match it against log_excerpt, if it can't find any occurrence there,
+        it will download the entire log file and try to match on it.
+        Returns True if it's able to find the log_regex, False otherwise"""
+        def search_regex(log_regex, log_excerpt, log_url):
+            if re.search(log_regex, log_excerpt):
+                return True
+
+            log_content = get_log(log_url)
+            if log_content and re.search(log_regex, log_content):
+                return True
```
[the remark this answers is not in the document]
JenySadadia: Thanks for the input.
Yes, we'll come back here at the time of follow-up PR.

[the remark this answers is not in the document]
JenySadadia: I need to mark the conversation resolved to be able to merge the PR.

helen-fornazier (approved):

kcidb/tools/kcidb_match.py
```diff
@@ -0,0 +1,429 @@
...
+            "version": KCIDB_IO_VERSION,
+            "incidents": incidents
+        }
+
+    def generate_incidents_from_test(self, test):
+        """Generate incident from test object"""
+        kcidb_io_object = {"tests": [test._data],
+                           "builds": [test.build._data],
+                           "checkouts": [test.build.checkout._data]}
+        return self.generate_incidents_from_db(kcidb_io_object)
```
spbnick, line 319: You cannot make an I/O object from an OO object in general, even if it looks like you can. You just cannot count on that. An OO object is a processed I/O object, and there can be data loss. Instead deal with OO objects directly everywhere. Where you need to process I/O objects from stdin and command-line interface, load them into an sqlite database using the database client, get them as OO objects from there, and then process.

OO objects were specifically made to make the things you're doing here easier to do. Like walk related objects and so on. We can have a call this week and go over all the concerns and options regarding this.

kcidb/monitor/subscriptions/create_incidents.py
```diff
@@ -0,0 +1,42 @@
...
+            CLIENT = kcidb.Client(project_id=project_id, topic_name=topic_name)
+    return CLIENT
+
+
+def match_test(test):
+    """Generate incident for matching test"""
+    client = get_client()
+    if client:
+        incident_generator = kcidb_match.IncidentGenerator()
+        incidents = incident_generator.generate_incidents_from_test(test)
```
spbnick, line 26: Considering that `IncidentGenerator` encapsulates a database client connection, why not create and get it similarly to `get_client()`, instead of creating a new one for *every* object matched?

spbnick (changes requested):
Sorry to leave a possibly overwhelming review, asking to change things a lot. Let's discuss this and see what's viable at this moment. However, this would all be for nothing if I don't fix that performance problem I was working on before Plumbers, and don't enable notifications. This code would simply not be called without that. So I'll go and work on that, instead of going further into detailed review :see_no_evil:

kcidb/tools/kcidb_match.py
spbnick: BTW, there's also a schema for the raw OO data in `kcidb.orm.data`.

PR #601, opened 2024-11-06
Codec summary report: handle "No tests found"

JenySadadia:
Add checks for fluster tests to avoid generating emails with "No tests found".

### DIFF
--- a/kcidb/monitor/subscriptions/codec_summary.py
+++ b/kcidb/monitor/subscriptions/codec_summary.py
@@ -18,6 +18,15 @@ def match_revision(revision):
     if not revision.tests_root["fluster"].nodes:
         return ()
 
+    # If only "validate-fluster-results" found or environment compatible
+    # is not defined for any tests
+    if all(
+        t.path.endswith('.validate-fluster-results') or
+        not t.environment_compatible
+        for t in revision.tests_root["fluster"].tests
+    ):
+        return ()
+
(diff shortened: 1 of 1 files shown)


kcidb/monitor/subscriptions/codec_summary.py
```diff
@@ -18,6 +18,16 @@ def match_revision(revision):
...
 
+    # If only "validate-fluster-results" found or environment compatible
+    # is not defined for any tests
+    fluster_tests = [
+        t.path.split('.')[-1] for t in revision.tests_root["fluster"].tests
+        if t.path.split('.')[-1] != 'validate-fluster-results'
+        and t.environment_compatible
+    ]
+    if not fluster_tests:
+        return()
```
spbnick, line 29:
```suggestion
    # If nothing but "validate-fluster-results" found or
    # there are no tests with environment compatible
    if all(
        t.path.endswith('.validate-fluster-results') or
        not t.environment_compatible
        for t in revision.tests_root["fluster"].tests
    ):
        return()
```

spbnick (dismissed):
LGTM, just left one readability/style suggestion, but it's optional.

kcidb/monitor/subscriptions/codec_summary.py
JenySadadia: Thanks for the suggestion.

spbnick (approved):
LGTM!

PR #607, opened 2024-11-11
fluster report: specify data source in the report link

JenySadadia:
`var-datasource=default` in the fluster test summary link leads to `playground` rather than `production` on the dashboard. Specify the exact data source instead of relying on the default value.

### DIFF
--- a/kcidb/templates/fluster_revision_description.txt.j2
+++ b/kcidb/templates/fluster_revision_description.txt.j2
@@ -34,9 +34,9 @@ REVISION
 
 See complete and up-to-date report at:
     {% if ns.tree is not none %}
-        https://grafana.kernelci.org/d/codecs/codecs?orgId=1&var-platform=All&var-datasource=default&var-git_commit_hash={{revision.git_commit_hash | urlencode }}&var-tree={{ns.tree}}&var-branch={{ns.branch}}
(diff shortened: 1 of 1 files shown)


spbnick:
The default datasource has switched unexpectedly after Grafana upgrade. I fixed it for that dashboard, so there's no need for this change. Let's keep the links in emails as short as possible.

spbnick:
Actually, you can remove this whole piece of the URLs: `orgId=1&var-platform=All&var-datasource=default&` Can you try that?

spbnick:
You will need to at least remove `var-datasource=default` to fix the links, though.

spbnick:
Let's reopen this, as the links are still broken because of that.

JenySadadia:
I think it's better to specify the exact data source in the URL rather than relying on the default one.
Otherwise, it could break in the future just like it's breaking now.

spbnick (dismissed):

JenySadadia:
> However, Grafana is difficult to keep stable in general, since there's a lot manual work involved in setting everything up, and upgrades shift things, so other things may break as well. FWIW, it's better to make as little assumptions as possible about the way it works (like how datasources are identified), in case they change. That's also why I suggest making URLs as short as possible.

If shortening the URL is better from Grafana's pov, let's go with that.

spbnick (approved):

PR #613, opened 2024-11-19
Fix unlimited archiving

spbnick:


spbnick:
The purge test failed again. I'll merge this and see if we can make the test better.

PR #618, opened 2024-11-29
Fix empty `stable-rt` reports

JenySadadia:
Add a check to verify if there are builds from selected origins before generating the report.


JenySadadia:
Added the commit from this PR to https://github.com/kernelci/kcidb/pull/619.

PR #619, opened 2024-11-29
Fix empty reports: `stable-rc`, `stable-rt`, `fluster`

JenySadadia:
Add a check to verify if there are builds from selected origins before generating the report.

### DIFF
--- a/kcidb/monitor/subscriptions/linux_stable_rt.py
+++ b/kcidb/monitor/subscriptions/linux_stable_rt.py
@@ -16,11 +16,12 @@ def match_revision(revision):
     if repo_url not in revision.repo_branch_checkouts:
         return ()
 
-    if revision.builds_valid is None:
-        return ()
+    selected_origins = {'maestro', 'broonie'}
 
-    # If the revision is not from 'maestro' or 'broonie' origin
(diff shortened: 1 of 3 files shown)


kcidb/monitor/subscriptions/linux_stable_rt.py
```diff
@@ -16,11 +16,19 @@ def match_revision(revision):
...
+    # Return if there are no builds or status is unknown for all the builds
     if revision.builds_valid is None:
         return ()
 
-    # If the revision is not from 'maestro' or 'broonie' origin
-    if not {c.origin for c in revision.checkouts} & {'maestro', 'broonie'}:
+    # If no builds found from 'maestro' or 'broonie' origin
+    builds = [build for build in revision.builds
+              if build.origin in selected_origins]
+    if not builds:
```
spbnick, line 31: You can replace all this with:
```suggestion
    if all(b.origin not in selected_origins or b.valid is None
           for b in revision.builds):
```
You don't really have to check the checkout origins.

JenySadadia: Thanks for the suggestion. This is neat.
Updated both subscription modules. Also, added a fix for fluster test report.

spbnick (approved):
LGTM!

PR #625, opened 2025-02-19
subscriptions: drop notifications we are not using anymore

padovan:
We were really able to the stable-rc notification really useful for example. So let's drop it and solve this through the new notifications infra.


spbnick:
Thank you, @padovan! Please also drop the corresponding test in `kcidb/test_monitor.py` to fix CI.

PR #627, opened 2025-03-08
(tests.yaml): Remove kver tests

nuclearcat:
This is more "canary" or PoC test, than real test, also by some reason it often gives false positives, not adding any value.
We decided it is better to remove it.


spbnick:
Thanks a lot, @nuclearcat!

PR #636, opened 2025-04-01
orm: Stub out log_excerpts

spbnick:
Remove `log_excerpt` from ORM, thus stopping retrieving it, and saving memory when fetching objects. Stub it out with a property returning None in Checkout, Build, and Test to keep templates unchanged, but thinking we don't have anything. This should tide us over until we can think of something better.


tales-aparecida (approved):

tales-aparecida:
I don't know how things are connected today, does this affect the dashboard in any way?

spbnick:
> I don't know how things are connected today, does this affect the dashboard in any way?

This only concerns notifications generated by KCIDB itself. There will be some functionality loss, but let's see if anyone notices.

PR #647, opened 2025-05-12
Update kcidb documentation

nuclearcat:
With introduction of kcidb-rest we need to update documentation, and even before that it need to be updated about CloudSQL role, and fact that BigQuery is mostly deprecated.


padovan (approved):
It looks good to me! It describes the transition well.

PR #652, opened 2025-07-01
Fix kcidb-submit NotImplemented error

nuclearcat:
kcidb-submit uses iterative submission functions, which are not migrated to REST interface yet. This patch implement REST submissions with backward compatibility, not fastest yet, but most stable for a moment.

### DIFF
--- a/kcidb/__init__.py
+++ b/kcidb/__init__.py
@@ -245,6 +245,19 @@ class Client:
             `NotImplementedError`, if not supplied with a project ID or an MQ
             topic name at initialization time.
         """
+        # TODO: Implement futures/parallelism
+        if self._resturi:
+            try:
+                for data in data_iter:
+                    submission_id = self.rest_submit(data)
+                    if done_cb:
+                        done_cb(submission_id)
+            # pylint: disable=broad-exception-caught
+            except Exception as e:
+                print(f"Error submitting report: {e}",
+                      file=sys.stderr)
+                return None
+            return None
         if not self.mq_publisher:
             raise NotImplementedError
         return self.mq_publisher.publish_iter(data_iter, done_cb=done_cb)


tales-aparecida (approved):

kcidb/__init__.py
```diff
@@ -245,6 +245,19 @@ def submit_iter(self, data_iter, done_cb=None):
             `NotImplementedError`, if not supplied with a project ID or an MQ
             topic name at initialization time.
         """
+        # TODO: Implement futures/parallelism
+        if self._resturi:
+            for data in data_iter:
+                try:
```
tales-aparecida: Nit picking:

https://docs.astral.sh/ruff/rules/try-except-in-loop/

Prior to Python 3.11, exception handling incurred some performance overhead, even if no exception was raised. Therefore, placing try-except inside a tight loop could potentially impact performance, especially in highly optimized code.

tales-aparecida:
pylint is complaining about "broad-except", but since there are a lot of occurrences of it in this file, I'd recommend `# pylint: disable=broad-except` today and writing a separate PR cleaning all of them at once (in multiple commits, to be clear)

kcidb/__init__.py
```diff
@@ -245,6 +245,19 @@ def submit_iter(self, data_iter, done_cb=None):
...
+                    submission_id = self.rest_submit(data)
+                    if done_cb:
+                        done_cb(submission_id)
+                except Exception as e:
+                    print(f"Error submitting report: {e}",
+                          file=sys.stderr)
+                    # Handle submission error
+                    return
+            return
         if not self.mq_publisher:
```
tales-aparecida:
```suggestion
                    return None
            return None
        if not self.mq_publisher:
```

To comply with pylint

kcidb/__init__.py
nuclearcat: Thanks! Adjusted, even it is temporary code and actually slower than should (it is supposed to be done over futures and etc, but i prefer to finish migration asap), i think it is better to do right.
Thank for info, will keep in mind in future.
Amended/fixed.

tales-aparecida:
A possible fix for the `check_auxiliary_files` https://github.com/kernelci/kcidb/pull/653 (but it'll probably get stuck on the pytests)


EDIT: nevermind, it's not blocking and it's likely dropping requests from github actions

tales-aparecida (approved):

nuclearcat:
> A possible fix for the `check_auxiliary_files` #653 (but it'll probably get stuck on the pytests)
> 
> EDIT: nevermind, it's not blocking and it's likely dropping requests from github actions

I am afraid many tests are broken and i am considering to rewrite codebase and tests as soon as all submitters sync to REST interface and latest schema.

PR #653, opened 2025-07-01
test-catalog: Update Connectathon_test_suite URL

tales-aparecida:
The wiki subdomain was timing out


tales-aparecida:
They're probably dropping requests from gitlab (or just overwhelmed by bots like us tbh)

PR #658, opened 2025-07-29
client: Use logger rather than console prints

broonie:
Currently the REST implementation uses print() for logging rather than
the module's logger, logging every submission.  Convert this to a debug
log via the standard python logging framework.


broonie:
The check_auxiliary_files failure looks to be a false positive due to networking or other infra issues.

broonie:
The check_auxiliary_files failure looks to be a false positive due to networking or other infra issues.

PR #659, opened 2025-07-30
kcidb-submit: implement futures for faster submission

nuclearcat:


kcidb/__init__.py
```diff
@@ -222,16 +272,20 @@ def future_submit(self, data):
         assert io.SCHEMA.is_compatible(data)
         assert LIGHT_ASSERTS or io.SCHEMA.is_valid(data)
         # Submit over rest if self._rest is set
-        # TODO: future?
         if self._resturi:
-            return self.rest_submit(data)
+            # Create a thread pool executor and submit the REST request with retry
+            executor = concurrent.futures.ThreadPoolExecutor(max_workers=1)
```
broonie: I think rather than creating an executor per call we should be creating one on the object (probably on first use I guess?). When I throw a bunch of submissions into the library I'm expecting it to deal with concurrency limiting as required to avoid overwhelming the server. This is also going to be a bunch cheaper than spawning threads I expect.

kcidb/__init__.py
```diff
@@ -240,13 +294,15 @@ def submit_iter(self, data_iter, done_cb=None):
...
+            max_workers: Maximum number of parallel submission threads.
+                        Defaults to 10.
 
         Raises:
             `NotImplementedError`, if not supplied with a project ID or an MQ
             topic name at initialization time.
         """
-        # TODO: Implement futures/parallelism
         if self._resturi:
+<<<<<<< Updated upstream
```
broonie: Conflict marker?

kcidb/__init__.py
```diff
@@ -240,13 +294,15 @@ def submit_iter(self, data_iter, done_cb=None):
             done_cb:    A function to call when a report is successfully
                         submitted. Will be called with the submission ID of
                         each report returned by the iterator, in order.
+            max_workers: Maximum number of parallel submission threads.
```
broonie: If we go with a worker pool on the kcidb object then the max_workers configurability would be on the object creation.

kcidb/__init__.py
```diff
@@ -258,6 +314,38 @@ def submit_iter(self, data_iter, done_cb=None):
                       file=sys.stderr)
                 return None
             return None
+=======
+            # Convert iterator to list to preserve order for done_cb
+            data_list = list(data_iter)
+            submission_results = []
+
+            with concurrent.futures.ThreadPoolExecutor(max_workers=max_workers) as executor:
+                # Submit all tasks with retry logic
```
broonie: Should the retry logic just be in the standard future submission call and this just be implemented as calling future_submit()?

kcidb/__init__.py
```diff
@@ -258,6 +314,38 @@ def submit_iter(self, data_iter, done_cb=None):
...
+                }
+
+                # Process completed futures in order
+                for future in concurrent.futures.as_completed(future_to_data):
+                    idx, data = future_to_data[future]
+                    try:
+                        submission_id = future.result()
+                        submission_results.append((idx, submission_id, None))
+                    except Exception as e:
+                        print(f"Error submitting report: {e}",
```
broonie: Should use logger.

broonie (changes requested):

kcidb/__init__.py
nuclearcat: done

kcidb/__init__.py
nuclearcat: Done

broonie:
Sorry, missed this when you initially pushed it - it looks good on code review and seems to be testing OK in my system. Only issue is the volume of 'Submitting over REST API' log messages but that's not a blocker.

nuclearcat:
> Sorry, missed this when you initially pushed it - it looks good on code review and seems to be testing OK in my system. Only issue is the volume of 'Submitting over REST API' log messages but that's not a blocker.

I will remove it, thanks for testing!

broonie:
Could we get this deployed please? It'd make my life building my infra much easier to not be relying on applying an out of tree patch.

PR #662, opened 2026-02-02
Remove old pubsub

nuclearcat:
It is very difficult decision and move, but we need to declutter code from proprietary Google Cloud stuff, as we try to stay provider agnostic.


tales-aparecida:
FWIW, CKI is not using this lib anymore. We hit the API directly

